PR #5, opened 2023-05-02
ldbsearch utility added

patriki01:


pbrezina (changes requested):

sssd_test_framework/misc/__init__.py
```diff
@@ -97,3 +98,16 @@ def to_list_without_none(r_list: list[Any]) -> list[Any]:
     :rtype: list[Any]
     """
     return [x for x in r_list if x is not None]
+
+
+def parse_ldb_output(output: str) -> list[list[str]]:
```
pbrezina, line 103: ldb prints the output in ldif format so `parse_ldif` would be a better name. Also please add unit tests for this.

sssd_test_framework/misc/__init__.py
```diff
@@ -97,3 +98,16 @@ def to_list_without_none(r_list: list[Any]) -> list[Any]:
...
+    """
+    Remove comment lines (with #) then group records by empty line.
+
+    :param output: output of ldbsearch
```
pbrezina, line 107: rename parameter to ldif

sssd_test_framework/misc/__init__.py
pbrezina: And it should return the records in dict format directly, ie. instead of calling `attrs_parse(parse_ldb_output(...))` you want to get the result directly from `parse_ldb_output(...)`.

sssd_test_framework/utils/ldb.py
```diff
@@ -0,0 +1,61 @@
...
+
+from pytest_mh import MultihostHost, MultihostUtility
+
+from sssd_test_framework.misc import attrs_parse, parse_ldb_output
+
+
+class LBDUtils(MultihostUtility[MultihostHost]):
+    """
+    Class for ldb tools.
+
```
pbrezina, line 11: Remove the line.

sssd_test_framework/utils/ldb.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import annotations
+
+from pytest_mh import MultihostHost, MultihostUtility
+
+from sssd_test_framework.misc import attrs_parse, parse_ldb_output
+
+
+class LBDUtils(MultihostUtility[MultihostHost]):
+    """
+    Class for ldb tools.
```
pbrezina, line 10: LDB Tools wrapper?

sssd_test_framework/utils/ldb.py
```diff
@@ -0,0 +1,61 @@
...
+
+    """
+
+    def search(
+        self,
+        path: str,
+        basedn: str | None = None,
+        scope: str | None = None,
```
pbrezina, line 18: this should be enum

sssd_test_framework/utils/ldb.py
```diff
@@ -0,0 +1,61 @@
...
+        filter: str | None = None,
+        args: list[str] | None = None,
+    ) -> dict[str, dict[str, list[str]]]:
+        """
+        Run ``ldbsearch -H path`` command and transform the data to ``dictionary``.
+        Can be called with additional arguments.
```
pbrezina, line 24: Separate with blank line.

pbrezina (changes requested):
Hi, thank you. This looks better now, but there are still few comments inside.

sssd_test_framework/utils/ldb.py
```diff
@@ -0,0 +1,67 @@
...
+
+        ONE_LEVEL = "one"
+        SUBTREE = "sub"
+        BASE = "base"
+
+    def search(
+        self,
+        path: str,
+        basedn: str | None = None,
+        scope: "LDBUtils.Scope" | None = None,
```
pbrezina, line 28: Quotes can be removed when `from __future__ import annotations` is present.

pbrezina: Default scope should be subtree. This is the default scope when calling ldbsearch without `-s`. Even though it is unlikely that the default value will change, we should hard code it here to avoid issues if it happens.

sssd_test_framework/utils/ldb.py
```diff
@@ -0,0 +1,67 @@
...
+            additional += ["-b", basedn]
+
+        if filter is not None:
+            additional.append(filter)
+
+        cmd = ["ldbsearch", *args, "-H", path, *additional]
+
+        stdout = self.host.ssh.exec(cmd).stdout
+
+        return parse_ldif(stdout)
```
pbrezina, line 67: This is shorter:

```python
        result = self.host.ssh.exec(["ldbsearch", *args, "-H", path, *additional])
        return parse_ldif(result.stdout)
```

tests/test_misc.py
```diff
@@ -100,3 +101,62 @@ def test_to_list_of_strings(value, expected):
...
+# record 2
+debug: prvni
+dn: cn=2
+cn: 2
+debug: druhy
+"""
+
+out2 = {
+    "cn=1": {"dn": ["cn=1"], "cn": ["1"], "debug": ["prvni"]},
+    "cn=2": {"dn": ["cn=2"], "cn": ["2"], "debug": ["prvni", "druhy"]},
```
pbrezina, line 135: Use English please :-)

tests/test_misc.py
```diff
@@ -100,3 +101,62 @@ def test_to_list_of_strings(value, expected):
...
+
+out3 = {
+    "cn=sssd": {"dn": ["cn=sssd"], "cn": ["sssd", "sssd"]},
+}
+
+
+@pytest.mark.parametrize(
+    "value,expected",
+    [
+        (inp1, out1),
```
pbrezina, line 156: Expand the variables here, please. You can use nice indentation, then dedent it with textwrap.dedent().

Using vague names like inp1, out1 does not work very well, since you can have more parametrized tests.

pbrezina (changes requested):
Hi, thank you. We are almost there, please see few minor comments inside.

tests/test_misc.py
```diff
@@ -100,3 +103,109 @@ def test_to_list_of_strings(value, expected):
...
+cn: sssd
+
+# returned 1 records
+# 1 entries
+# 0 referrals
+"""
+
+out3 = {
+    "cn=sssd": {"dn": ["cn=sssd"], "cn": ["sssd", "sssd"]},
+}
```
pbrezina, line 152: This can be removed now.

pbrezina (changes requested):

tests/test_misc.py
patriki01: Of course. Done.

pbrezina:
I did some testing, everythings works correctly. I added a test case for empty result and let's push it when CI is finished.

pbrezina (approved):

PR #6, opened 2023-05-02
tools.id() can accept str or int now

patriki01:


pbrezina:
Thank you. Can you just amend the commit message to "tools: id can accept str or int"?

patriki01:
Yes. Renamed.

PR #8, opened 2023-05-05
sss_override: add utility to manage local overrides

sidecontrol:
- OverrideUtils Object, user and group sub classes
- OverrideUser Object adding override.user add, delete, get, modify, backup and restore functions
- OverrideGroup Object adding override.group add, delete, get, modify, backup and restore functions

### DIFF
--- a/docs/guides/sss_override.rst
+++ b/docs/guides/sss_override.rst
@@ -1,0 +1,37 @@
+Local override of users and groups
+###################################
+
+Class :class:`sssd_test_framework.utils.sss_override.SSSOverrideUtils` provides
+an API to manage local overrides for users and groups. Also known as ID views,
+instead of being stored in IPA/LDAP server the override data is stored locally
+in SSSD's cache.
+
+SSSD must be running for this API to work and it must be restarted after the
+first override is created so SSSD can start looking into newly created local
+view.
+
+.. code-block:: python
+    :caption: Examples
+
+    @pytest.mark.topology(KnownTopology.LDAP)
+    @pytest.mark.topology(KnownTopology.AD)
+    def test_local_overrides__user(client: Client, provider: GenericProvider):
+        # Add user
+        provider.user("user-1").add(uid=10001, gid=10001, gecos="gecos")
+
+        # SSSD must be running for sss_override to work
+        client.sssd.start()
+
+        # Create local override for the user
+        client.sss_override.user("user-1").add(name="o-user-1", uid=20001, gid=20001, gecos="o-gecos")
+
+        # SSSD must be restarted so newly created view can be applied
+        client.sssd.restart()
+
+        # Check the result
+        result = client.tools.getent.passwd("o-user-1")
+        assert result is not None
+        assert result.name == "o-user-1"
+        assert result.uid == 20001
+        assert result.gid == 20001
+        assert result.gecos == "o-gecos"
--- a/sssd_test_framework/roles/client.py
+++ b/sssd_test_framework/roles/client.py
@@ -7,6 +7,7 @@ from ..topology import SSSDTopologyMark
 from ..utils.automount import AutomountUtils
 from ..utils.ldb import LDBUtils
 from ..utils.local_users import LocalUsersUtils
+from ..utils.sss_override import SSSOverrideUtils
 from ..utils.sssctl import SSSCTLUtils
 from ..utils.sssd import SSSDUtils
 from .base import BaseLinuxRole
@@ -62,6 +63,11 @@ class Client(BaseLinuxRole[ClientHost]):
         self.local: LocalUsersUtils = LocalUsersUtils(self.host)
         """
         Managing local users and groups.
+        """
+
+        self.sss_override: SSSOverrideUtils = SSSOverrideUtils(self.host, self.fs)
+        """
+        Managing local overrides users and groups.
         """
 
     def setup(self) -> None:
--- a/sssd_test_framework/utils/sss_override.py
+++ b/sssd_test_framework/utils/sss_override.py
@@ -1,0 +1,340 @@
+from __future__ import annotations
+
+from pytest_mh import MultihostHost, MultihostUtility
+from pytest_mh.cli import CLIBuilder, CLIBuilderArgs
+from pytest_mh.utils.fs import LinuxFileSystem
+
+__all__ = [
+    "SSSOverrideUtils",
+    "SSSOverrideUser",
+    "SSSOverrideGroup",
+]
+
+
+class SSSOverrideUtils(MultihostUtility[MultihostHost]):
+    """
+    Management of local override users and groups, using sss_override.
+    """
+
+    def __init__(self, host: MultihostHost, fs: LinuxFileSystem) -> None:
+        """
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        :param fs: Linux file system
+        :type fs: LinuxFileSystem
+        """
+        super().__init__(host)
+
+        self.cli: CLIBuilder = CLIBuilder(host.ssh)
+        self.fs: LinuxFileSystem = fs
+
+    def user(self, name: str) -> SSSOverrideUser:
+        """
+        Get local override user object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.LDAP)
+            def test_example(client: Client, ldap: LDAP):
+                # Add user
+                ldap.user("user-1").add(uid=10001, gid=10001, gecos="gecos")
+
+                # SSSD must be running for sss_override to work
+                client.sssd.start()
+
+                # Create local override for the user
+                client.sss_override.user("user-1").add(name="o-user-1", uid=20001, gid=20001, gecos="o-gecos")
+
+                # SSSD must be restarted so newly created view can be applied
+                client.sssd.restart()
+
+                # Check the result
+                result = client.tools.getent.passwd("o-user-1")
+                assert result is not None
+                assert result.name == "o-user-1"
+                assert result.uid == 20001
+                assert result.gid == 20001
+                assert result.gecos == "o-gecos"
+
+        :param name: User.
+        :type name: str
+        :return: New user local override object.
+        :rtype: SSSOverrideUser
+        """
+        return SSSOverrideUser(self, name)
+
+    def group(self, group: str) -> SSSOverrideGroup:
+        """
+        Get local override group object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.LDAP)
+            def test_example(client: Client, ldap: LDAP):
+                # Add group
+                ldap.group("group-1").add(gid=10001)
+
+                # SSSD must be running for sss_override to work
+                client.sssd.start()
+
+                # Create local override for the user
+                client.sss_override.group("group-1").add(name="o-group-1", gid=20001)
+
+                # SSSD must be restarted so newly created view can be applied
+                client.sssd.restart()
+
+                # Check the result
+                result = client.tools.getent.group("o-group-1")
+                assert result is not None
+                assert result.name == "o-group-1"
+                assert result.gid == 20001
+
+
+        :param group: Group name.
+        :type group: str
+        :return: New group object.
+        :rtype: SSSOverrideGroup
+        """
+        return SSSOverrideGroup(self, group)
+
+    def export_data(
+        self,
+        *,
+        users: str | None = "/tmp/sss_override_users.bak",
+        groups: str | None = "/tmp/sss_override_groups.bak",
+    ) -> None:
+        """
+        Exports local override data for **all** users and groups.
+
+        :param users: File location where users will be exported, if ``None``
+            then user export is omitted. Defaults to
+            ``/tmp/sss_override_users.bak``.
+        :type users: str | None
+        :param groups: File location where groups will be exported, if ``None``
+            then group export is omitted. Defaults to
+            ``/tmp/sss_override_groups.bak``.
+        :type groups: str | None
+        """
+        if users:
+            self.logger.info(f"Exporting user local overrides data to {users} on {self.host.hostname}")
+            self.fs.backup(users)
+            self.host.ssh.exec(["sss_override", "user-export", users])
+
+        if groups:
+            self.logger.info(f"Exporting group local overrides data to {groups} on {self.host.hostname}")
+            self.fs.backup(groups)
+            self.host.ssh.exec(["sss_override", "group-export", groups])
+
+    def import_data(
+        self,
+        *,
+        users: str | None = "/tmp/sss_override_users.bak",
+        groups: str | None = "/tmp/sss_override_groups.bak",
+    ) -> None:
+        """
+        Import users and groups local override data.
+
+        :param users: File location with user overrides that will be imported,
+            if ``None`` then user import is omitted. Defaults to
+            ``/tmp/sss_override_users.bak``.
+        :type users: str | None
+        :param groups: File location with group overrides that will be imported,
+            if ``None`` then group import is omitted. Defaults to
+            ``/tmp/sss_override_groups.bak``.
+        :type groups: str | None
+        """
+        if users:
+            self.logger.info(f"Importing user local overrides data from {users} on {self.host.hostname}")
+            self.fs.backup(users)
+            self.host.ssh.exec(["sss_override", "user-import", users])
+
+        if groups:
+            self.logger.info(f"Importing group local overrides data from {groups} on {self.host.hostname}")
+            self.fs.backup(groups)
+            self.host.ssh.exec(["sss_override", "group-import", groups])
+
+
+class SSSOverrideUser:
+    """
+    Management of local override for users using sss_override.
+    """
+
+    def __init__(self, util: SSSOverrideUtils, user: str) -> None:
+        """
+        :param util: OverrideUser object.
+        :type util: SSSOverrideUser
+        :param user: User name.
+        :type user: str
+        """
+        self.util: SSSOverrideUtils = util
+        self.user: str = user
+
+    def add(
+        self,
+        *,
+        uid: int | None = None,
+        gid: int | None = None,
+        name: str | None = None,
+        home: str | None = None,
+        gecos: str | None = None,
+        shell: str | None = None,
+        certificate: str | None = None,
+    ) -> SSSOverrideUser:
+        """
+        Create new local override for user.
+
+        :param uid: User id, defaults to None
+        :type uid: int | None, optional
+        :param gid: Primary group id, defaults to None
+        :type gid: int | None, optional
+        :param name: | None, optional
+        :type name: str | None, optional
+        :param home: Home directory, defaults to None
+        :type home: str | None, optional
+        :param gecos: GECOS, defaults to None
+        :type gecos: str | None, optional
+        :param shell: Login shell, defaults to None
+        :type shell: str | None, optional
+        :param certificate: PKI certificate defaults to None
+        :type certificate: str | None, optional
+        :return: Self.
+        :rtype: SSSOverrideUser
+        """
+        args: CLIBuilderArgs = {
+            "name": (self.util.cli.option.VALUE, name),
+            "uid": (self.util.cli.option.VALUE, uid),
+            "gid": (self.util.cli.option.VALUE, gid),
+            "home": (self.util.cli.option.VALUE, home),
+            "gecos": (self.util.cli.option.VALUE, gecos),
+            "shell": (self.util.cli.option.VALUE, shell),
+            "certificate": (self.util.cli.option.VALUE, certificate),
+        }
+
+        self.util.logger.info(f"Creating local override for user {self.user} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "user-add", self.user] + self.util.cli.args(args))
+
+        return self
+
+    def delete(self) -> SSSOverrideUser:
+        """
+        Delete the local override for user.
+        """
+        self.util.logger.info(f"Deleting local override for user {self.user} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "user-del", self.user])
+
+        return self
+
+    def get(self, attrs: list[str] | None = None) -> dict[str, list[str]] | None:
+        """
+        Get local override data for user.
+
+        :param attrs: If set, only requested attributes are returned, defaults to None, returning all attributes
+        :type attrs: list[str] | None, optional
+        :return: Dictionary with attribute name as a key.
+        :rtype: dict[str, list[str]]
+        """
+
+        self.util.logger.info(f"Fetching local override data for user {self.user} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "user-show", self.user])
+        if not output.stdout:
+            return None
+
+        lst = output.stdout.split(":")
+        if len(lst) != 8:
+            raise ValueError(f"Unexpected output: {output.stdout}")
+
+        result = [
+            {
+                "user": lst[0],
+                "name": lst[1],
+                "uid": lst[2],
+                "gid": lst[3],
+                "gecos": lst[4],
+                "home": lst[5],
+                "shell": lst[6],
+                "certificate": lst[7],
+            }
+        ]
+
+        return {k: [str(v)] for k, v in result[0].items() if not attrs or k in attrs}
+
+
+class SSSOverrideGroup:
+    """
+    Management of local override for group.
+    """
+
+    def __init__(self, util: SSSOverrideUtils, group: str) -> None:
+        """
+        :param util: SSSOverrideUtils utility object.
+        :type util: SSSOverrideUtils
+        :param group: Group name.
+        :type group: str
+        """
+        self.util: SSSOverrideUtils = util
+        self.group: str = group
+
+    def add(
+        self,
+        *,
+        name: str | None = None,
+        gid: int | None = None,
+    ) -> SSSOverrideGroup:
+        """
+        Create new local override for group.
+
+        :param name: Group local override name, defaults to None
+        :type name: str | None, required
+        :param gid: Group id, defaults to None
+        :type gid: int | None, optional
+        :return: Self.
+        :rtype: SSSOverrideGroup
+        """
+        args: CLIBuilderArgs = {
+            "name": (self.util.cli.option.VALUE, name),
+            "gid": (self.util.cli.option.VALUE, gid),
+        }
+
+        self.util.logger.info(f"Creating local override for group {self.group} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "group-add", self.group] + self.util.cli.args(args))
+
+        return self
+
+    def delete(self) -> SSSOverrideGroup:
+        """
+        Delete the local override for group.
+        """
+        self.util.logger.info(f"Deleting local override for group {self.group} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "group-del", self.group])
+
+        return self
+
+    def get(self, attrs: list[str] | None = None) -> dict[str, list[str]] | None:
+        """
+        Get local override attributes for group.
+
+        :param attrs: If set, only requested attributes are returned, defaults to None, returning all attributes
+        :type attrs: list[str] | None, optional
+        :return: Dictionary with attribute name as a key.
+        :rtype: dict[str, list[str]]
+        """
+        self.util.logger.info(f"Fetching local override group {self.group} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "group-show", self.group])
+        if not output.stdout:
+            return None
+
+        lst = output.stdout.split(":")
+        if len(lst) != 3:
+            raise ValueError(f"Unexpected output: {output.stdout}")
+
+        result = [
+            {
+                "group": lst[0],
+                "name": lst[1],
+                "gid": lst[2],
+            }
+        ]
+
+        return {k: [str(v)] for k, v in result[0].items() if not attrs or k in attrs}
--- a/docs/guides/index.rst
+++ b/docs/guides/index.rst
@@ -11,5 +11,6 @@ How to guides
    testing-ldap-krb5
    testing-netgroups
    testing-passkey
+   sss_override
    local-users
 


sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+
+        :param name: Group name.
+        :type name: str
+        :return: New group object.
+        :rtype: OverrideGroup
+        """
+        return OverrideGroup(self, name)
+
+
+class OverrideUser(object):
```
jakub-vavra-cz, line 91: I think that implicit inheritance from object is not needed in python 3:
https://www.w3docs.com/snippets/python/why-do-python-classes-inherit-object.html

sssd_test_framework/roles/client.py
```diff
@@ -52,6 +53,11 @@ def __init__(self, *args, **kwargs) -> None:
         Managing local users and groups.
         """
 
+        self.override: OverrideUtils = OverrideUtils(self.host)
+        """
+        Managing overrides users and groups
```
pbrezina, line 58: Add dot at the end.

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+from pytest_mh.ssh import SSHLog
+
+__all__ = [
+    "OverrideUtils",
+    "OverrideUser",
+    "OverrideGroup",
+]
+
+
+class OverrideUtils(MultihostUtility[MultihostHost]):
```
pbrezina, line 17: It  would be better to rename it to LocalOverrideUtils (also use local overrides in comments).

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+
+        self.cli: CLIBuilder = CLIBuilder(host.ssh)
+        self._users: list[str] = []
+        self._groups: list[str] = []
+
+    def teardown(self) -> None:
+        """
+        Teardown
+        """
+        super().teardown()
```
pbrezina, line 42: You need to remove added overrides here.

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+        """
+        Get override user object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
+                # Create override
+                client.override.users('user1').add(name='o-user1', uid=10001, gid=10001, gecos="test")
```
pbrezina, line 54: Use " instead of ', outside of comments this is automatically changed by the black tool.

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+        """
+        Get override group object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
+                # Create override
+                client.override.group('users').add(name='o-users', gid=10001)
```
pbrezina, line 78: Same here.

sssd_test_framework/utils/override.py
pbrezina: That is true, but rest of the code also use object explicitly.

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+        self.util = util
+        self.name = name
+
+    def add(
+        self,
+        *,
+        uid: int | None = None,
+        gid: int | None = None,
+        name: str | None = None,
+        password: str | None = "Secret123",
```
pbrezina, line 112: Password can not be overridden.

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+        certificate: str | None = None,
+    ) -> OverrideUser:
+        """
+        Create new override user.
+
+        :param uid: User id, defaults to None
+        :type uid: int | None, optional
+        :param gid: Primary group id, defaults to None
+        :type gid: int | None, optional
+        :param home: Home directory, defaults to None
```
pbrezina, line 125: Name is missing.

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+        args: CLIBuilderArgs = {
+            "name": (self.util.cli.option.VALUE, name),
+            "uid": (self.util.cli.option.VALUE, uid),
+            "gid": (self.util.cli.option.VALUE, gid),
+            "home": (self.util.cli.option.VALUE, home),
+            "gecos": (self.util.cli.option.VALUE, gecos),
+            "shell": (self.util.cli.option.VALUE, shell),
+            "certificate": (self.util.cli.option.VALUE, certificate), }
+
+        self.util.logger.info(f'Creating override user "{self.name}" on {self.util.host.hostname}')
```
pbrezina, line 145: Creating local user override of...

sssd_test_framework/utils/override.py
```diff
@@ -0,0 +1,375 @@
...
+
+        return self
+
+    def modify(
+        self,
+        *,
+        uid: int | None = None,
+        gid: int | None = None,
+        name: str | None = None,
+        password: str | None = "Secret123",
```
pbrezina, line 158: Password is invalid.

pbrezina (changes requested):
Hi Dan, thank you. It looks good, please see minor comments inside.

pbrezina:
Also please fix CI issues.

sssd_test_framework/utils/override.py
sidecontrol: I tried this, and it failed. sss_override command failed because the SSSD was no longer configured or running.  I'll check again to be sure.

pbrezina: I see. I that case it may be good to make it part of the SSSDUtils to be able to teardown correctly.

sidecontrol:
Ack, wrapping up the tests, https://github.com/SSSD/sssd/pull/6717/files , which should be done in the next couple of days.

sidecontrol:
@pbrezina  ping

sidecontrol:
I ran it, but the checks are still failing. I ran tox locally, I'm new to this so any advise on resolving the check failures is appreciated. 

tox:lint
lint: FAIL ✖ in 34.94 seconds
docs: install_deps> python -I -m pip install -r /home/runner/work/sssd-test-framework/sssd-test-framework/docs/requirements.txt
docs: install_package_deps> python -I -m pip install jc pytest 'pytest-mh>=1.0.2' python-ldap
docs: install_package> python -I -m pip install --force-reinstall --no-deps /home/runner/work/sssd-test-framework/sssd-test-framework/.tox/.tmp/package/4/sssd_test_framework-0.1.dev1+gf7848aa.tar.gz
docs: freeze> python -m pip freeze --all
docs: alabaster==0.7.13,Babel==2.12.1,certifi==2023.5.7,charset-normalizer==3.1.0,colorama==0.4.6,docutils==0.18.1,gevent==22.10.2,greenlet==2.0.2,idna==3.4,imagesize==1.4.1,iniconfig==2.0.0,jc==1.23.2,Jinja2==3.1.2,MarkupSafe==2.1.2,packaging==23.1,parallel-ssh==2.12.0,pip==23.1.2,pluggy==1.0.0,pyasn1==0.5.0,pyasn1-modules==0.3.0,Pygments==2.15.1,pytest==7.3.1,pytest-mh==1.0.3,python-ldap==3.4.3,PyYAML==6.0,requests==2.31.0,ruamel.yaml==0.17.26,ruamel.yaml.clib==0.2.7,setuptools==67.7.2,snowballstemmer==2.2.0,Sphinx==6.2.1,sphinx-rtd-theme==1.2.1,sphinx_design==0.4.1,sphinxcontrib-applehelp==1.0.4,sphinxcontrib-devhelp==1.0.2,sphinxcontrib-htmlhelp==2.0.1,sphinxcontrib-jquery==4.1,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.3,sphinxcontrib-serializinghtml==1.1.5,ssh-python==1.0.0,ssh2-python==1.0.0,sssd-test-framework @ file:///home/runner/work/sssd-test-framework/sssd-test-framework/.tox/.tmp/package/4/sssd_test_framework-0.1.dev1%2Bgf7848aa.tar.gz#sha256=ad9c8df33409cacf91f7bbd355199c0a6caf928125121d12247792a5d68ba82e,urllib3==2.0.2,wheel==0.40.0,xmltodict==0.13.0,zope.event==4.6,zope.interface==6.0
tox:docs
.pkg: _exit> python /opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/pyproject_api/_backend.py True hatchling.build
  py3: OK (26.60=setup[26.04]+cmd[0.56] seconds)
  py310: OK (20.68=setup[20.28]+cmd[0.40] seconds)
  lint: FAIL code 1 (34.94=setup[17.45]+cmd[1.06,0.95,0.24,15.23] seconds)
  docs: OK (40.22=setup[22.71]+cmd[17.51] seconds)
  evaluation failed :( (122.54 seconds)

Error: Process completed with exit code 255.

pbrezina:
In PRCI, click on details. There is arrow in the output to expand individual steps, e.g. "> lint". This is the issue:

```
   lint: commands[0]> flake8 sssd_test_framework tests
  lint: commands[1]> pycodestyle sssd_test_framework tests
  lint: commands[2]> isort --check-only sssd_test_framework tests
  lint: commands[3]> mypy --install-types --non-interactive sssd_test_framework tests
  Collecting types-PyYAML
    Downloading types_PyYAML-6.0.12.10-py3-none-any.whl (14 kB)
  Collecting types-Pygments
    Downloading types_Pygments-2.15.0.1-py3-none-any.whl (16 kB)
  Collecting types-colorama
    Downloading types_colorama-0.4.15.11-py3-none-any.whl (5.6 kB)
  Collecting types-setuptools
    Downloading types_setuptools-67.8.0.0-py3-none-any.whl (44 kB)
       ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 44.6/44.6 kB 2.5 MB/s eta 0:00:00
  Collecting types-docutils (from types-Pygments)
    Downloading types_docutils-0.20.0.1-py3-none-any.whl (17 kB)
  Installing collected packages: types-setuptools, types-PyYAML, types-docutils, types-colorama, types-Pygments
  Successfully installed types-PyYAML-6.0.12.10 types-Pygments-2.15.0.1 types-colorama-0.4.15.11 types-docutils-0.20.0.1 types-setuptools-67.8.0.0
  sssd_test_framework/utils/local_override.py:159: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:214: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:226: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:243: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:260: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:276: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  Installing missing stub packages:
  /home/runner/work/sssd-test-framework/sssd-test-framework/.tox/lint/bin/python -m pip install types-PyYAML types-Pygments types-colorama types-setuptools
  
  
  Found 14 errors in 1 file (checked 36 source files)
  sssd_test_framework/utils/local_override.py:279: error: "None" has no attribute "stdout"  [attr-defined]
  sssd_test_framework/utils/local_override.py:337: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:371: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:384: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:397: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:414: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:429: error: Incompatible types in assignment (expression has type "SSHProcessResult", variable has type "None")  [assignment]
  sssd_test_framework/utils/local_override.py:433: error: "None" has no attribute "stdout"  [attr-defined]
  lint: exit 1 (15.23 seconds) /home/runner/work/sssd-test-framework/sssd-test-framework> mypy --install-types --non-interactive sssd_test_framework tests pid=3078
```

When you run tox locally, it should print it directly, without the need to expand anything.

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,445 @@
...
+    ) -> LocalOverrideUser:
+        """
+        :param file: File name.
+        :type file: str | None, defaults to "backup"
+        :return: Self.
+        :rtype: LocalOverrideUser
+        """
+        self.util.logger.info(f"Exporting local override user to {file} on {self.util.host.hostname}")
+        self.output = self.util.host.ssh.run(
+            f"sss_override user-export {file}", raise_on_error=False, log_level=SSHLog.Error
```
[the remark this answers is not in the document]
pbrezina: You need to backup the file first to make sure that you do not override an existing file. Use https://pytest-mh.readthedocs.io/en/latest/api/pytest_mh.utils.fs.html#pytest_mh.utils.fs.LinuxFileSystem.backup for that.

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,445 @@
...
+
+        :param group: Group name.
+        :type group: str
+        :return: New group object.
+        :rtype: LocalOverrideGroup
+        """
+        return LocalOverrideGroup(self, group)
+
+    @property
+    def users(self):
```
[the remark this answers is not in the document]
sidecontrol: I'm not sure. Ah, it was generated by my IDE.

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,445 @@
...
+
+    def __init__(self, util: LocalOverrideUtils, user: str) -> None:
+        """
+        :param util: OverrideUser object.
+        :type util: LocalOverrideUser
+        :param user: User name.
+        :type user: str
+        """
+        self.util = util
+        self.user = user
```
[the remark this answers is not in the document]
sidecontrol: Name is an extra argument, it's not the same as user.

sssd_test_framework/utils/local_override.py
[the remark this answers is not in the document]
sidecontrol: It's not a configuration file or an existing file. The file will only be there if the test fails to teardown incorrectly from a prior run. Ensuring the file doesn't exist before and then after the test will make more sense.

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,445 @@
...
+            "gecos": (self.util.cli.option.VALUE, gecos),
+            "shell": (self.util.cli.option.VALUE, shell),
+            "certificate": (self.util.cli.option.VALUE, certificate),
+        }
+
+        password = f" && passwd --stdin {self.user}"
+        self.util.logger.info(f"Creating local override user {self.user} on {self.util.host.hostname}")
+        self.output = self.util.host.ssh.run(
+            self.util.cli.command(f"sss_override user-add {self.user}", args),
+            input=password,
```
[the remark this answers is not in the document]
sidecontrol: I'm not sure why it's even there. It's not needed.

sssd_test_framework/utils/local_override.py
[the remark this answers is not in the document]
pbrezina: Yes and know. You're making assumptions about the machine where you're doing changes. Any change that you do must be reverted, this is an important thing to keep in mind. When a test is finished, it must leave the machine in the exact state as it was when the test started. This is to ensure that each test starts with exactly the same environment.

The backup method does what you want - if the file already exist, it restores it. If it does not exist, it removes it.

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,414 @@
...
+        """
+        Get local override user object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
+                # Create override
+                client.local_override.users("user1").add(name="o-user1", uid=10001, gid=10001, gecos="test")
```
pbrezina, line 57: users -> user

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,414 @@
...
+        uid: int | None = None,
+        gid: int | None = None,
+        name: str | None = None,
+        home: str | None = None,
+        gecos: str | None = None,
+        shell: str | None = None,
+        certificate: str | None = None,
+    ) -> LocalOverrideUser:
+        """
+        Create new local override user.
```
pbrezina, line 121: Crete new user local override.

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,414 @@
...
+            "name": (self.util.cli.option.VALUE, name),
+            "uid": (self.util.cli.option.VALUE, uid),
+            "gid": (self.util.cli.option.VALUE, gid),
+            "home": (self.util.cli.option.VALUE, home),
+            "gecos": (self.util.cli.option.VALUE, gecos),
+            "shell": (self.util.cli.option.VALUE, shell),
+            "certificate": (self.util.cli.option.VALUE, certificate),
+        }
+
+        self.util.logger.info(f"Creating local override user {self.user} on {self.util.host.hostname}")
```
pbrezina, line 150: user local override
(or local override for user, whatever is better)

pbrezina: Same on other places.

Because you are not creating new user, you are creating the override. "local override user" suggest that you create new user object.

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,414 @@
...
+        self.group: str = group
+
+    def add(
+        self,
+        *,
+        name: str | None = None,
+        gid: int | None = None,
+    ) -> LocalOverrideGroup:
+        """
+        Create new local override group.
```
pbrezina, line 304: Same here -> group local override or local override for group

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,414 @@
...
+from pytest_mh.cli import CLIBuilder, CLIBuilderArgs
+
+__all__ = [
+    "LocalOverrideUtils",
+    "LocalOverrideUser",
+    "LocalOverrideGroup",
+]
+
+
+class LocalOverrideUtils(MultihostUtility[MultihostHost]):
```
pbrezina, line 13: Perhaps rename all classes to SSSOverride* and file to sss_override.py since it is a wrapper around specific tool? What do you think is better?

sssd_test_framework/utils/local_override.py
```diff
@@ -0,0 +1,414 @@
...
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
+
+        self.cli: CLIBuilder = CLIBuilder(host.ssh)
+        self._users: list[str] = []
+        self._groups: list[str] = []
+
+    def teardown(self) -> None:
```
pbrezina, line 29: I was thinking about this and it might be better to avoid this and rather change SSSDUtils like this:

```diff
diff --git a/sssd_test_framework/utils/sssd.py b/sssd_test_framework/utils/sssd.py
index 6e14029..adfd943 100644
--- a/sssd_test_framework/utils/sssd.py
+++ b/sssd_test_framework/utils/sssd.py
@@ -110,6 +110,11 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         )
         self.svc.reload_daemon()
 
+        # Backup existing SSSD data
+        self.fs.backup("/etc/sssd")
+        self.fs.backup("/var/lib/sss")
+        self.fs.backup("/var/log/sssd")
+
         if self.__load_config:
             self.config_load()
             return
```

This will make sure that all changes in SSSD data is automatically reverted regardless of any tool that we use and we do not have to deal with it individually. And just document it here.

pbrezina (changes requested):

sssd_test_framework/utils/local_override.py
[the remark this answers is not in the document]
pbrezina: What do you mean?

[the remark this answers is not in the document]
sidecontrol: The command is this,

sss_override user-add pbrezina --uid 1234 --gid 1234 --name pavel 

We can change it to username?

sssd_test_framework/utils/local_override.py
sidecontrol: I wouldn't name it reload, but reset. It's confusing because reload when talking system services just means refreshing the configuration without needing to restart the daemon.

sssd_test_framework/utils/local_override.py
sidecontrol: That makes more sense actually.

sssd_test_framework/utils/local_override.py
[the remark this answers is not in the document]
pbrezina: Ah, I see. But it is fine, `self.name` and `name` are two different things. It's perfectly fine to use both, although it may create little confusion, it is better to solve the confusion by making a comment where you introduce local `name`.

If you use `self.username` and `name` or the other way around, imho the confusion is still there (ie which name is which). You could solve it by using `override_name` but that is too long to use nicely.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,420 @@
...
+            "gid": (self.util.cli.option.VALUE, gid),
+            "home": (self.util.cli.option.VALUE, home),
+            "gecos": (self.util.cli.option.VALUE, gecos),
+            "shell": (self.util.cli.option.VALUE, shell),
+            "certificate": (self.util.cli.option.VALUE, certificate),
+        }
+
+        self.util.logger.info(f"Creating local override for user {self.user} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(
+            ["sss_override", "user-add", self.user] + self.util.cli.args(args), raise_on_error=False
```
pbrezina, line 156: Why raise_on_error is False?

```
def test_override(client: Client, provider: GenericProvider):
    provider.user("tuser").add()
    client.sss_override.user("tuser").add(name="o-tuser", uid=50000, gid=60000)
    client.sssd.start()

    result = client.tools.id("o-tuser")
    print(str(result))
```

The test above does not work, because sss_overrride fails with:

```
INFO     2023-06-08 11:48:02,921 Executing command #11
                                 Host: client.test
                                 Shell: /usr/bin/bash -c
                                 User: root
                                 Command: sss_override user-add tuser --name o-tuser --uid 50000 --gid 60000
                                 Input: None
                                 Working directory: None
                                 Extra environment: 
INFO     2023-06-08 11:48:03,441 Previous command completed with exit code 1
                                 Output: 
                                 Error output: |
                                   [sssd] [confdb_expand_app_domains] (0x0010): No domains configured, fatal error!
                                   [sssd] [sss_tool_route] (0x0010): Command user-add does not handle initialization error [1432158246] No domain is enabled
```

But I didn't know it because no exception was risen.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,420 @@
...
+        :type fs: LinuxFileSystem
+        """
+        super().__init__(host)
+
+        self.cli: CLIBuilder = CLIBuilder(host.ssh)
+        self._users: list[str] = []
+        self._groups: list[str] = []
+        self.fs: LinuxFileSystem = fs
+
+    def teardown(self) -> None:
```
pbrezina, line 33: This can be removed. Together with self._users and self._groups. It will be handled automatically by SSSDUtils by https://github.com/SSSD/sssd-test-framework/commit/b50970458306818169a15671edf0887382b7bf9e.

pbrezina (changes requested):
Hi, I implemented  the backup of all SSSD configuration and data, it also required a fix in pytest-mh so please update that as well for your testing.

Other then that, I tried the, code, it is working correctly. Just see some comments inside. Once the remaining comments are fixed, we can push it.

sssd_test_framework/utils/local_override.py
sidecontrol: Yes, Just a thought.

sssd_test_framework/utils/sss_override.py
sidecontrol: I should have removed it. I left 'raise_on_error=False' on the delete function. Right now, the SSSD teardown occurs before this, so this will always fail.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        :param fs: Linux file system
+        :type fs: LinuxFileSystem
+        """
+        super().__init__(host)
+
+        self.cli: CLIBuilder = CLIBuilder(host.ssh)
+        self._users: list[str] = []
+        self._groups: list[str] = []
```
pbrezina, line 30: Remove as it is no longer needed since teardown was removed.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
+                # Create local override
+                # SSSD must be running for this to work
+                client.sss_override.user("user1").add(name="o-user1", uid=10001, gid=10001, gecos="test")
+
+                # Assert the overridden name, uid, gid and gecos
+                assert client.tools.getent.passwd("o-user1").uid == 10001
```
pbrezina, line 47:
```python
result = client.tools.getent.passwd("o-user1")
assert result is not None
assert result.uid == 10001
...
```

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
+                # Create local override
+                # SSSD must be running for this to work
+                client.sss_override.group("users").add(name="o-users", gid=10001)
+
+                # Assert the overridden name and gid
+                assert client.tools.getent.group("o-users").gid == 10001
```
pbrezina, line 72:
```python
result = client.tools.getent.group("o-users")
assert result is not None
assert result.gid == 10001
```

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+            "gid": (self.util.cli.option.VALUE, gid),
+            "home": (self.util.cli.option.VALUE, home),
+            "gecos": (self.util.cli.option.VALUE, gecos),
+            "shell": (self.util.cli.option.VALUE, shell),
+            "certificate": (self.util.cli.option.VALUE, certificate),
+        }
+
+        self.util.logger.info(f"Creating local override for user {self.user} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "user-add", self.user] + self.util.cli.args(args))
+        self.util._users.append(self.user)
```
pbrezina, line 140: Can be removed.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        self.util.host.ssh.exec(["sss_override", "user-del", self.user], raise_on_error=False)
+
+        return self
+
+    def backup(
+        self,
+        file: str = "/tmp/sss_override_user.bak",
+    ) -> SSSOverrideUser:
+        """
+        Exports local override data for **all** users.
```
pbrezina, line 207: Add new line.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        :param file: File name.
+        :type file: str | None, defaults to "/tmp/sss_override_user.bak"
+        :return: Self.
+        :rtype: SSSOverrideUser
+        """
+        self.util.logger.info(f"Exporting local override user data to {file} on {self.util.host.hostname}")
+        self.util.fs.backup(f"{file}")
```
pbrezina, line 214: `self.util.fs.backup(file)`

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        :type attrs: list[str] | None, optional
+        :return: Dictionary with attribute name as a key.
+        :rtype: dict[str, list[str]]
+        """
+
+        self.util.logger.info(f"Fetching local override data for user {self.user} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "user-show", self.user])
+        lst = output.stdout.split(":")
+        if len(lst) < 8:
+            return {}
```
pbrezina, line 248: Shouldn't you raise error here?

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        """
+        args: CLIBuilderArgs = {
+            "name": (self.util.cli.option.VALUE, name),
+            "gid": (self.util.cli.option.VALUE, gid),
+        }
+
+        self.util.logger.info(f"Creating local override for group {self.group} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "group-add", self.group] + self.util.cli.args(args))
+
+        self.util._groups.append(self.group)
```
pbrezina, line 305: Can be removed.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        """
+        Exports local override data for **all** groups.
+
+        :param file: File name.
+        :type file: str | None, defaults to "/tmp/sss_override_group.bak"
+        :return: Self.
+        :rtype: SSSOverrideGroup
+        """
+        self.util.logger.info(f"Exporting local override group data to {file} on {self.util.host.hostname}")
+        self.util.fs.backup(f"{file}")
```
pbrezina, line 356: `self.util.fs.backup(file)`

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        self.util.host.ssh.exec(["sss_override", "group-export", file])
+
+        return self
+
+    def restore(
+        self,
+        file: str = "/tmp/sss_override_group.bak",
+    ) -> SSSOverrideGroup:
+        """
```
pbrezina, line 365: Add description.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        self.util.fs.backup(f"{file}")
+        self.util.host.ssh.exec(["sss_override", "user-export", file])
+
+        return self
+
+    def restore(
+        self,
+        file: str = "/tmp/sss_override_user.bak",
+    ) -> SSSOverrideUser:
+        """
```
pbrezina, line 223: Add description.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,400 @@
...
+        :type attrs: list[str] | None, optional
+        :return: Dictionary with attribute name as a key.
+        :rtype: dict[str, list[str]]
+        """
+        self.util.logger.info(f"Fetching local override group {self.group} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "group-show", self.group])
+
+        lst = output.stdout.split(":")
+        if len(lst) < 3:
+            return {}
```
pbrezina, line 390: Shouldn't you raise error here?

pbrezina (changes requested):
Hi, see few nitpicks inside.

Can you please also add how to guide? E.g. https://tests.sssd.io/en/latest/guides/local-users.html

sssd_test_framework/utils/sss_override.py
pbrezina: I think you should also remove it from delete as it is not called during teardown anymore.

pbrezina:
Also please remove the merge commit and just rebase.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        Get local override group object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
+                # Create local override
+                # SSSD must be running for this to work
```
pbrezina, line 66: Provide working example. That is add call to `client.sssd.start()`

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,404 @@
...
+        self.util.host.ssh.exec(["sss_override", "group-add", self.group] + self.util.cli.args(args))
+
+        return self
+
+    def delete(self) -> SSSOverrideGroup:
+        """
+        Delete the local override for group.
+        """
+        self.util.logger.info(f"Deleting local override for group {self.group} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "group-del", self.group], raise_on_error=False)
```
pbrezina, line 342: Remove raise_on_error=false.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,404 @@
...
+        :rtype: dict[str, list[str]]
+        """
+        self.util.logger.info(f"Fetching local override group {self.group} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "group-show", self.group])
+
+        lst = output.stdout.split(":")
+        try:
+            assert len(lst) == 2
+        except IndexError:
+            print("Unexpected index value, unable to get group data")
```
pbrezina, line 394: Avoid using assert outside test. This should be:

```python
if len(lst) != 2:
    raise ValueError(f"Unexpected output: {output.stdout}")

File: sssd_test_framework/utils/sss_override.py [hunk 38]
```diff
@@ -0,0 +1,404 @@
...
+        :rtype: dict[str, list[str]]
+        """
+
+        self.util.logger.info(f"Fetching local override data for user {self.user} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "user-show", self.user])
+        lst = output.stdout.split(":")
+        try:
+            assert len(lst) == 7
+        except IndexError:
+            print("Unexpected index value, unable to get user data")
```
pbrezina, line 250: Avoid using assert outside test. This should be:

```python
if len(lst) != 7:
    raise ValueError(f"Unexpected output: {output.stdout}")
```

`assert len(lst) == 7` will not even raise `IndexError` but `AssertionError` so that is a dead code. And if you catch the exception youjust print it to stdout but the test would continue. We want to raise exception so the test fails.

docs/guides/sss_override.rst
```diff
@@ -0,0 +1,30 @@
+Local override (sss_override) users and groups
```
pbrezina, line 1: Correct me if I am wrong, but isn't "Local overrides **of** users and groups" better?

docs/guides/sss_override.rst
```diff
@@ -0,0 +1,30 @@
...
+        gecos="user",
+        shell="/bin/bash",
+        password="Secret123"
+    )
+    client.sssd.start()
+    client.sss_override.user("user1").add(name="o-user1")
+    client.sssd.restart()
+
+    assert client.ssh("user1", "Secret123")
+    assert client.ssh("o-user1", "Secret123")
```
pbrezina, line 30: Correct indentation. Always execute the test that you add as an example, because if the documentation will not contain working examples, we are doomed.

pbrezina:
Also please rename the commit subject to "sss_override: add utility to manage local overrides" to follow the "area: what" naming convention.

sssd_test_framework/utils/sss_override.py
sidecontrol: Okay, I understand what I didn't do, I didn't actually raise an exception.

docs/guides/sss_override.rst
sidecontrol: It is, but I thought having the CLI command in the title is helpful, the class name is probably sufficient.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def user(self, name: str) -> SSSOverrideUser:
+        """
+        Get local override user object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
```
pbrezina: I'm sorry, this example still does not work. Please, try to actually run it. The client topology will not work because you need to create a user. So the topology should be for example LDAP and you need to create the user that you override.

```python
def test_override(client: Client, ldap: LDAP):
    # Add user
    ldap.user("user-1").add(uid=10001, gid=10001, gecos="gecos")

    # SSSD must be running for sss_override to work
    client.sssd.start()

    # Create local override for the user
    client.sss_override.user("user-1").add(name="o-user-1", uid=20001, gid=20001, gecos="o-gecos")

    # SSSD must be restarted so newly created view can be applied
    client.sssd.restart()

    # Check the result
    result = client.tools.getent.passwd("o-user-1")
    assert result is not None
    assert result.uid == 20001
    assert result.gid == 20001
    assert result.gecos == "o-gecos"
```

pbrezina: Same for group.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+        :param attrs: If set, only requested attributes are returned, defaults to None, returning all attributes
+        :type attrs: list[str] | None, optional
+        :return: Dictionary with attribute name as a key.
+        :rtype: dict[str, list[str]]
+        """
+
+        self.util.logger.info(f"Fetching local override data for user {self.user} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "user-show", self.user])
+        lst = output.stdout.split(":")
+        if len(lst) != 7:
```
pbrezina: Actually, it should be 8 as you are creating 8 items.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+            raise ValueError(f"Unexpected output: {output.stdout}")
+
+        result = [
+            {
+                "userid": lst[0],
```
pbrezina: Rename to user since it is a name and not an uid. This will make it the same as group.get()

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+class SSSOverrideGroup:
+    """
+    Management of local override for group.
+    """
+
+    def __init__(self, util: SSSOverrideUtils, group: str) -> None:
+        """
+        :param util: LocalOverrideUtils utility object.
```
pbrezina: SSSOverrideUtils

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+        :param attrs: If set, only requested attributes are returned, defaults to None, returning all attributes
+        :type attrs: list[str] | None, optional
+        :return: Dictionary with attribute name as a key.
+        :rtype: dict[str, list[str]]
+        """
+        self.util.logger.info(f"Fetching local override group {self.group} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "group-show", self.group])
+
+        lst = output.stdout.split(":")
+        if len(lst) != 2:
```
pbrezina: Should be 3.

docs/guides/sss_override.rst
```diff
@@ -0,0 +1,30 @@
+Local override of users and groups
+###################################
+
+Class :class:`sssd_test_framework.utils.sss_override.SSSOverrideUtils` provides an
+API to manage local overrides for users and groups. Also known as ID views, instead
+of being stored in IPA/LDAP server the override data is stored locally in SSSD's
+cache.
+
+SSSD must be running for this API to work.
```
pbrezina:
```rst
.. note::

    SSSD must be running for this API to work and it must be restarted after the
    first override is created so SSSD can start looking into newly created local
    view.
```

docs/guides/sss_override.rst
```diff
@@ -0,0 +1,30 @@
...
+    @pytest.mark.topology(KnownTopology.AD)
+    def test_local_overrides__user(client: Client, provider: GenericProvider):
+        provider.user("user1").add(
+        uid=999011,
+        gid=999011,
+        home="/home/user1",
+        gecos="user",
+        shell="/bin/bash",
+        password="Secret123"
+    )
```
pbrezina: Indent correctly.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+        :type file: str | None, defaults to "/tmp/sss_override_user.bak"
+        :return: Self.
+        :rtype: SSSOverrideUser
+        """
+        self.util.logger.info(f"Importing local override user data from {file} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "user-import", file])
+
+        return self
+
+    def get(self, attrs: list[str] | None = None) -> dict:
```
pbrezina: return dict[str, list[str]] | None

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+        Get local override data for user.
+
+        :param attrs: If set, only requested attributes are returned, defaults to None, returning all attributes
+        :type attrs: list[str] | None, optional
+        :return: Dictionary with attribute name as a key.
+        :rtype: dict[str, list[str]]
+        """
+
+        self.util.logger.info(f"Fetching local override data for user {self.user} on {self.util.host.hostname}")
+        output = self.util.host.ssh.exec(["sss_override", "user-show", self.user])
```
pbrezina, line 241: Return none if no override exist.

pbrezina: Same for group.

sssd_test_framework/utils/sss_override.py
pbrezina: Same for group.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+    def delete(self) -> SSSOverrideUser:
+        """
+        Delete the local override for user.
+        """
+        self.util.logger.info(f"Deleting local override for user {self.user} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "user-del", self.user])
+
+        return self
+
+    def backup(
```
pbrezina: Move this to SSSOverrideUtils since it does not operate on single user and rename it to export since this is the operation that is called on sss_override to avoid confusion. Same for restore and groups.

pbrezina: Since import cannot be used as name lets go with export_data and import_data.

sssd_test_framework/utils/sss_override.py
```diff
@@ -0,0 +1,402 @@
...
+            "shell": (self.util.cli.option.VALUE, shell),
+            "certificate": (self.util.cli.option.VALUE, certificate),
+        }
+
+        self.util.logger.info(f"Creating local override for user {self.user} on {self.util.host.hostname}")
+        self.util.host.ssh.exec(["sss_override", "user-add", self.user] + self.util.cli.args(args))
+
+        return self
+
+    def modify(
```
pbrezina: Modify is not supported by sss_override, you would have to combine get/delete/add but it is not worth it. Let's delete it.

pbrezina: Same for group.

pbrezina (changes requested):
Hi Dan, I started testing the code and found numerous issues. Please, try to always run the code that you write to avoid this kind of stuff. I took the liberty to fix all my comments directly in a new commit. If you agree with the changes, I will squash it and we can merge this PR.

https://github.com/SSSD/sssd-test-framework/pull/8/commits/15b80f8f4bd6587245a819620b847f7bd9e14e2c

sidecontrol:
> Hi Dan, I started testing the code and found numerous issues. Please, try to always run the code that you write to avoid this kind of stuff. I took the liberty to fix all my comments directly in a new commit. If you agree with the changes, I will squash it and we can merge this PR.
> 
> [15b80f8](https://github.com/SSSD/sssd-test-framework/commit/15b80f8f4bd6587245a819620b847f7bd9e14e2c)

I'm fine with all your changes, I do have one question though. Isn't it 8 items, because list[0] counts as the first item?

pbrezina (approved):

pbrezina:
> > Hi Dan, I started testing the code and found numerous issues. Please, try to always run the code that you write to avoid this kind of stuff. I took the liberty to fix all my comments directly in a new commit. If you agree with the changes, I will squash it and we can merge this PR.
> > [15b80f8](https://github.com/SSSD/sssd-test-framework/commit/15b80f8f4bd6587245a819620b847f7bd9e14e2c)
> 
> I'm fine with all your changes, I do have one question though. Isn't it 8 items, because list[0] counts as the first item?

Yes

PR #12, opened 2023-05-30
sssctl: cache: remove, list and expire

patriki01:
Methods for removing, making cache expired and listing cache files.

### DIFF
--- a/sssd_test_framework/utils/sssctl.py
+++ b/sssd_test_framework/utils/sssctl.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from pytest_mh import MultihostHost, MultihostUtility
+from pytest_mh.ssh import SSHProcessResult
 from pytest_mh.utils.fs import LinuxFileSystem
 
 __all__ = [
@@ -20,6 +21,8 @@ class SSSCTLUtils(MultihostUtility[MultihostHost]):
 
         self.fs: LinuxFileSystem = fs
         """Filesystem utils."""
+        self.cache_path = "/var/lib/sss/mc"
+        """Path to cache files"""
 
     def passkey_register(
         self,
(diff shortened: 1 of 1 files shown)


sssd_test_framework/utils/sssctl.py
```diff
@@ -90,3 +93,28 @@ def passkey_register(
             )
 
         return result.stdout_lines[-1].strip()
+
+    def cache_expire(self, *args: str) -> SSHProcessResult:
```
pbrezina, line 97: I just added this in another pull request as I needed it to test a netgroup bug. Please, remove this one.

sssd_test_framework/utils/sssctl.py
```diff
@@ -90,3 +93,28 @@ def passkey_register(
...
+        """
+        Run ``sssctl cache-expire`` command.
+
+        :param args: Additional arguments.
+        :type args: str
+        """
+        return self.host.ssh.exec(["sssctl", "cache-expire", *args])
+
+    def cache_ls(self) -> SSHProcessResult:
```
pbrezina, line 106: Why is this needed?

This is not part of sssctl command so it should not be here.

sssd_test_framework/utils/sssctl.py
```diff
@@ -90,3 +93,28 @@ def passkey_register(
...
+        """
+        Run ``ls 'cache_path'`` command.
+        Result output of ``ls`` is stored in result.stdout
+        """
+        return self.host.ssh.exec(["ls", self.cache_path])
+
+    def remove_cache_files(self) -> SSHProcessResult:
```
pbrezina, line 113: Why is this needed?

This is not part of sssctl command so it should not be here.

If anything, there is `cache-remove`, but I'm not sure from the top of my head if it removes memory cache.

pbrezina (changes requested):

sssd_test_framework/utils/sssctl.py
patriki01: It is probably not needed and even if, it shouldnt be there.

PR #13, opened 2023-06-12
sssd: improve setup performance

pbrezina:
Avoid repeating operations that can be done only once:
- systemd override
- backup

### DIFF
--- a/sssd_test_framework/hosts/client.py
+++ b/sssd_test_framework/hosts/client.py
@@ -1,0 +1,138 @@
+"""IPA multihost host."""
+
+from __future__ import annotations
+
+from pytest_mh.ssh import SSHLog
+
+from .base import BaseBackupHost
+
+__all__ = [
+    "ClientHost",
+]
+
+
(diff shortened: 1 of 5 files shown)


sssd_test_framework/hosts/client.py
```diff
@@ -0,0 +1,135 @@
...
+            """
+            set -ex
+
+            backuppath="/tmp/mh.client.systemd.backup"
+            overridedir="/etc/systemd/system/sssd.service.d"
+
+            rm -fr "$overridedir"
+            if [ -d "$overridedir" ] || [ -f "$overridedir" ]; then
+                mv --force "$backuppath" "$overridedir"
+            fi
```
ikerexxe, line 77: Should we reload systemd to apply the changes?

pbrezina: Yes, nice catch.

ikerexxe (approved):
LGTM!

PR #15, opened 2023-06-15
sssctl: user-checks added

patriki01:
Adding `sssctl user-checks name` to framework.

### DIFF
--- a/sssd_test_framework/utils/sssctl.py
+++ b/sssd_test_framework/utils/sssctl.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from pytest_mh import MultihostHost, MultihostUtility
 from pytest_mh.cli import CLIBuilder, CLIBuilderArgs
+from pytest_mh.ssh import SSHProcessResult
 from pytest_mh.utils.fs import LinuxFileSystem
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


sssd_test_framework/utils/sssctl.py
```diff
@@ -171,3 +174,29 @@ def passkey_register(
...
+    class Action(Enum):
+        """
+        Action parameter for ``user_checks``.
+        """
+        AUTH = "auth"
+        ACCT = "acct"
+        SETC = "setc"
+        CHAU = "chau"
+        OPEN = "open"
+        CLOS = "clos"
```
pbrezina, line 187: I think it would be nicer to use  PAM function names instead of the shortcuts from `sssctl user-checks` argument parser, i.e. https://github.com/SSSD/sssd/blob/32f578229d38766b208f33130e28317ca69001d2/src/tools/sssctl/sssctl_user_checks.c#L268-L300

* pam_authenticate -> authenticate
* pam_setcred -> setcred
* ...

pbrezina: But I wonder if enum is a good choice here, as it is useful only for user_checks but has a generic name "Action". This would create a conflict at some point with other command. We can rename it to UserChecksAction but that is long, so maybe we can remove the enumeration and just provide a string to the action argument.

What do you think?

patriki01: Thank you for your feedback. I do not have a strong opinion on this. I think the use of enum here comes in handy especially when the attributes are mostly abbreviations, that are easily typed incorrectly. However, I see your point about potential naming conflicts. So if we are keeping this enum, change of the name is necessary. So I would probably go with UserChecksAction, as you mentioned, although it is a bit long because I cannot think of better name. What do you think?

pbrezina: I think `client.sssctl.user_checks(action="auth")` is better then `client.sssctl.user_checks(action=client.sssctl.UserChecksAction.AUTHENTICATE)`. Enum values would be better if we provide something for general use, but just passing the arguments to make the test case shorter and more intuitive (you want to execute sssctl user-checks command just in python tests so you follow the same arguments) is better choice.

pbrezina: You already know the value of action parameter since you want to test specific command. You are not creating something new, just testing exact command.

pbrezina:
I just changed Pam to PAM, I'll wait for CI to finish and push.

pbrezina (approved):

PR #16, opened 2023-06-27
sssctl: config_check command added

patriki01:
Command `sssctl config-check` added

### DIFF
--- a/sssd_test_framework/utils/sssctl.py
+++ b/sssd_test_framework/utils/sssctl.py
@@ -187,3 +187,21 @@ class SSSCTLUtils(MultihostUtility[MultihostHost]):
         :rtype: SSHProcessResult
         """
         return self.host.ssh.exec(["sssctl", "user-checks", username, "-a", action, "-s", service])
+
+    def config_check(self, config: str | None = None, snippet: str | None = None) -> SSHProcessResult:
+        """
+        Call ``sssctl config-check`` with additional arguments
+
+        :param config: Non default config file, defaults to None
+        :type config: str
+        :param snippet: Non default snippet dir, defaults to None
+        :type snippet: str
+        :return: Result of called command
+        :rtype: SSHProcessResult
+        """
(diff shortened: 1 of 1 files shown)


pbrezina:
If config-check fails, it returns non-zero return code and an exception is returned. I think it would be good to set raise_on_error to False and let the caller to assert the return code and output.

pbrezina:
Please, rebase on top of current master.

sssd_test_framework/utils/sssctl.py
```diff
@@ -187,3 +187,21 @@ def user_checks(self, username: str, action: str = "acct", service: str = "syste
...
         """
         return self.host.ssh.exec(["sssctl", "user-checks", username, "-a", action, "-s", service])
+
+    def config_check(self, config: str | None = None, snippet: str | None = None) -> SSHProcessResult:
+        """
+        Call ``sssctl config-check`` with additional arguments
+
+        :param config: Non default config file. Defaults to None
+        :type config: str
+        :param snippet: Non default snippet dir. Deafaults to None
```
pbrezina, line 197: Non default snippet dir, defaults to None

sssd_test_framework/utils/sssctl.py
```diff
@@ -187,3 +187,21 @@ def user_checks(self, username: str, action: str = "acct", service: str = "syste
         :rtype: SSHProcessResult
         """
         return self.host.ssh.exec(["sssctl", "user-checks", username, "-a", action, "-s", service])
+
+    def config_check(self, config: str | None = None, snippet: str | None = None) -> SSHProcessResult:
+        """
+        Call ``sssctl config-check`` with additional arguments
+
+        :param config: Non default config file. Defaults to None
```
pbrezina, line 195: Non default config file, defaults to None

pbrezina (changes requested):

pbrezina (approved):

PR #14, opened 2023-06-14
Adding service parameter to GetentUtils

sidecontrol:

PR #17, opened 2023-06-28
tools: getent.passwd() takes additional arguments

patriki01:
I am coverting some test and need `getent passwd --service $service $user`.  So I added it to getent.passwd(), but I wonder if I should add it to the other getents as well.
If so, I will wait for review and then just copy&paste it to the other getents as well.


pbrezina:
Thank you. There is already https://github.com/SSSD/sssd-test-framework/pull/14 that implements precisely this.

PR #20, opened 2023-07-04
samba: add nis-domain if gid is added to group

pbrezina:
--nis-domain is required if group is created with POSIX gid, otherwise
we get:

```
Command #9 exited with return code 255:
  Command:
    samba-tool group add tgroup --gid-number 20001 --group-scope Global --group-type Security
  Error output:
    ERROR: Both --gid-number and --nis-domain have to be set for a RFC2307-enabled group. Operation cancelled.
```

The value is just stored in msSFU30NisDomain attribute of the group
which is not used by SSSD and does not have any meaning to use or samba.
Therefore we can just hard code the value instead of making it
configurable.

### DIFF
--- a/sssd_test_framework/roles/samba.py
+++ b/sssd_test_framework/roles/samba.py
@@ -564,6 +564,12 @@ class SambaGroup(SambaObject):
             "group-type": (self.cli.option.VALUE, category),
         }
 
+        # NIS Domain is required by samba-tool if gid number is set.
+        # It is stored in msSFU30NisDomain attribute of the group which is not
+        # used by SSSD so we can just provide hard coded value.
+        if gid is not None:
+            attrs["nis-domain"] = (self.cli.option.VALUE, "samba")
+
         self._add(attrs)
         return self
 


pbrezina:
Simple reproducer:

```python
from __future__ import annotations

import pytest
from sssd_test_framework.roles.client import Client
from sssd_test_framework.roles.generic import GenericProvider
from sssd_test_framework.topology import KnownTopologyGroup

@pytest.mark.topology(KnownTopologyGroup.AnyProvider)
def test_samba_id(client: Client, provider: GenericProvider):
    u = provider.user("tuser").add(uid=10000, gid=20000)
    g = provider.group("tgroup").add(gid=20001).add_member(u)
    input()

    client.sssd.domain["ldap_id_mapping"] = "false"
    client.sssd.start()

    result = client.tools.id("tuser")
    assert result is not None
    assert result.user.name == "tuser"
    assert result.user.id == 10000
    assert result.group.id == 20000
    assert result.memberof("tgroup")
```

patriki01 (approved):
Looks good. :+1:

ikerexxe:
> Kind of. It fixes issues with Patrik's tests.

Perfect then! :smile:

PR #21, opened 2023-07-10
DBUS: Add support for D-Bus

aplopez:
Provides a set of objects able to handle D-Bus operations: method invocation, property access and node navigation.

All this is automatically done by using introspection to detect the structure.

The back-end uses dbus-send as it must be executed through ssh.

Some example tests can be seen [here](https://github.com/aplopez/sssd/blob/dbus/src/tests/system/tests/test_infopipe.py). Once this PR approved, a new PR will be opened to merge these tests.

Documentation is part of this PR.

Resolves: https://issues.redhat.com/browse/SSSD-6323

### DIFF
--- a/docs/guides/testing-dbus.rst
+++ b/docs/guides/testing-dbus.rst
@@ -1,0 +1,195 @@
+Testing D-Bus Services
+======================
+Infopipe
+--------
+Access to the Infopipe services is achieved through the
+:class:`sssd_test_framework.roles.client.Client` class. It provides the
+:meth:`sssd_test_framework.roles.client.Client.infopipe()` method that
+returns a pre-instantiated object of class
+:class:`sssd_test_framework.utils.sbus.DBUSDestination`, which gives you
+direct access to the `Infopipe` destination.
+
+With the destination, it is possible to get the different objects it provides.
+Each object has its own ``object path``. These objects are obtained by calling
+the destination's :meth:`sssd_test_framework.utils.sbus.DBUSDestination.getObject()`
+method. The object path must be passed as argument. A list containing all the
+available object paths can be retrieved with the method
+:meth:`sssd_test_framework.utils.sbus.DBUSDestination.getObjectPaths()`.
+
+.. note::
+    Passing an ``object path`` that doesn't exist will not generate an error nor
+    raise an exception. It will, instead, create a valid object with no attributes.
+    This is a strange behavior of ``Introspection``.
+
+Once you have one of these objects, you can access other D-BUS sub-objects
+provided by that destination by simply treating them as attributes of the
+parent object. **Methods** and **properties** are accessed in the same way.
+
+.. note::
+    Accessing an attribute that can't be mapped to a subobject, attribute or
+    method, will raise an :class:`AttributeError` exception.
+
+In the following example, there is an LDAP domain called `test`.
+
+.. code-block:: python
+    :caption: Infopipe Example
+
+    @pytest.mark.topology(KnownTopology.LDAP)
+    def test_infopipe__GetUserProperties(client: Client, provider: GenericProvider):
+        provider.user("user-1").add(uid=10001, gid=20001)
+
+        client.sssd.start()
+
+        users = client.ifp.getObject("/org/freedesktop/sssd/infopipe/Users");
+        user_path = users.FindByName("user-1")
+        assert "/org/freedesktop/sssd/infopipe/Users/test/10001" == user_path
+
+        user = client.ifp.getObject(user_path)
+        uid = user.Get("org.freedesktop.sssd.infopipe.Users.User", "uidNumber")
+        assert uid == 10001
+
+        gid = user.gidNumber
+        assert gid == 20001
+
+        props = user.GetAll("org.freedesktop.sssd.infopipe.Users.User")
+        assert props["uidNumber"] == 10001
+        assert props["gidNumber"] == 20001
+        assert props["name"] == "user-1"
+        assert props["homeDirectory"] == "/home/user-1"
+
+Internals
+---------
+Although :class:`sssd_test_framework.roles.client.Client` provides the
+:meth:`sssd_test_framework.roles.client.Client.infopipe()` method, the framework
+allows you to access any D-Bus destination. To achieve that you need to
+instantiate the :class:`sssd_test_framework.utils.sbus.DBUSDestination` class,
+providing the MultihostHost where the D-Bus services are running, the destination
+(the application) to contact and the bus to connect to.
+
+The bus can be the bus' path as a string, or one of the predefined buses:
+
+* :attr:`sssd_test_framework.utils.sbus.DBUSKnownBus.SYSTEM`
+* :attr:`sssd_test_framework.utils.sbus.DBUSKnownBus.SESSION`
+* :attr:`sssd_test_framework.utils.sbus.DBUSKnownBus.MONITOR`
+
+.. code-block:: python
+    :caption: Accessing the monitor
+
+    @pytest.mark.topology(KnownTopologyGroup.AnyProvider)
+    def test_example__monitor(client: Client):
+        client.sssd.start()
+
+        monitor = DBUSDestination(client.host, dest="sssd.monitor", bus=DBUSKnownBus.MONITOR)
+
+        sssd = monitor.getObject(objpath="/sssd")
+
+        res = sssd.debug_level
+        assert res == 0xFFF0
+
+        sssd.debug_level = 0x0070
+
+
+A Note On Type Conversion
+~~~~~~~~~~~~~~~~~~~~~~~~~
+All methods and properties accept and return Python types. Internally they are
+converted to some specific classes helping to treat them and map them to D-Bus
+types.
+
+Objects of these types represent values that are passed to/from the methods and
+properties. Their type is given by the class. For instance, DBUSTypeString is a
+D-Bus string.
+
+These classes are all subclasses of the abstract class
+:class:`sssd_test_framework.utils.dbus.types.DBUSType` and they provide the
+following methods:
+
+* :attr:`sssd_test_framework.utils.dbus.types.DBUSType.value`: A property to read
+  and set the (Python) value.
+* :meth:`sssd_test_framework.utils.dbus.types.DBUSType.mimic()`: a method to copy
+  itself without copying the value while maintaining the structure (subtypes).
+* :meth:`sssd_test_framework.utils.dbus.types.DBUSType.param()`: The string
+  representation of the value in a format suitable to be used as a parameter for
+  ``dbus-send``.
+* :meth:`sssd_test_framework.utils.dbus.types.DBUSType.parse()`: Parses the
+  string resulting from an execution of ``dbus-send`` and set the value to the
+  object.
+
+Basic types (integers, strings, booleans, etc.) are subclasses of the
+:class:`sssd_test_framework.utils.dbus.types.DBUSTypeBasic` abstract class.
+
+Container types -- that is, the subclasses of the abstract class
+:class:`sssd_test_framework.utils.dbus.types.DBUSTypeContainer` -- accept a
+parameter for their constructors, another
+:class:`sssd_test_framework.utils.dbus.types.DBUSType` object of the expected type.
+
+.. code-block:: python
+    :caption: Declaring an array of unit32
+
+    s = "array [ uint32 1 uint32 2 uint32 3 ]"
+    a = DBUSTypeArray(DBUSTypeUInt32())
+    a.parse(DBUSResult(s))
+    print(a.value)
+    [1, 2, 3]
+
+In some cases it may not be possible to know in advance the type of the elements
+of a container type. In that case, no object is passed to the constructor. The
+type will be guessed while parsing the result from ``dbus-send``.
+
+.. note::
+    ``dbus-send`` doesn't explain very well how container types are combined as
+    parameters, and so far we didn't use them. So we might have to adapt the
+    results of param() if they are ever used.
+
+.. note::
+    Using instrospection it is possible to get the methods and properties
+    signatures. Nevertheless, the signature for ``variant`` types does not include
+    the type of the contained type, as it does for the arrays and dictionaries.
+    Because of this, it is not possible to know in advance which type to expect
+    and will have to be guessed while parsing.
+
+Implemented Types
+~~~~~~~~~~~~~~~~~
+The following classes are already implemented.
+
+* :class:`sssd_test_framework.utils.dbus.types.DBUSType`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeBoolean`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeString`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeObjectPath`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeInteger`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeByte`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeInt16`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeInt32`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeInt64`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeUInt16`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeUInt32`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeUInt64`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeDouble`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeContainer`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeArray`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeDict`
+* :class:`sssd_test_framework.utils.dbus.types.DBUSTypeVariant`
+
+.. note::
+    Although the `D-Bus specification`_ considers ``dict entry`` a separate type,
+    we didn't implement it as such because there is no use case for it outside
+    of an array, in which case the array becomes a dictionary.
+
+.. _D-Bus specification: https://dbus.freedesktop.org/doc/dbus-specification.html#type-system
+
+Not Implemented Types
+~~~~~~~~~~~~~~~~~~~~~
+Some other classes were not implemented because they are not accepted by
+``dbus-send``:
+
+* signature
+* UNIX FD
+* struct
+
+Helper Classes
+~~~~~~~~~~~~~~
+Class :class:`sssd_test_framework.utils.dbus.types.DBUSSignatureReader` provides
+a single class method
+:meth:`sssd_test_framework.utils.dbus.types.DBUSSignatureReader.read()`
+used to read a method or property signature from a string and generate the
+corresponding :class:`sssd_test_framework.utils.dbus.types.DBUSType` objects
+required for the method or property.
--- a/sssd_test_framework/roles/client.py
+++ b/sssd_test_framework/roles/client.py
@@ -9,6 +9,7 @@ from ..topology import SSSDTopologyMark
 from ..utils.automount import AutomountUtils
 from ..utils.ldb import LDBUtils
 from ..utils.local_users import LocalUsersUtils
+from ..utils.sbus import DBUSDestination, DBUSKnownBus
 from ..utils.sss_override import SSSOverrideUtils
 from ..utils.sssctl import SSSCTLUtils
 from ..utils.sssd import SSSDUtils
@@ -70,6 +71,13 @@ class Client(BaseLinuxRole[ClientHost]):
         self.sss_override: SSSOverrideUtils = SSSOverrideUtils(self.host, self.fs)
         """
         Managing local overrides users and groups.
+        """
+
+        self.ifp: DBUSDestination = DBUSDestination(
+            self.host, dest="org.freedesktop.sssd.infopipe", bus=DBUSKnownBus.SYSTEM
+        )
+        """
+        The D-bus destination for infopipe.
         """
 
     def setup(self) -> None:
--- a/sssd_test_framework/utils/sbus.py
+++ b/sssd_test_framework/utils/sbus.py
@@ -1,0 +1,450 @@
+"""Manage D-Bus operations"""
+
+from __future__ import annotations
+
+import xml.etree.ElementTree as xml
+from abc import ABC
+from typing import Any, Final
+
+from pytest_mh import MultihostHost, MultihostUtility
+from pytest_mh.ssh import SSHProcessResult
+
+from .dbus.types import DBUSResult, DBUSSignatureReader, DBUSType, DBUSTypeString, DBUSTypeVariant
+
+__all__ = [
+    "DBUSDestination",
+    "DBUSKnownBus",
+]
+
+
+class DBUSKnownBus(str):
+    """
+    Weel-known D-Bus buses.
+    """
+
+    SYSTEM: Final = "--SYSTEM--"
+    """The system bus"""
+
+    SESSION: Final = "--SESSION--"
+    """The session bus"""
+
+    MONITOR: Final = "unix:path=/var/lib/sss/pipes/private/sbus-monitor"
+    """Monitor's private bus"""
+
+
+class ProxyObject(ABC):
+    """
+    Abstract class representing a D-Bus object (method or property).
+
+    Corresponding concrete classes are provided for their instantiation.
+    """
+
+    def __init__(
+        self,
+        host: MultihostHost,
+        dest: str,
+        objpath: str,
+        bus: str,
+        /,
+        interface: str | None = None,
+        child: xml.Element | None = None,
+    ):
+        """
+        Constructor to be called by the subclasses either explicitely or by inheritance.
+        """
+        match bus:
+            case DBUSKnownBus.SYSTEM:
+                _bus_param = "--system"
+            case DBUSKnownBus.SESSION:
+                _bus_param = "--session"
+            case _:
+                _bus_param = "--bus=" + bus
+
+        self.host = host
+        self.dest = dest
+        self.objpath = objpath
+        self.bus = bus
+        self.interface = objpath[1:].replace("/", ".") if interface is None else interface
+        self.child = child
+        self.cmd = f"dbus-send {_bus_param} --print-reply --dest={dest} {self.objpath}"
+
+    def _run(self, command: str, *args):
+        """
+        Run dbus-send over ssh on the remote host.
+        """
+        cmd = self.cmd + f" {command}"
+        for arg in args:
+            cmd += " " + arg.param()
+
+        result = self.host.ssh.run(cmd)
+        if result.stdout_lines[0].startswith("method return"):
+            lines = result.stdout_lines[1:]
+        else:
+            lines = result.stdout_lines
+
+        return SSHProcessResult(result.rc, lines, result.stderr_lines)
+
+
+class ProxyProperty(ProxyObject):
+    """
+    A D-Bus property accessible as an object
+    """
+
+    # interface and child are mandatory for this class
+    def __init__(
+        self,
+        host: MultihostHost,
+        dest: str,
+        objpath: str,
+        bus: str,
+        /,
+        interface: str,
+        child: xml.Element,
+        type: DBUSType,
+    ):
+        super().__init__(host, dest, objpath, bus, interface=interface, child=child)
+        # Properties always return a variant of the provided type.
+        self.type: DBUSTypeVariant = DBUSTypeVariant(type)
+        # access can be "read", "write" or "readwrite"
+        self.readable = "read" in child.attrib["access"]
+        self.writable = "write" in child.attrib["access"]
+        self.arg_interface = DBUSTypeString()
+        self.arg_interface.value = self.interface
+        self.arg_name = DBUSTypeString()
+        self.arg_name.value = self.child.attrib["name"]  # type: ignore[union-attr]
+
+    def get_value(self) -> Any:
+        """
+        Get the property's value
+        """
+        if not self.readable:
+            raise PermissionError("Execution failed: not a readable property")
+
+        res = self._run("org.freedesktop.DBus.Properties.Get", self.arg_interface, self.arg_name)
+        if res.rc != 0:
+            raise RuntimeError(res.stderr)
+
+        ret = self.type.mimic()
+        ret.parse(DBUSResult(res.stdout))
+        return ret.value
+
+    def set_value(self, value) -> None:
+        """
+        Set the property's value
+        """
+        if not self.writable:
+            raise RuntimeError("Execution failed: not a writable property")
+
+        val = self.type.mimic()
+        val.value = value
+
+        res = self._run("org.freedesktop.DBus.Properties.Set", self.arg_interface, self.arg_name, val)
+        if res.rc != 0:
+            raise RuntimeError(res.stderr)
+
+
+class ProxyMethod(ProxyObject):
+    """
+    A D-Bus method accessible as an object
+    """
+
+    def __init__(
+        self,
+        host: MultihostHost,
+        dest: str,
+        objpath: str,
+        bus: str,
+        /,
+        interface: str | None = None,
+        child: xml.Element | None = None,
+        input: list[DBUSType] | None = None,
+        output: list[DBUSType] | None = None,
+    ):
+        super().__init__(host, dest, objpath, bus, interface, child)
+        self.input = input
+        self.output = output
+
+    def __call__(self, *args) -> Any:
+        """
+        Execute the method
+        """
+        # In some internal cases we don't provide the child nor the interface at
+        # at instantiation because we don't use them, but on normal calls they are
+        # required.
+        if self.child is None:
+            raise RuntimeError("Execution failed: no child.")
+        if self.interface is None:
+            raise RuntimeError("Execution failed: no interface.")
+
+        objargs = []
+        if self.input is not None:
+            for i in range(len(self.input)):
+                obj = self.input[i].mimic()
+                try:
+                    obj.value = args[i]
+                except IndexError:
+                    raise RuntimeError(
+                        f"Execution failed: {len(self.input)} " f"arguments required but only {i} provided."
+                    )
+                objargs.append(obj)
+
+        method = f'{self.interface}.{self.child.attrib["name"]}'
+        res = self._run(method, *tuple(objargs))
+        if res.rc != 0:
+            raise RuntimeError(f'Execution of \'{self.child.attrib["name"]}{args}\' failed: ' + res.stderr)
+
+        if self.output is None:
+            ret = None
+        else:
+            values = []
+            result = DBUSResult(res.stdout)
+            for i in range(len(self.output)):
+                obj = self.output[i].mimic()
+                obj.parse(result)
+                values.append(obj.value)
+
+            ret = tuple(values) if len(values) > 1 else values[0]
+
+        return ret
+
+
+class DBUSObject:
+    """
+    A D-Bus object accesible as a Python object.
+
+    Objects of this class represent D-Bus nodes.
+    """
+
+    def __init__(self, host: MultihostHost, *, dest: str, objpath: str, bus: str):
+        self.host = host
+        self._properties: dict[Any, Any] = {}
+        self._objpath = objpath
+        self._dest = dest
+        self._bus = bus
+
+        introspection = ProxyMethod(host, dest, objpath, bus)
+        res = introspection._run("org.freedesktop.DBus.Introspectable.Introspect")
+        if res.rc != 0:
+            raise RuntimeError(f"Instrospection failed for {objpath}")
+
+        xmlstr = res.stdout.removeprefix('   string "').removesuffix('"')
+        self._xml = xml.fromstring(xmlstr)
+
+    def _find_child_node(self, root, name: str):
+        """
+        Find an XML child element of the given name.
+        """
+        for child in root:
+            if child.attrib["name"] == name:
+                return child
+            if child.tag == "interface":
+                res = self._find_child_node(child, name)
+                if res is not None:
+                    return res
+
+        return None
+
+    def _find_interface(self, elem: xml.Element) -> xml.Element | None:
+        """
+        Find the interface the provided XML element belongs to.
+        """
+        for child in self._xml:
+            if child.tag == "interface":
+                found = self._find_child_node(child, elem.attrib["name"])
+                if found == elem:
+                    return child
+
+        return None
+
+    def _get_method_args(self, method: xml.Element, dir: str) -> list[DBUSType]:
+        """
+        Get a list of DBUSType objects produced from the given method's signature.
+
+        The direction must be "in" or "out."
+        """
+        params = []
+        for arg in method:
+            if arg.tag == "arg" and arg.attrib["direction"] == dir:
+                params.append(DBUSSignatureReader.read(arg.attrib["type"]))
+        return params
+
+    def _get_property_type(self, prop: xml.Element) -> DBUSType:
+        return DBUSSignatureReader.read(prop.attrib["type"])
+
+    def _make_prop(self, child: xml.Element) -> ProxyProperty:
+        """
+        Add a D-Bus proxy property to the current node.
+        """
+        iface = self._find_interface(child)
+        if iface is None:
+            raise RuntimeError(f"No interface found for {child.attrib['name']}")
+
+        type = self._get_property_type(child)
+
+        prop = ProxyProperty(
+            self.host, self._dest, self._objpath, self._bus, interface=iface.attrib["name"], child=child, type=type
+        )
+        self._properties[child.attrib["name"]] = prop
+        return prop.get_value()
+
+    def _make_method(self, child) -> ProxyMethod:
+        """
+        Add a D-Bus proxy method to the current node.
+        """
+        iface = self._find_interface(child)
+        if iface is None:
+            raise RuntimeError(f"No interface found for {child.attrib['name']}")
+
+        input = self._get_method_args(child, "in")
+        output = self._get_method_args(child, "out")
+
+        method = ProxyMethod(
+            self.host,
+            self._dest,
+            self._objpath,
+            self._bus,
+            interface=iface.attrib["name"],
+            child=child,
+            input=input,
+            output=output,
+        )
+        self.__setattr__(child.attrib["name"], method)
+        return method
+
+    def _make_node(self, child) -> DBUSObject:
+        """
+        Add a D-Bus child node to the current node.
+        """
+        node = DBUSObject(
+            self.host, objpath=self._objpath + "/" + child.attrib["name"], dest=self._dest, bus=self._bus
+        )
+        self.__setattr__(child.attrib["name"], node)
+        return node
+
+    def _make_attr(self, child):
+        """
+        Make a D-Bus attribute and add it to the current node.
+        """
+        match child.tag:
+            case "property":
+                obj = self._make_prop(child)
+            case "method":
+                obj = self._make_method(child)
+            case "node":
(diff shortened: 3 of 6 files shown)


ikerexxe (approved):
LGTM. It was a complex review but I was able to learn some things about python. Thank you!

aplopez:
I made some minor changes mainly improving the docstrings.
Also moved the `Variant` class out of the DBUSObject class.

ikerexxe (approved):

pbrezina:
Hi, I didn't go through it thoroughly, just the base concept. Great job! I love it.

If I read the code correctly, I see two things missing:
* it does not allow to pass different object path, this is needed. E.g. `ifp.Users.User.UpdateGroupsList(..., object_path=...)`
* it can not listen to a signal (which can be added later)

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,459 @@
...
+    Types of predefined D-Bus buses.
+    """
+
+    SYSTEM = auto()
+    """The system bus"""
+    SESSION = auto()
+    """The session bus"""
+
+
+class Variant:
```
pbrezina, line 32: DBUSTypeVariant?

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,459 @@
...
+            self.objpath = objpath
+            self.bus = bus
+            self.interface = objpath[1:].replace("/", ".") if interface is None else interface
+            self.child = child
+            self.cmd = f"dbus-send {_bus_param} --print-reply --dest={dest} {self.objpath}"
+
+        @classmethod
+        def _get_dbus_type(cls, var):
+            """
+            Produce a basic python type from a text string describing a basic D-Bus type.
```
pbrezina, line 87: This method does the opposite - produce string from python type.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,459 @@
...
+        Abstract class representing a D-Bus attribute (method or property).
+
+        Corresponding concrete classes are provided for their instantiation.
+        """
+
+        def __init__(
+            self,
+            host: MultihostHost,
+            dest: str,
+            objpath: str,
```
pbrezina, line 62: default_object_path as it should be able to pass different path to methods

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,459 @@
...
+            bus: str | DBUSBus,
+            /,
+            interface: str | None = None,
+            child: xml.Element | None = None,
+        ):
```
pbrezina, line 67: Missing docstring.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+class DBUSObject(MultihostUtility[MultihostHost]):
+    """
+    D-Bus access.
+
+    Objects of this class represent the root or child D-Bus nodes.
+    """
+
+    class ProxyAttribute(ABC):
```
pbrezina, line 51: attribute kind of conflits with property, how about ProxyObject/ProxyBase?

pbrezina (changes requested):
Hi, here is my initial review. I will continue when these comments are resolved, as they might change the code a bit.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,459 @@
...
+            # dbus-send doesn't return "dict {...}" but "array [ dict entry (...) ... ]"
+            res = re.match("^ *array +\\[( *dict entry +.+)\\] *$", text)
+            if res is not None:
+                return cls._get_python_dict(res.group(1)), res.span()[1]
+
+            res = re.match("^ *array +\\[(.+)\\] *$", text)
+            if res is not None:
+                return cls._get_python_list(res.group(1)), res.span()[1]
+
+            raise TypeError(f"Unhandled D-Bus type {text}")
```
pbrezina, line 200: Move these to standalone classes DBUSTypeArray, DBUSTypeDict, DBUSTypeScalar with parser/builder/getter/setter?

aplopez:
> * it does not allow to pass different object path, this is needed. E.g. `ifp.Users.User.UpdateGroupsList(..., object_path=...)`

I'm not sure I understand you. I don't see why a method taking an object class as parameter would not work. Or do you mean using a different object path for the method call itself? What would be the use case for this? (Sorry, I see no `ifp.Users.User.UpdateGroupList()` method with `d-feet`).

> * it can not listen to a signal (which can be added later)
Right. I didn't though it was needed. I can add it in a second step to reduce the impact on this PR.

sssd_test_framework/utils/sbus.py
aplopez: Of course. I didn't think of renaming the type after taking it out f the `DBUSObject` class.

sssd_test_framework/utils/sbus.py
aplopez: Gosh! Same thing for `_get_dbus_param()`. Fixed both.

sssd_test_framework/utils/sbus.py
aplopez: Changed it to ProxyObject.

pbrezina:
> > * it does not allow to pass different object path, this is needed. E.g. `ifp.Users.User.UpdateGroupsList(..., object_path=...)`
> 
> I'm not sure I understand you. I don't see why a method taking an object class as parameter would not work. Or do you mean using a different object path for the method call itself? What would be the use case for this? (Sorry, I see no `ifp.Users.User.UpdateGroupList()` method with `d-feet`).

The latter. DBus method is called like `method(destination, object_path, args)`. The object path is important concept, because interfaces are registered on object paths.

You will get different introspection result for different object paths, like introspect("/org/freedesktop/sssd/infopipe") and introspect("/org/freedesktop/sssd/infopipe/Users").

See ifp_register_sbus_interface(), it first defines the interfaces that assigns them to particular object paths at the end of the function with sbus_connection_add_path_map().


> 
> > * it can not listen to a signal (which can be added later)
> 
> Right. I didn't though it was needed. I can add it in a second step to reduce the impact on this PR.

Yes, it is not important for IFP since we don't use signal there at this moment, so it can wait.

sssd_test_framework/utils/sbus.py
aplopez: Done. I also improved the parsing a bit.

aplopez:
> The latter. DBus method is called like `method(destination, object_path, args)`. The object path is important concept, because interfaces are registered on object paths.

I think I see where the confusion lies.

In the current implementation the objects already have the destination and object path. You don't  need to specify them. For instance, if you want to call

`ListByName(destination='org.freedesktop.sssd.infopipe', object_path='/org/freedesktop/sddd/infopipe/Users', 'allopez', 0)`

you just need to instantiate the destination objects and dive the `object_path` as attributes:

```
ifp = client.sbus().ifp
ifp.Users.ListByName('allopez', 0)
```

In the same way, to call the method `IsOnline()` on the object path '/org/freedesktop/sddd/infopipe/Domains/test' you can use:

`ifp.Domains.test.IsOnline()`

Is this what you were talking about?

aplopez:
Added a link to the documentation in its current state to the description.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,672 @@
...
+            self.child = child
+            self.cmd = f"dbus-send {_bus_param} --print-reply --dest={dest} {self.objpath}"
+
+        def _run(self, command: str, *args):
+            """
+            Run dbus-send over ssh on the remote host.
+            """
+            cmd = self.cmd + f" {command}"
+            for param in args:
+                cmd += " " + DBUSType.get_dbus_param(param)
```
pbrezina, line 422: The parameter type information is present in the introspection, we need to make sure we send what is expected. This is especially important for numbers since it is not possible to differentiate between uint32, uint64, ...

Same for variant, you can't deduce that you want to pass string as variant.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,672 @@
...
+
+
+class DBUSObject(MultihostUtility[MultihostHost]):
+    """
+    D-Bus access.
+
+    Objects of this class represent the root or child D-Bus nodes.
+    """
+
+    class ProxyObject(ABC):
```
pbrezina, line 380: It would be good to avoid class nesting in this case. It makes the code harder to read and there is no use for it.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,672 @@
...
+
+            rslt = self.host.ssh.run(cmd)
+            if rslt.stdout_lines[0].startswith("method return"):
+                lines = rslt.stdout_lines[1:]
+            else:
+                lines = rslt.stdout_lines
+
+            return SSHProcessResult(rslt.rc, lines, rslt.stderr_lines)
+
+    class ProxyProperty(ProxyObject):
```
pbrezina, line 432: I like what you did with properties, but it is also little problematic. D-Bus allows to have the same name for property and method or signal. Nothing prevents that.

Let's drop this code for now (keep a copy though), focus on minimizing and improving the rest of this internals. After that we will see if and how can we bring it back. If it won't fit, we will just add it by calling org.freedesktop.dbus.property interface similar way you add ifp now.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,672 @@
...
+            return {}
+
+        if attr in self._properties:
+            return self._properties[attr].get_value()
+
+        c = self._find_child(self._xml, attr)
+        if c is not None:
+            return self._make_attr(c)
+
+        return super().__getattribute__(attr)
```
pbrezina, line 616: Why? You already know that the attribute does not exist so calling __getattribute__ will either raise AttributeError or go into infinite recursion, not sure which. You want to raise AttributeError here.

pbrezina (changes requested):
Hi, thank you. This is a good work, I see some space for simplification but there some caveats and corner cases that seem to be unhandled.

I still need to think what to do about the object paths, but that does not affect existing code much. What you have currently in _make_node only work if the names of the object path and interface match, it will break otherwise.

Currently, the DBusObject attributes are a mix of properties, nodes and interfaces. Any name collision and any inconsistency in naming will screw up, we need to stick to only one or hide it all under sub-attributes.

Let's work on the simplifications first.

sssd_test_framework/utils/sbus.py
aplopez: The parent class could have the attribute. This is the recommended behavior.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,672 @@
...
+
+    def __getattr__(self, attr) -> Any:
+        """
+        When an acccessing an attribute that doesn't belong to the object,
+        create it if it should exist.
+        """
+
+        # Required for the initialization
+        if attr == "_properties":
+            return {}
```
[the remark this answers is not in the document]
aplopez: IIRC, this was done because of a race condition at initialization in the contructor.

aplopez:
@pbrezina @ikerexxe I reworked everything. I think it is much better now.

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,290 @@
...
+
+The bus to provide to :meth:`sssd_test_framework.utils.sbus.SBUSUtils.getDestination()`
+must be an object of class :class:`sssd_test_framework.utils.sbus.DBUSKnownBus` which
+provides some predefined buses:
+
+* :attr:`sssd_test_framework.utils.sbus.DBUSKnownBus.SYSTEM`
+* :attr:`sssd_test_framework.utils.sbus.DBUSKnownBus.SESSION`
+* :attr:`sssd_test_framework.utils.sbus.DBUSKnownBus.MONITOR`
+
+(If no bus is provided, the system bus is used.)
```
ikerexxe, line 37: I'd rephrase this and add the following note: `The default bus is system.`

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,290 @@
...
+.. code-block:: python
+    :caption: Accessing the destination ``sssd.nss`` on domain ``MyDomain``
+
+    bus = DBUSKnownBus.forDomain("MyDomain")
+    dest = client.sbus().getDestination(dest="sssd.nss", bus=bus)
+
+With the destination, it is possible to get the different objects it provides. Each object
+has its own ``object path``. These objects are obtained by calling the destination's
+:meth:`sssd_test_framework.utils.sbus.DBUSDestination.getObject()` method. The object path
+must be passed as argument. A list conataining all the available object paths can be retrieved
```
ikerexxe, line 52: s/conataining/containing

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,290 @@
...
+with the method :meth:`sssd_test_framework.utils.sbus.DBUSDestination.getObjectPaths()`.
+
+.. note::
+    Passing an ``object path`` that doesn't exist will not generate an error nor
+    raise an exception. It will, instead, create a valid object with no attributes.
+    This is a strange behavior of ``Introspection``.
```
ikerexxe, line 58: What is introspection? Is this some python feature?

ikerexxe (commented):
I don't remember what I reviewed last time so I began from the beginning again. As such, I've started with the documentation to understand the high level operation. Once I'm familiar with it I'll proceed to review the actual code. I've left some comments inline.

docs/guides/testing-dbus.rst
aplopez: Fixed (although not pushed yet)

docs/guides/testing-dbus.rst
aplopez: Fixed (although not pushed yet)

docs/guides/testing-dbus.rst
aplopez: Long story short, introspection is a mechanism provided by D-Bus for objects to describe themselves. The description includes which methods and properties it provides, their signatures, and its children.

docs/guides/testing-dbus.rst
ikerexxe: By note I meant something like `.. note::` as it's being used in other parts of the document

aplopez: Done

docs/guides/testing-dbus.rst
ikerexxe: Thanks!

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    def parse(self, result: DBUSResult) -> None:
+        text = result.text
+        res = re.match(r"^ *boolean +(([Tt][Rr][Uu][Ee])|([Ff][Aa][Ll][Ss][Ee]))", text)
+        if res is None:
+            raise TypeError("Not a boolean")
+
+        result.text = text[res.span()[1] :]
+        self._value = res.group(1).lower() == "true"
+
+    # Param:   boolean:false|true
```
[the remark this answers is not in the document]
aplopez: I put them there on purpose. The code is inherited from the superclass, but that comments allows to have an idea of what the expected values are.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+            self._str = val
+
+        return self._str
+
+    def clone(self):
+        return self.__class__(self._model_key, self._model_value)
+
+    # We may need to handle dict entry as a separate type as that's how D-Bus
+    # considers it.
+    # For the moment, there is no gain is doint it, so we keep this.
```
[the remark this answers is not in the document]
aplopez: You guess right.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+        result.text = result2.text.removeprefix(match2.group(0)).strip()
+        self._value = pdict
+        self._dict = odict
+
+    # Param: dict:<key type>:<value type>:<key>,<value>[,<key>,<value>]...
+    def param(self) -> str:
+        if self._model_key is None:
+            raise ValueError("No subtype set")
+
+        # We need to do this ridicuouls thing to keep mypy and pycodestyle happy.
```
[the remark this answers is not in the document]
aplopez: Although here it is a false positive, a real positive could be problematic for the two ignored problems (attr-defined, union-attr).
But we could ignore in general the rules about the length of the lines... Not sure it is worth.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,476 @@
...
+    SESSION: Final = "--SESSION--"
+    """The session bus"""
+
+    MONITOR: Final = "unix:path=/var/lib/sss/pipes/private/sbus-monitor"
+    """Monitor's private bus"""
+
+    @classmethod
+    def forDomain(cls, name: str):
+        """Produce the bus for the given domain"""
+        return cls(f"unix:path=/var/lib/sss/pipes/private/sbus-dp_{name}")
```
pbrezina, line 43: It's tempting to add this here, but this is for internal communication with SSSD, so not something we should be testing in system tests. The system tests are black box tests, we should not expect any implementation details. We can only tests stuff that is visible to the user (only the infopipe currently).

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,476 @@
...
+                obj = self.output[i].clone()
+                obj.parse(result)
+                values.append(obj.value)
+
+            ret = tuple(values) if len(values) > 1 else values[0]
+
+        return ret
+
+
+class DBUSObject(MultihostUtility[MultihostHost]):
```
pbrezina, line 222: MultihostUtility should only be the top level class that you add to the role object.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,476 @@
...
+        return super().__setattr__(attr, value)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(*args):
+        pass
+
+
+class DBUSDestination:
```
pbrezina, line 392: This should be MultihostUtility.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,476 @@
...
+from __future__ import annotations
+
+import xml.etree.ElementTree as xml
+from abc import ABC
+from typing import Any, Final
+
+from pytest_mh import MultihostHost, MultihostUtility
+from pytest_mh.ssh import SSHProcessResult
+
+from sssd_test_framework.utils.dbus_types import (
```
pbrezina, line 12: Can you move dbus_types to a subfolder so it is clear that it is not utility itself? And also, please use relative import for our modules. I.e. `from .dbus.dbus_types import ...`

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,476 @@
...
+
+    def _run(self, command: str, *args):
+        """
+        Run dbus-send over ssh on the remote host.
+        """
+        cmd = self.cmd + f" {command}"
+        for arg in args:
+            cmd += " " + arg.param()
+
+        rslt = self.host.ssh.run(cmd)
```
pbrezina, line 90: Please, use `result` instead of `rslt`.

This call will automatically raise an error with all the information you can get if rc != 0. So you don't have to handler that in callers of this method. You can just return modified stdout instead of modifying the result.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,476 @@
...
+        self.arg_interface.value = self.interface
+        self.arg_name = DBUSTypeString()
+        self.arg_name.value = self.child.attrib["name"]  # type: ignore[union-attr]
+
+    def get_value(self) -> Any:
+        """
+        Get the property's value
+        """
+        if not self.readable:
+            raise RuntimeError("Execution failed: not a readable property")
```
pbrezina, line 132: This is PermissionError.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,476 @@
...
+        return DBUSObject(self._host, dest=self._dest, objpath=objpath, bus=self._bus)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(*args):
+        pass
+
+
+class SBUSUtils:
```
pbrezina, line 456: This is not needed. Add only infopipe to the client role using DBUSDestination directly.

pbrezina (changes requested):
Hi Alejandro, I think this looks good. I have done a thorough review of the dbus types, there are lots of python related comments, but it is good. Please, don't hate me :-)

I need to do thorough review of sbus.py, but I will do it after this round is fixed.

aplopez:
I closed all the old remarks. I know I'm not supposed to do that, but I was getting lost with so many open, and they were obsolete now.

sssd_test_framework/utils/dbus_types.py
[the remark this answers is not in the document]
aplopez: But there is no method. The method is inherited from a parent class.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    """
+    Abstract class to create classes that handle type conversion
+    between D-Bus and Python.
+    """
+
+    # It would be good to define here the following abstract class members:
+    #    _type_prefix
+    #
+    # But there is not such thing as _abstract class members_ in Python and
+    # linters will complain although it will work because the contrete
```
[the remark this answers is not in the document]
aplopez: Fixed

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    @value.setter
+    @abstractmethod
+    def value(self, val):
+        pass
+
+    def __str__(self) -> str:
+        """
+        Produce a string representation for the instantiated type's Python value.
+        """
+        return str(self._value)  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
aplopez: Right.

sssd_test_framework/utils/sbus.py
aplopez: Fixed

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+
+    @classmethod
+    def _guess_next_type(cls, result: DBUSResult) -> DBUSType:
+        """
+        Try to guess which is the next type.
+
+        This is mostly needed for variants whose signature doesn't include the subtype.
+        """
+        for subcls in cls.__subclasses__():
+            try:
```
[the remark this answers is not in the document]
aplopez: The call to `subcls._guess_next_type(result)` can also throw a `TypeError` exception that I want to skip.

I find cleaner the code this way, in particular because the other invocations (like `len()`, `__subclass__()` and `subcls()`), I believe, will not throw this exception. With these changes the code would become:

```
        for subcls in cls.__subclasses__():
            if len(subcls.__subclasses__()) == 0:
                obj = subcls()
                try:
                    obj.parse(DBUSResult(result.text))
                except TypeError:
                    continue
            else:
                try:
                    obj = subcls._guess_next_type(result)
                except TypeError:
                    continue
            return obj

        raise TypeError("Couldn't guess the type")
```

Do you still want me to change it?

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+                if len(subcls.__subclasses__()) == 0:
```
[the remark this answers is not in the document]
aplopez: Done.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+        return self._value
+
+    @value.setter
+    def value(self, val: bool):
+        self._value = bool(val)
+
+    # Parsing:   boolean false|true
+    def parse(self, result: DBUSResult) -> None:
+        text = result.text
+        res = re.match(r"^ *boolean +(([Tt][Rr][Uu][Ee])|([Ff][Aa][Ll][Ss][Ee]))", text)
```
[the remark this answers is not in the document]
aplopez: Geez! I missed the flags.

Fixed.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+        if (model_key is None or model_value is None) and model_key != model_value:
+            raise ValueError("Both arguments or none must be None")
+
+        self._model_key = model_key
+        self._model_value = model_value
+        self._str: str | None = None
+
+    @property
+    def value(self) -> dict:
+        return dict(self._value)
```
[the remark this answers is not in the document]
aplopez: Yes. I just want to create a new dict so that the caller can do whatever they want without messing with the object's internal state.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+
+    @value.setter
+    def value(self, val: dict):
+        if self._model_key is None or self._model_value is None:
+            raise ValueError("No subtype set")
+
+        self._value = dict(val)
```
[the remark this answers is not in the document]
aplopez: It's done on purpose. On one side this will check the passed value is a `dict` as requested, and on the other side it will create a new `dict` so that we don't kept the caller's object which they can modify after it was assigned.

sssd_test_framework/utils/sbus.py
aplopez: Done

sssd_test_framework/utils/sbus.py
aplopez: Done

sssd_test_framework/utils/sbus.py
aplopez: Yep. Forgot to change that when I created this class.

sssd_test_framework/utils/sbus.py
aplopez: Done.

sssd_test_framework/utils/sbus.py
aplopez: Removed

sssd_test_framework/utils/dbus_types.py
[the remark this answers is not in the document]
pbrezina: Ok, keep it as is.

sssd_test_framework/utils/dbus_types.py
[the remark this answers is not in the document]
pbrezina: Ok.

sssd_test_framework/utils/dbus_types.py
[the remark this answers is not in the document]
pbrezina: Ok.

sssd_test_framework/utils/dbus_types.py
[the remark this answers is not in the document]
aplopez: I think I will end up using the class documentation.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    The result from the D-Bus operation executed by dbus-send.
+    It is used to simplify parsing of the result as the string
+    must be modified and have no carriage return.
+    """
+
+    def __init__(self, text: str):
+        self.text = text.replace("\n", "")
+
+
+class DBUSType(ABC):
```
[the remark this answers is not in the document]
aplopez: Nice. I didn't know the existence of generics. Something like the C++ templates.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    """
+    Abstract class to create classes that handle type conversion
+    between D-Bus and Python.
+    """
+
+    # It would be good to define here the following abstract class members:
+    #    _type_prefix
```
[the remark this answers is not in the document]
aplopez: > Abstract class is there to define what HAS to be implemented by inherited classes and to ENFORCE it to avoid bugs. This does not enforce anything, it does not help.

I know very well what an abstract class is. That's why I added that comment explaining why I didn't do it as I would had liked to. The idea of this comment is to let other developers know it when they wonder "why did he do it this way and not this other way?."

> Set it in the constructor.

I considered this, but that would make the variables instance variables instead of class variables.

> Make it an abstract property

Yes, that's what I said in my comment. But I find it overkill to implement all that, considering the alternative solution I used also works and that none of the solutions is perfect. Mine doesn't enforce the class variables, but is simpler to implement and maintain; specially considering that the chances of D-Bus adding a new integer type and us implementing a new subclass of `DBUStypeInt` are very low.

> There is no reason to use class members here and something like "abstract class member" does not exist because it does not make any sense.

Oh. I strongly disagree with this. I see plenty of reasons for it to exist. This  use case is one of them.

> Class member is a class member, it it accessible only via the class name,

And this seem to be the mother of the problem. Python behaves in this strange way. Normally, a class member is accessible through the class and through an object, and it always remains a class member. Python behaves differently and it is not going to change any time soon, so we have no choice but to deal with it.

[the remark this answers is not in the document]
pbrezina: Ok, let's agree to disagree.

> > Set it in the constructor.
> 
> I considered this, but that would make the variables instance variables instead of class variables.
> 
> > Make it an abstract property
> 
> Yes, that's what I said in my comment. But I find it overkill to implement all that, considering the alternative solution I used also works and that none of the solutions is perfect. Mine doesn't enforce the class variables, but is simpler to implement and maintain; specially considering that the chances of D-Bus adding a new integer type and us implementing a new subclass of `DBUStypeInt` are very low.

Ok. I don't agree with you and I don't like it, but feel free to make it the way you feel is right.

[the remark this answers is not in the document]
aplopez: I'm trying to implement the property as you suggested, but I'm facing some problems related to Python. For instance, I can define the property in the ``DBUSType`` class as a concrete inheritable method:
```
@property
def type_prefix(self) -> str:
    return self._type_prefix
```
In this case the variable ``_type_prefix``, which is declared in the concrete classes, will not be recognized by this abstract class and I will have to ignore the error. This is what we are trying to avoid.

Or I can make the method abstract and not use this variable:
```
@property
@abstractmethod
def type_prefix(self) -> str:
    pass
```
And every single concrete class will have to implement this method (which is right) but all the implementations will be identical, replicating the code above on every concrete class.

Is there a third option? Which is the pythonic way of doing this?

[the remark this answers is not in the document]
pbrezina: Hmm, I don't think I understand the problem.

The first approach:
```
@property
def type_prefix(self) -> str:
    return self._type_prefix
```
... is in my opinion still wrong, since you expect that `_type_prefix` is set in the child classes and it is pretty much the same as what you have now. You are creating requirement that is not checked. Python allows this. That's why we use tools like mypy to detect these errors to avoid runtime bugs.

The second approach is correct, but I'm not sure I understand `but all the implementations will be identical, replicating the code above on every concrete class`. Do you mean that every concrete class will have to implement this property but with different content? So you have issue with the fact that every class will contain the following three lines just with different string?

```
@property
def type_prefix(self) -> str
    return "the-concrete-class-prefix"
```

Let me know if I misunderstood you, otherwise this is the correct approach, there is no shortcut. Other way would be to use the constructor. Either way, we get static analyzers return an error if this property is not defined in the concrete classes.

It would be also possible to write your own metaclass and check that this field is set, essentially implementing your own abstract class variable, but that would be overthinking it.

---

But really, if you feel strongly about the current implementation, I am not going to block it.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+
+    def __str__(self) -> str:
+        """
+        Produce a string representation for the instantiated type's Python value.
+        """
+        return str(self._value)  # type: ignore[attr-defined]
+
+    def clone(self):
+        """Clone this object. Structure is deepcopied, but values are not."""
+        return self.__class__()
```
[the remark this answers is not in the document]
pbrezina: Ok, got it. Then I found this name and comment misleading. Maybe use `obj.__class__(...)` directly instead of this method? That would make it clear. Or maybe use different name. I also think that class method would be better., but it is pretty much just different syntax for the same thing. E.g.

```
@classmethod
def Create(cls, *args, **kwargs):
    """
    Create new instance of this class.
    """
    return cls(*args, **kwargs)
```

sssd_test_framework/utils/dbus_types.py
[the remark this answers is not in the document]
aplopez: During the week-end I found the third method I was looking for. It happened to be what you propose in this message.

I was returning in the property the value of the class variable `_type_prefix`, which produced generic code which could be inherited but had the problem of not be accepted by linters. Then I noticed I didn't need that variable any more and each class could return its value directly. Code is not shared but the variable is no longer needed.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    # It will be invoked from the concrete subclasses.
+    # It consumes the beginning of the provided result.
+    def parse(self, result: DBUSResult) -> None:
+        text = result.text
+        res = re.match(self._match_exp, text)  # type: ignore[attr-defined]
+        if res is None:
+            raise TypeError("Invalid integer")
+
+        result.text = text[res.span()[1] :].strip()
+        self._value = int(res.group(1))
```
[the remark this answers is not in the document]
aplopez: Done

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    # This makes that, in practice, this class is concrete because all the
+    # abtract methods declared in DBUSType are implemented, even if they
+    # will only be used from the contrete subclasses.
+    #
+    # We could establish abstract getters and setters for them, but that
+    # would be overkill.
+
+    @property
+    def value(self) -> int:
+        return self._value  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
aplopez: Done

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+    @property
+    def value(self) -> str:
+        return self._value
+
+    @value.setter
+    def value(self, val: str):
+        self._value = str(val)
+
+    def __str__(self) -> str:
+        return f"'{self._value}'"
```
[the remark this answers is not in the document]
aplopez: Please note this version includes single quotes around the string. This is required by `dbus-send`.

sssd_test_framework/utils/dbus_types.py
```diff
@@ -0,0 +1,675 @@
...
+        value.parse(result)
+
+        res = re.match(r"^ *\) *", result.text)
+        if res is None:
+            raise TypeError("Malformed dictionary entry: unfinished expression")
+
+        result.text = result.text.removeprefix(res.group(0)).strip()
+        return key, value
+
+    # Parsing: (without the newlines)
```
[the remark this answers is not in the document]
aplopez: If there is a newline, matching may fail depending on the case.

re.MULTILINE could help in this case, but I still need to remove the new lines as the expressions are consumed. So instead of doing it all at once, I would be doing it bit by bit while processing the expressions. Considering this is for internal use of the framework (users should not call `parse()` themselves), I prefer to keep it like it is.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,443 @@
...
+        return self
+
+    def __exit__(*args):
+        pass
+
+
+class DBUSDestination(MultihostUtility[MultihostHost]):
+    def __init__(self, host: MultihostHost, dest: str, bus: str):
+        """
+        Create a destination object associated to a bus and a name.
```
pbrezina, line 383: Please, provide how to use it example here.

```rst
.. code-block:: python
    :caption: Example

    ... code snippet ...
```

sssd_test_framework/roles/client.py
```diff
@@ -97,6 +98,16 @@ def setup(self) -> None:
 
                 self.sssd.import_domain(domain, role)
 
+    def infopipe(self) -> DBUSDestination:
```
pbrezina, line 101: There is no need to have this as a method and no reason to postpone creation of the object since the constructor does not do anything difficult. Can you rather move it into `__init__`? And maybe name it `ifp` to make it shorter?

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,181 @@
...
+    def test_example__infopipe(client: Client, provider: GenericProvider):
+        provider.user("user-1").add(uid=10001)
+        client.sssd.start()
+
+        with client.infopipe() as ifp:
+            users = ifp.getObject("/org/freedesktop/sssd/infopipe/Users")
+
+            result = users.ListByName("user-1", 0)
+            assert "/org/freedesktop/sssd/infopipe/Users/test/10001" in result
+            assert len(result) == 1
```
pbrezina, line 46: Can you make the example little bit more complicated or providing one more example? Maybe calling "FindByName" to get the object path and then requesting and asserting user's properties?

pbrezina (changes requested):
Hi Alejandro, thank you for your hard work and for your patience with my review. You did an excellent job here. Please see few comments inline, but those are just very minor things.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,443 @@
...
+        :param objpath: The path to the object at destination.
+        :type objpath: str
+        """
+        return DBUSObject(self.host, dest=self._dest, objpath=objpath, bus=self._bus)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(*args):
+        pass
```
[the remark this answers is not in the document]
aplopez: I have no problem in removing it, but It was a requirement from you to make it work with the `with` statement.

Just let me know what you want me to do.

sssd_test_framework/utils/sbus.py
```diff
@@ -0,0 +1,443 @@
...
+        :param bus: The bus to use for the communications. Defaults to ``DBUSBus.SYSTEM``.
+            For other cases, a string can be provided with the explicit bus path.
+        :type bus: str | DBUSBus, optional
+        """
+        super().__init__(host)
+        self._dest = dest
+        self._bus = bus
+        self._objPaths: list[str] | None = None
+
+    def _instrospect(self, objPath: str) -> xml.Element:
```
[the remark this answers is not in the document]
aplopez: Fixed

sssd_test_framework/roles/client.py
pbrezina: > The constructor doesn't do anything complex, as you said, but I wanted to avoid connecting to the bus if infopipe will not be used (most of the non-dbus tests). What do you think?

But it currently doesn't connect in constructor, or does it? We call dbus-send only when we run introspection via `getObject`?

sssd_test_framework/utils/sbus.py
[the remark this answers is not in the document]
pbrezina: I anticipated that we will maintain connection in some way, then it would make sense. But since we rely on dbus-send and don't maintain any resources on our own, we don't need it.

docs/guides/testing-dbus.rst
aplopez: I used the `test_infopipe__GetUserProperties` test.

sssd_test_framework/utils/sbus.py
aplopez: Done

pbrezina (approved):

pbrezina:
Thank you, ack! Good job!

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,192 @@
+Testing D-Bus Services
+======================
+Infopipe
+--------
+Access to the Infopipe services is achieved through the
+:class:`sssd_test_framework.roles.client.Client` class. It provides the
+:meth:`sssd_test_framework.roles.client.Client.infopipe()` method that
+returns a pre-instantiated object of class :class:`sssd_test_framework.utils.sbus.DBUSDestination`, which gives you
```
ikerexxe, line 8: Line too long

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,192 @@
...
+* :meth:`sssd_test_framework.utils.dbus.types.DBUSType.parse()`: Parses the
+  string resulting from an execution of `dbus-send` and set the value to the
+  object.
+
+Basic types (integers, strings, booleans, etc.) are subclasses of the
+:class:`sssd_test_framework.utils.dbus.types.DBUSTypeBasic` abstract class.
+
+Container types -- that is, the subclasses of the abstract class
+:class:`sssd_test_framework.utils.dbus.types.DBUSTypeContainer` -- accept a
+parameter for their constructors, another :class:`sssd_test_framework.utils.dbus.types.DBUSType` object of the expected type.
```
ikerexxe, line 121: Line too long

ikerexxe (commented):
Just two minor comments in the documentation and we are done

docs/guides/testing-dbus.rst
aplopez: Fixed

docs/guides/testing-dbus.rst
aplopez: Fixed

ikerexxe (approved):
LGTM. Thanks for the hard work!

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,194 @@
...
+
+* :attr:`sssd_test_framework.utils.dbus.types.DBUSType.value`: A property to read
+  and set the (Python) value.
+* :meth:`sssd_test_framework.utils.dbus.types.DBUSType.mimic()`: a method to copy
+  itself without copying the value while maintaining the structure (subtypes).
+* :meth:`sssd_test_framework.utils.dbus.types.DBUSType.param()`: The string
+  representation of the value in a format suitable to be used as a parameter for
+  ``dbus-send``.
+* :meth:`sssd_test_framework.utils.dbus.types.DBUSType.parse()`: Parses the
+  string resulting from an execution of `dbus-send` and set the value to the
```
pbrezina, line 114: add double backtick around dbus-send

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,194 @@
...
+
+    s = "array [ uint32 1 uint32 2 uint32 3 ]"
+    a = DBUSTypeArray(DBUSTypeUInt32())
+    a.parse(DBUSResult(s))
+    print(a.value)
+    [1, 2, 3]
+
+In some cases it may not be possible to know in advance the type of the elements
+of a container type. In that case, no object is passed to the constructor. The
+type will be guessed while parsing the result from ```dbus-send``.
```
pbrezina, line 136: remov the third backtick

docs/guides/testing-dbus.rst
```diff
@@ -0,0 +1,194 @@
...
+.. note::
+    Although the `D-Bus specification`_ considers ``dict entry`` a separate type,
+    we didn't implement it as such because there is no use case for it outside
+    of an array, in which case the array becomes a dictionary.
+
+.. _D-Bus specification: https://dbus.freedesktop.org/doc/dbus-specification.html#type-system
+
+Not Implemented Types
+~~~~~~~~~~~~~~~~~~~~~
+Some other classes were not implemented because they are not accepted by `dbus-send`:
```
pbrezina, line 181: add double backtick around dbus-send

pbrezina (changes requested):

docs/guides/testing-dbus.rst
aplopez: What? This is one of the things I fixed yesterday...

Checking the history, something went wrong and removed that change. I don't understand what happened.

I'll fix it again...

docs/guides/testing-dbus.rst
aplopez: Done

docs/guides/testing-dbus.rst
aplopez: Done

pbrezina (approved):

ikerexxe (approved):

PR #28, opened 2023-08-04
Add a module for nsldc support.

aborah-sudo:
Create a module that will configure nslcd in client machine.

### DIFF
--- a/sssd_test_framework/utils/sssd.py
+++ b/sssd_test_framework/utils/sssd.py
@@ -4,12 +4,13 @@ from __future__ import annotations
 
 import configparser
 from io import StringIO
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from pytest_mh import MultihostHost, MultihostRole, MultihostUtility
 from pytest_mh.ssh import SSHLog, SSHProcess, SSHProcessResult
 
 from ..hosts.base import BaseDomainHost
+from ..misc import to_list
 
 if TYPE_CHECKING:
     from pytest_mh.utils.fs import LinuxFileSystem
@@ -782,3 +783,78 @@ class SSSDCommonConfiguration(object):
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def proxy(
+        self,
+        proxy: Literal["files", "ldap"] = "files",
+        provider: str | list[str] = "id",
+        proxy_pam_target: str | None = None,
+        proxy_pam_stack: str | None = None,
+        server_hostname: str | None = None,
+        domain: str | None = None,
+    ):
+        """
+        Configure files or ldap proxy domain.
+
+        :param proxy: ``ldap`` or ``files``, defaults to ``files``
+        :type proxy: Literal["files", "ldap"]
+        :param provider: SSSD providers (``id``, ``auth``, ``chpass``, ...), defaults to ``id``
+        :type provider: str | list[str]
+        :param proxy_pam_target: SSSD option proxy_pam_target, defaults to
+            ``None`` (= ``system-auth`` (files), ``sssdproxyldap`` (ldap))
+        :type proxy_pam_target: str | None
+        :param proxy_pam_stack: Custom PAM stack written to
+            /etc/pam.d/@proxy_pam_target, defaults to ``None`` (= ignored (files), pam_ldap.so (ldap))
+        :type proxy_pam_stack: str | None
+        :param server_hostname: LDAP server hostname for ldap proxy (ldap), ignored (files), defaults to ``None``
+        :type server_hostname: str | None
+        :param domain: Proxy domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        """
+        if domain is None:
+            domain = self.sssd.default_domain
+
+        if domain is None:
+            raise ValueError("No domain specified and default domain is not set!")
+
+        if domain is not None and self.sssd.default_domain is None:
+            self.sssd.default_domain = domain
+
+        match proxy:
+            case "ldap":
+                if proxy_pam_target is None:
+                    proxy_pam_target = "sssdproxyldap"
+
+                if proxy_pam_stack is None:
+                    proxy_pam_stack = """
+                        auth     required pam_ldap.so
+                        account  required pam_ldap.so
+                        password required pam_ldap.so
+                        session  required pam_ldap.so
+                    """
+
+                if server_hostname is None:
+                    raise ValueError("No server_hostname specified!")
+
+                self.sssd.fs.write(
+                    "/etc/nslcd.conf", f"uid nslcd\ngid ldap\nuri ldap://{server_hostname}\n", dedent=False
+                )
+                self.sssd.svc.restart("nslcd")
+            case "files":
+                if proxy_pam_target is None:
+                    proxy_pam_target = "system-auth"
+            case _:
+                raise ValueError(f"Unknown proxy type: {proxy}")
+
+        if proxy_pam_stack is not None:
+            self.sssd.fs.write(f"/etc/pam.d/{proxy_pam_target}", proxy_pam_stack)
+
+        options = {
+            "enabled": "true",
+            "proxy_lib_name": proxy,
+            "proxy_pam_target": proxy_pam_target,
+            **{f"{x}_provider": "proxy" for x in to_list(provider)},
+        }
+
+        self.sssd.dom(domain).clear()
+        self.sssd.dom(domain).update(options)


pbrezina:
Hi, how are we going to use nslcd for SSSD testing?

aborah-sudo:
> Hi, how are we going to use nslcd for SSSD testing?

This is the one and only setup that used for nslcd setup for current test framework: 

https://github.com/SSSD/sssd/blob/master/src/tests/multihost/alltests/conftest.py#L754

pbrezina:
I went over the use cases and it seems that we only use it to test the proxy provider.

For this, I think it would be better create a generic function that would configure the proxy provider to [SSSDCommonConfiguration](https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/utils/sssd.py#L704).

Something like:

```python
    def proxy(self, domain=None, proxy=$proxy, provider=$provider, role: MultihostRole = None):
        """
        ... $proxy can be ldap or files
        ... $provider can be one of sssd providers (id, auth, chpass, ...)
        """

       if domain is None:
           domain = self.default_domain

        match proxy:
           case "ldap":
               # configure and start nslcd, create pam stack for pam_ldap
               # uri is role.host.hostname
               # basedn is not needed, nslcd can figure it on its own
           case "files":
               # create pam stack for pam_unix
```

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,44 @@ def autofs(self) -> None:
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def proxy(self, host: MultihostHost, domain=None, proxy=None, provider=None, role: MultihostRole = None):
+        """
+        ... $proxy can be ldap or files
+        ... $provider can be one of sssd providers (id, auth, chpass, ...)
```
sidecontrol, line 790: Please extend the comments to the format of other functions. 

"""
description
:param name: $name
:type: $type

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,44 @@ def autofs(self) -> None:
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def proxy(self, host: MultihostHost, domain=None, proxy=None, provider=None, role: MultihostRole = None):
```
sidecontrol, line 787: Please provide type hints for the parameters, i.e.  domain  = str | None = None

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def proxy(self,
+              host: MultihostHost,
```
pbrezina, line 788: Not needed.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+              domain: str = None,
+              proxy: str = None,
```
pbrezina, line 790: we can use type `Literal["files", "ldap"]` since we expect specific value. None does not make sense since you don't default to anything so lets make it a required parameter.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def proxy(self,
+              host: MultihostHost,
+              domain: str = None,
```
pbrezina, line 789: `str | None`

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+              proxy: str = None,
+              provider: str = None,
```
pbrezina, line 791: let's make the type `str | list[str]` and default to "id". So it either set one provider or multiple providers.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+              role: MultihostRole = None):
+        """
+        :param host: Multihost host.
+        :type host: MultihostHost
```
pbrezina, line 795: host is not needed

sssd_test_framework/utils/sssd.py
pbrezina: Since this is not going to be used often, put it as last parameter.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        :param proxy: Ldap or files.
+        :type host: str
```
pbrezina, line 799: type proxy

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+              domain: str = None,
+              proxy: str = None,
+              provider: str = None,
+              role: MultihostRole = None):
+        """
+        :param host: Multihost host.
+        :type host: MultihostHost
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        :param proxy: Ldap or files.
```
pbrezina, line 798: You expect "ldap" not "Ldap". Use the following format to make it clear.
```
``ldap`` or ``files``
```

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+        :type host: str
+        :param provider: Sssd providers
```
pbrezina, line 800: uppercase SSSD

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+        :type provider: Sssd providers (id, auth, chpass, ...)
```
pbrezina, line 801: this should go to parameter description, here you want to specify the type

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+        :param role: MultihostRole
```
pbrezina, line 802: Multihost role, which host will be used as nslcd LDAP server.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+            case "ldap":
+                self.sssd.fs.write('/etc/pam.d/sssdproxyldap', '''
+                        auth    required pam_ldap.so
+                        account required pam_ldap.so
+                        password required pam_ldap.so
+                        session required pam_ldap.so
+                        ''')
+                self.sssd.fs.write('/etc/nslcd.conf', f'uid nslcd\ngid ldap\nuri '
+                                                 f'ldap://{role.host.hostname}\n', dedent=False)
+                self.sssd.config[f'domain/{domain}'] = {
```
pbrezina, line 818: `self.sssd.dom(domain).update(options)`

You need to update sssd configuration for both ldap and files so this can go after the match statement. Inside it, you just need to set the options.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+                    f'{provider}'+'_provider': 'proxy',
+                    'proxy_lib_name': proxy,
+                    'proxy_pam_target': 'sssdproxyldap'
+                }
+                host.host.ssh.run("systemctl restart nslcd")
```
pbrezina, line 823: This will yield error because host.host attribute does not exist. Please, actually test your code before submitting pull request.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+            case "files":
+                self.sssd.fs.write('/etc/pam.d/mylogin', '''
```
pbrezina, line 825: sssdproxyfiles so it is the same as for ldap case?

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,54 @@ def autofs(self) -> None:
...
+                        auth       required     pam_setcred.so
+                        session    required     pam_unix.so 
+                        password   required     pam_unix.so nullok obscure minlen=8
+                        ''')
+                self.sssd.config[f'domain/{domain}'] = {
+                    f'{provider}' + '_provider': 'proxy',
+                    'proxy_lib_name': proxy,
+                    'auth_provider': 'ldap',
+                    'proxy_pam_target': 'mylogin'
+                }
```
pbrezina, line 836: You need to add a fallback to the match statement and raise error if the caller did not set ldap nor files.

pbrezina (changes requested):
Hi Anuj, thank you for implementing the idea, but there are lots of issues in your code.
https://tests.sssd.io/en/latest/concepts.html#core-concepts-and-coding-style
Please, also fix CI issues. It would be good if you create a habit to automatically check your code before submitting it (testing it since current code will yield error about host.host attribute missing) and checking the coding style - you can just run `tox` command to do it for you. See https://tests.sssd.io/en/latest/concepts.html#core-concepts-and-coding-style on what tools you should run.

sssd_test_framework/utils/sssd.py
aborah-sudo: This is required to "systemclt restart nsldc"  Or how can we do that ? 

as self.host does not exists

sssd_test_framework/utils/sssd.py
aborah-sudo: I did :

(Pdb) host.host.ssh.run("systemctl restart nslcd")
<pytest_mh.ssh.SSHProcessResult object at 0x7f737b5a8850>
(Pdb)

aborah-sudo: If i remove host parameter . How i am suppose to restart nsdcl ?

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,50 @@ def autofs(self) -> None:
...
         self.sssd.enable_responder("autofs")
+
+    def proxy(self,
+              proxy: str,
+              provider: str | list[str] = 'id',
+              role: MultihostRole = None,
+              domain: str | None = None,):
+        """
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
```
pbrezina, line 794: Move to the end. Order of parameters here must follow the parameters of the function.

This docstring is used to render documenation.

sssd_test_framework/utils/sssd.py
pbrezina: This was not resolved.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,50 @@ def autofs(self) -> None:
...
+        :param proxy: ``ldap`` or ``files``
+        :type proxy: str
+        :param provider: SSSD providers(id, auth, chpass, ...)
+        :type provider: SSSD providers
+        :param role: MultihostRole(LDAP server)
+        :type role: MultihostRole
+        """
+        target = "sssdproxyldap"
```
pbrezina, line 802: move this to `case "ldap"`

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,50 @@ def autofs(self) -> None:
...
+              proxy: str,
+              provider: str | list[str] = 'id',
+              role: MultihostRole = None,
+              domain: str | None = None,):
+        """
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        :param proxy: ``ldap`` or ``files``
+        :type proxy: str
+        :param provider: SSSD providers(id, auth, chpass, ...)
```
pbrezina, line 797: Add space before (

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,50 @@ def autofs(self) -> None:
...
+        :type provider: SSSD providers
+        :param role: MultihostRole(LDAP server)
```
pbrezina, line 799: Add space before (

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,50 @@ def autofs(self) -> None:
...
+            case "files":
+                target = "sssdproxyfiles"
+                self.sssd.fs.write('/etc/pam.d/sssdproxyfiles', '''
+                        auth       required     pam_unix.so nullok_secure
+                        auth       required     pam_setcred.so
+                        session    required     pam_unix.so 
+                        password   required     pam_unix.so nullok obscure minlen=8
+                        ''')
+            case _:
+                raise Exception("Proxy can be ldap or files but nothing was provided .")
```
pbrezina, line 826: Remove space before dot.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,50 @@ def autofs(self) -> None:
...
+              provider: str | list[str] = 'id',
+              role: MultihostRole = None,
+              domain: str | None = None,):
+        """
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        :param proxy: ``ldap`` or ``files``
+        :type proxy: str
+        :param provider: SSSD providers(id, auth, chpass, ...)
+        :type provider: SSSD providers
```
pbrezina, line 798: type is `str | list[str]`

sssd_test_framework/utils/sssd.py
pbrezina: You didn't update the code, just moved it.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +783,50 @@ def autofs(self) -> None:
...
+                        auth       required     pam_unix.so nullok_secure
+                        auth       required     pam_setcred.so
+                        session    required     pam_unix.so 
+                        password   required     pam_unix.so nullok obscure minlen=8
+                        ''')
+            case _:
+                raise Exception("Proxy can be ldap or files but nothing was provided .")
+
+        self.sssd.config[f'domain/{domain}'] = {
+            f'{provider}' + '_provider': 'proxy',
```
pbrezina, line 829: You don't handle the case where provider is list.

```python
options = {
    "proxy_lib_name": proxy,
    "proxy_pam_target": target,
}

for x in to_list(provider):
    options[f"{x}_provider"] = "proxy"

self.sssd.dom(domain).update(options)
```

pbrezina (changes requested):
Please also:

* rename commit to "sssd: add common configuration for proxy domain"
* run black, isort, flake8 and mypy to automatically reformat and validate your code to pass the PR CI and follow the coding style

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,50 @@ def autofs(self) -> None:
...
+                        auth       required     pam_setcred.so
+                        session    required     pam_unix.so 
+                        password   required     pam_unix.so nullok obscure minlen=8
+                        ''')
+            case _:
+                raise Exception("Proxy can be ldap or files but nothing was provided.")
+
+        options = {"proxy_lib_name": proxy,
+                   "proxy_pam_target": target, }
+        for x in pd.Series(provider).tolist():
```
pbrezina, line 831: Please, use https://tests.sssd.io/en/latest/api/sssd_test_framework.misc.html#sssd_test_framework.misc.to_list

```
from ..misc import to_list
```

sssd_test_framework/utils/sssd.py
aborah-sudo: If we update :

[domain/test]
ldap_tls_reqcert = demand
ldap_tls_cacert = /data/certs/ca.crt
dns_discovery_domain = ldap.test
id_provider = proxy
ldap_uri = ldap://master.ldap.test
proxy_lib_name = ldap
proxy_pam_target = sssdproxyldap
proxy_fast_alias = true
debug_level = 0xfff0

We dont need all the parameters but only 

            "proxy_lib_name": proxy,
            "proxy_pam_target": target,
           [f"{x}_provider"] = "proxy"

aborah-sudo: Both Works

pbrezina: But there is also test where you have id_provider=ldap and chpass_provider=proxy. In that case you need to keep the existing stuff. So maybe add a switch to decide if it should override or update the configuration?

aborah-sudo: Update config works . So better to keep it that way . Its was just a info .

aborah-sudo:
> Please also:
> 
>     * rename commit to "sssd: add common configuration for proxy domain" and change the commit message since it is not really about nslcd
> 
>     * run black, isort, flake8 and mypy to automatically reformat and validate your code to pass the PR CI and follow the coding style

Its failing because nslcd  is  not installed in container

sidecontrol:
I want this to be reworked. 

SSSD proxy can use any PAM module, not just LDAP and Files. I'd like the proxy configuration to take arguments to accept any PAM module.  I suggest the setup changes to use authselect and a custom profile and the proxy setup with the added parameters to make the modules, well modular. 

I'd like to see the usage be something like this in the test code. 

sssd.config.proxy("pam_ldap.so", file="/etc/pam.d/sssd_proxy_auth", interface="auth", control="required")
utils.authselect(create custom profile, select profile) 

So the proxy function will take these parameters. 

module: str 
file: str | None = "default test file"
interface: str | None = "auth" 
control: str | None = "required"
args: str 

From this, we can construct the PAM file and with authselect, we can enable it.

aborah-sudo:
> I want this to be reworked.
> 
> SSSD proxy can use any PAM module, not just LDAP and Files. I'd like the proxy configuration to take arguments to accept any PAM module. I suggest the setup changes to use authselect and a custom profile and the proxy setup with the added parameters to make the proxy setup configurable.
> 
> I'd like to see the usage be something like this in the test code.
> 
> sssd.config.proxy("pam_ldap.so", file="/etc/pam.d/sssd_proxy_auth", interface="auth", control="required")
> 

> The proxy function should have these parameters, for example.
> 
> module: str file: str | None = "default test file" interface: str | None = "auth" control: str | None = "required" args: str | None = None


"sssd_proxy_auth" file need to created within the test ?
What will be the task of "control" here ? I mean what it will do ?
What will be the args ? How it will be used ? Will it be a list also ?

> 
> Then we can construct the pam file instead of hard coding it, and use the authselect to enable it to be quickly and safely undone.

sidecontrol:
"sssd_proxy_auth" file need to created within the test ? 
Exactly! We don't want to have to edit the framework code if we want to change the name of the file in the test. 

What will be the task of "control" here ?

control is the second column of the pam configuration. 

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system-level_authentication_guide/pam_configuration_files

I mean what it will do ? 

Specify whether the module is 
- required
- requisite
- sufficient
- optional 

What will be the args ?

That depends on which pam module is being used.  Specifically to the pam_ldap.so that you are using, the following can be args:
- config=<path>
- use_first_pass 
- try_first_pass
- ignore_unknown_user
- ignore_authinfo_unavail
- no_warn
- use_authtok
- debug

How it will be used?

The pam configuration you are using  is /etc/pam.d/sssdproxyfiles , you are adding the following in the content. We need to add things to the framework that can be used for other tests as well. So I propose that we create a class to configure this file, so it can all be done in the test code and will not require changes to the framework. 

```
# interface       #control      #module                  #args
auth                   required      pam_unix.so          nullok_secure
auth                   required      pam_setcred.so
session              required      pam_unix.so
password          required      pam_unix.so          nullok obscure minlen=8
```

https://linux.die.net/man/5/pam_ldap

This should really be handled in the pam class that I'm writing, hrmm.... 

Will it be a list also ?

Not sure I understand. 
 



Oh, Actually we already have  SSSDCommonConfiguration.local, your code to add a "files" proxy, so you don't have to write that at all! 

Let's do, SSSDCommonConfiguration.nslcd , take a look at the API, https://tests.sssd.io/en/latest/api/sssd_test_framework.utils.sssd.html , and I hope you understand what I'm getting at.

aborah-sudo:
@sidecontrol @pbrezina I have made things more easy . Now we can have pam stack written within the test (Because we dont know what test need what pam stack) and then the name of the pam stack should be provided to the module . That way things will more dynamic

sidecontrol (changes requested):

sidecontrol:
Please make sure static code checks are working before the next review. 

CI / Tox is 3.10 
CI / Tox is 3.x 
docs/readthedocs.org:sssd-tests 

are failing.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,42 @@ def autofs(self) -> None:
...
+            """
+                auth    required pam_ldap.so
+                account required pam_ldap.so
+                password required pam_ldap.so
+                session required pam_ldap.so
+                """,
+        )
+
+        self.sssd.fs.write(
+            "/etc/nslcd.conf", f"uid nslcd\ngid ldap\nuri " f"ldap://{role.host.hostname}\n", dedent=False
```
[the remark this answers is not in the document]
aborah-sudo: its black formatted

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,42 @@ def autofs(self) -> None:
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def nslcd(
+        self,
+        provider: str | list[str] = "id",
```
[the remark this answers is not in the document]
aborah-sudo: As  per this doc: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/domain_configuration_options-configuring_a_proxy_domain 

Yes . 

[domain/PROXY_PROXY]
auth_provider = proxy
id_provider = proxy
proxy_lib_name = ldap
proxy_pam_target = sssdproxyldap
cache_credentials = true

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,42 @@ def autofs(self) -> None:
...
+        role: Optional[MultihostRole] = None,
```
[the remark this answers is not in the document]
aborah-sudo: According to this doc :  https://wiki.samba.org/index.php/Nslcd

Yes . 

The nslcd service enables you to configure your local system to load users and groups from an LDAP directory, such as Active Directory (AD). To enable the nslcd service to load user and group information, you have to set the Unix attributes for users and groups in AD.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,42 @@ def autofs(self) -> None:
...
+        :param role: MultihostRole (LDAP server)
+        :type role: MultihostRole
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        """
+        if domain is None:
+            domain = self.sssd.default_domain
+
+        self.sssd.fs.write(
+            "/etc/pam.d/sssdproxyldap",
```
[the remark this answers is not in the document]
aborah-sudo: This function name will be nslcd_ldap . So this can be default config

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
aborah-sudo: For nsdcl ldap coffig it will work .

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
sidecontrol: Providing a document and the configuration of nslcd doesn't mean we need it. Looking at all of our old setup [code](https://github.com/SSSD/sssd/blob/master/src/tests/multihost/alltests/conftest.py#L754), there is no _auth_provider = proxy_  used, so do we need it?

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,42 @@ def autofs(self) -> None:
...
+
+    def proxy(
+        self,
+        pam_stack: str,
+        proxy: str,
+        provider: str | list[str] = "id",
+        role: Optional[MultihostRole] = None,
+        domain: str | None = None,
+    ):
+        """
```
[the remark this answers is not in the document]
sidecontrol: Please add a period at the end.

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
sidecontrol: It's easy to add another parameter, so it can be changed if needed, right?

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
sidecontrol: The other way can also be black formatted as well. Let's please, adhere to how this file is written.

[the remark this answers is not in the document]
sidecontrol: Can you please test it or is the rest of the configuration specified somewhere? I don't understand how it will get the userPassword from the directory with no encryption and privileged access to read the attribute to be able to log in.

sidecontrol (changes requested):

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
aborah-sudo: [root@client sssd]# ssh user-1@localhost
The authenticity of host 'localhost (<no hostip for proxy command>)' can't be established.
ECDSA key fingerprint is SHA256:QC8XilGyktQmXT84um5KuQjwvYjMAMRWpxqwbhpNNHM.
This key is not known by any other names
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
(user-1@localhost) Password: 
Last login: Fri Aug 11 10:30:28 2023 from ::1
-sh-5.2$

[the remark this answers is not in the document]
aborah-sudo: [domain/test]
proxy_lib_name = ldap
debug_level = 0xFFF0
proxy_pam_target = sssdproxyldap
id_provider = proxy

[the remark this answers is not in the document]
aborah-sudo: -sh-5.2$ cat /etc/nslcd.conf
uid nslcd
gid ldap
uri ldap://master.ldap.test
-sh-5.2$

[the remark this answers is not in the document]
aborah-sudo: [root@client sssd]# cat sssd.conf 
[sssd]
config_file_version = 2
services = nss, pam
domains = test
debug_level = 0xfff0

[domain/test]
proxy_lib_name = ldap
debug_level = 0xFFF0
proxy_pam_target = sssdproxyldap
id_provider = proxy
auth_provider = proxy
proxy_fast_alias = true

[autofs]
debug_level = 0xfff0

[ifp]
debug_level = 0xfff0

[kcm]
debug_level = 0xfff0

[nss]
debug_level = 0xfff0

[pac]
debug_level = 0xfff0

[pam]
debug_level = 0xfff0

[ssh]
debug_level = 0xfff0

[sudo]
debug_level = 0xfff0[root@client sssd]# ssh user-1@localhost
(user-1@localhost) Password: 
Last login: Sat Aug 12 07:06:04 2023 from ::1
-sh-5.2$ cat /etc/nslcd.conf
uid nslcd
gid ldap
uri ldap://master.ldap.test
-sh-5.2$

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
sidecontrol: Please either change the code or present a better argument of why this should be hardcoded?

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
sidecontrol: It does not present a good argument for why this needs to be parameterized, I think we can remove this and remove the for loop entire reducing this from 4 lines of code to 1. 

Please answer the questions or update the code accordingly before resolving the conversation. I don't see this as resolved yet.

sidecontrol:
> Please also:
> 
> * rename commit to "sssd: add common configuration for proxy domain" and change the commit message since it is not really about nslcd

I think we need to make it about nslcd, because it's under CommonConfigurations. Seeing that we already have CommonConfigurations.local that also uses the proxy provider, it makes sense to me that this should configure sssd to proxy to pam_ldap/nslcd to keep it simple.

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
sidecontrol: Can you tell me why it works without a bind user/password and/or ssl/tls please? I don't know, I want to understand why and how nslcd works works without them.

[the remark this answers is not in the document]
aborah-sudo: @pbrezina can answer this he asked me to remove the base_dn also from config . According to him nslcd is able to find those stuff on its own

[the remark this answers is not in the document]
sidecontrol: Please provide a better reason, then it's just "it's black formatted"

`f"uid nslcd\ngid ldap\nuri\n ldap://{role.host.hostname}\n"`

I'm sure this will pass the checks as well. It's cleaner, it matches the rest of the file. Why do we even need 

The f if now variable is not specified? It could also be "uid nslcd\nbid ldap\nuri " f"ldap://{role.host.hostname}\n"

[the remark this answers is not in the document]
aborah-sudo: I already changed this to :  f"uid nslcd\ngid ldap\nuri ldap://{role.host.hostname}\n"

pbrezina:
So, you went from overcomplicated idea back to my original proposal, just less generic. Keep in mind, that you are not testing nslcd. You are testing the proxy domain. It would be more helpful to have a way to create more types of proxy domains using a @pytest.mark.parametrize then to provide multiple tests suites one for files, one for ldap, ...

The fact that you have to use nss-pam-ldapd should be irrelevant to the test.

aborah-sudo:
> 

It should be fine now .

aborah-sudo:
> So, you went from overcomplicated idea back to my original proposal, just less generic. Keep in mind, that you are not testing nslcd. You are testing the proxy domain. It would be more helpful to have a way to create more types of proxy domains using a @pytest.mark.parametrize then to provide multiple tests suites one for files, one for ldap, ...
> 
> The fact that you have to use nss-pam-ldapd should be irrelevant to the test.

It should be fine now .

sidecontrol:
@pbrezina  I suggested this and I understand what you are saying. If this is going to go into common configurations, it should just configure it for that configuration. Seeing that we already have a sssd.common.local which uses proxy provider already.  We have the two main ones covered.   I didn't make sense to me that we didn't have a common.<id provider> , but everything in common is specific configuration types.  Enable autofs, sudo, kcm, krb5_auth, so I think just doing nslcd makes more sense.  

If we want to make it all things proxy, I think we need to refactor the code to use and remove common.local to use this function instead so some and files be the default behavior. I propose the following,

1. remove common.local and refactor test code
2. sssd.proxy() = configures proxy for files , default behavior
3. sssd.proxy("ldap") configures nslcd

pbrezina:
We can also add alias proxy_files and proxy_ldap to make it even easier to use.

sidecontrol:
@aborah-sudo Please resolve the CI/Tox errors.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,50 @@ def autofs(self) -> None:
...
+
+        match proxy:
+            case "ldap":
+                target = "sssdproxyldap"
+                self.sssd.fs.write(f"/etc/pam.d/{target}", proxy_file)
+                self.sssd.fs.write(
+                    "/etc/nslcd.conf", f"uid nslcd\ngid ldap\nuri " f"ldap://{sever_hostname}\n", dedent=False
+                )
+                self.sssd.svc.restart("nslcd")
+            case "files":
```
sidecontrol, line 819: I don't think this is correct and it's confusing by the naming. Proxy ldap forwards to an PAM stack while files doesn't, but it's called files. 

So in case "files": 
[domain/files]
id_provider = files

is all that is required. 

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/system-level_authentication_guide/configuring_domains#configuring-the-files-provider-for-sssd

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,50 @@ def autofs(self) -> None:
...
+                target = "sssdproxyfiles"
+                self.sssd.fs.write(f"/etc/pam.d/{target}", proxy_file)
+            case _:
+                raise Exception("Proxy can be ldap or files but nothing was provided.")
```
sidecontrol, line 823: Can we make files the default behavior? We have a lot more tests that will be using just files instead of nslcd?

sidecontrol:
I'm confused. We've gone to just nslcd again, but named proxy_ldap? To proxy and files, now back to nslcd just named something else? 

I made a mistake in my previous comment, the instructions I listed are just for files, without proxy, The correct configuration is 

```
id_provider="proxy",
proxy_lib_name="files",
proxy_pam_target="system-auth",
```

sidecontrol:
I haven't tested it, but I think this makes the most sense. 

```
    def proxy(
        self,
        proxy: str | None = "files",
        provider: str | list[str] = "id",
        proxy_file: str | None = None,
        server: str | None = None,
        domain: str | None = None,
    ):
        """
        :param proxy: Proxy provider, files or ldap
        :type proxy: str, optional
        :param provider: SSSD providers (proxy, id, auth, chpass, ...)
        :type provider: str | list[str]
        :param proxy_file: Config for proxy
        :type proxy_file: str | None, optional
        :param server: LDAP (LDAP server hostname)
        :type server: str | None, Optional
        :param domain: domain name, defaults to None (= default domain)
        :type domain: str | None, optional
        """
        options = {}
        if proxy == "files":
            options = {
                    "proxy_lib_name": "files",
                    "proxy_pam_target": "system-auth",
            }

        if proxy == "ldap":
            if domain is None:
                domain = self.sssd.default_domain

            if proxy_file is None:
                proxy_file = "sssdproxyldap"

            self.sssd.fs.write(f"/etc/pam.d/{proxy_file}", proxy_file)
            self.sssd.fs.write("/etc/nslcd.conf", f"uid nslcd\ngid ldap\nuri " f"ldap://{server}\n", dedent=False)
            self.sssd.svc.restart("nslcd")

            options = {
                "proxy_lib_name": "ldap",
                "proxy_pam_target": proxy_file,
            }

        for x in to_list(provider):
            options[f"{x}_provider"] = "proxy"

        self.sssd.dom(domain).update(options)
```

aborah-sudo:
> I'm confused. We've gone to just nslcd again, but named proxy_ldap? To proxy and files, now back to nslcd just named something else?
> 
> I made a mistake in my previous comment, the instructions I listed are just for files, without proxy, The correct configuration is
> 
> ```
> id_provider="proxy",
> proxy_lib_name="files",
> proxy_pam_target="system-auth",
> ```
> 
> Which uses the default pam file system-auth instead of sssdproxyfiles

Ok we make files the default behavior now

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,49 @@ def autofs(self) -> None:
...
+        :param proxy_file: Config for proxy
+        :type proxy_file: str
+        :param sever_hostname: LDAP (LDAP server hostname)
+        :type sever_hostname: str
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        """
+        if domain is None:
+            domain = self.sssd.default_domain
+        assert domain is not None
```
sidecontrol, line 810: Like this rest of the code in this file, if there isn't a domain, let's raise an exception. 

        if domain is None:
            raise ValueError("No domain specified!")

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,49 @@ def autofs(self) -> None:
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def proxy(
+        self,
+        proxy: Literal["files", "ldap"],
+        provider: str | list[str] = "id",
+        proxy_file: str = """""",
```
sidecontrol, line 792: Do you need to declare an empty variable? Why not use None instead?

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,55 @@ def autofs(self) -> None:
...
+        :param proxy_file: Config for proxy
+        :type proxy_file: str
+        :param sever_hostname: LDAP (LDAP server hostname)
+        :type sever_hostname: str
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        """
+        if domain is None:
+            domain = self.sssd.default_domain
+        assert domain is not None, "No domain specified!"
```
sidecontrol, line 810: We shouldn't be using assertions. Please raise an exception.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,55 @@ def autofs(self) -> None:
...
+
+        if proxy_file is None:
```
sidecontrol, line 812: This needs to have a default value for files, the proxy file for files is 'system-auth' so the code will stop executing here.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,55 @@ def autofs(self) -> None:
...
+            raise ValueError("No proxy_file specified!")
+
+        if sever_hostname is None:
+            raise ValueError("No sever_hostname specified!")
```
sidecontrol, line 816: If it's files, it doesn't need the server, so this will always raise an exception if files is used.

sidecontrol (changes requested):
It's looking better, just a few small things that need to be addressed.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,53 @@ def autofs(self) -> None:
...
+            domain = self.sssd.default_domain
+        if domain is None:
+            raise ValueError("No domain specified!")
+
+        proxy_pam_target = proxy_file
+        match proxy:
+            case "ldap":
+                if sever_hostname is None:
+                    raise ValueError("No sever_hostname specified!")
+                proxy = "ldap"
```
ikerexxe, line 818: You are already checking in a condition that proxy is `ldap`, so I don't think this is needed

aborah-sudo: changed

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,52 @@ def autofs(self) -> None:
...
+        :param proxy_file: Config for proxy
+        :type proxy_file: str
+        :param sever_hostname: LDAP (LDAP server hostname)
+        :type sever_hostname: str
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        """
+        if domain is None:
+            domain = self.sssd.default_domain
+        if domain is None:
```
pbrezina, line 810: Separate with new line please.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,52 @@ def autofs(self) -> None:
         """
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
+
+    def proxy(
+        self,
+        proxy: Literal["files", "ldap"] = "files",
+        provider: str | list[str] = "id",
+        proxy_file: str = "system-auth",
+        sever_hostname: str | None = None,
```
pbrezina, line 793: Typo sever -> "server_hostname"

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,52 @@ def autofs(self) -> None:
...
+        :type sever_hostname: str
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        """
+        if domain is None:
+            domain = self.sssd.default_domain
+        if domain is None:
+            raise ValueError("No domain specified!")
+
+        proxy_pam_target = proxy_file
```
pbrezina, line 813: You use proxy_file as filename here, but as file contents later. What is the intention?

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,52 @@ def autofs(self) -> None:
...
+        match proxy:
+            case "ldap":
+                if sever_hostname is None:
+                    raise ValueError("No sever_hostname specified!")
+                proxy_pam_target = "sssdproxyldap"
```
pbrezina, line 818: New line before please.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,52 @@ def autofs(self) -> None:
...
+            "proxy_lib_name": proxy,
+            "proxy_pam_target": proxy_pam_target,
+        }
+
+        for x in to_list(provider):
+            options[f"{x}_provider"] = "proxy"
+
+        self.sssd.dom(domain).clear()
+        self.sssd.dom(domain).update(options)
+        self.sssd.start()
```
pbrezina, line 835: Don't start SSSD here. That should be left on the caller.

pbrezina (changes requested):
Hi, please see few comments inside.

sssd_test_framework/utils/sssd.py
aborah-sudo: It will be the proxy configuration for ex:

       """
            auth    required pam_ldap.so
            account required pam_ldap.so
            password required pam_ldap.so
            session required pam_ldap.so
            """,
If could be different in different test

pbrezina: Ok, then I suggest to add two parameters:

* proxy_pam_target: str | None = None - if None, then select `system-auth` for `files` and `sssdproxyldap` for `ldap`
* proxy_pam_stack: str | None = None - if None, choose different default for `files` and `ldap`.

pbrezina: It depends what you want. PAM authentication can be handled for files proxy by pam_sss if you require fully qualified user names. In that case, customized PAM config can be used - but of course, default should be system-auth.

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,62 @@ def autofs(self) -> None:
...
+            if proxy_pam_target is None:
+                raise ValueError("No proxy_pam_target specified!")
+
+            if proxy_pam_stack is None:
+                raise ValueError("No proxy_pam_stack specified!")
+
+            if server_hostname is None:
+                raise ValueError("No server_hostname specified!")
+
+            self.sssd.fs.write(f"/etc/pam.d/{proxy_pam_target}", proxy_pam_stack)
```
pbrezina, line 830: You want to write this for files as well `if proxy_pam_stack is not None`.

pbrezina (changes requested):
Hi, thank you Anuj. From my perspective, the logic is there, we just need to move few lines after the if statement and fix types.

sssd_test_framework/utils/sssd.py
aborah-sudo: For files its not need, it will be :  "system-auth"

aborah-sudo: id_provider="proxy",
proxy_lib_name="files",
proxy_pam_target="system-auth"

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,66 @@ def autofs(self) -> None:
...
+        :param domain: Existing domain name, defaults to None (= default domain)
+        :type domain: str | None, optional
+        """
+        if domain is None:
+            domain = self.sssd.default_domain
+
+        if domain is None:
+            raise ValueError("No domain specified!")
+
+        proxy_pam_target = "system-auth"
```
pbrezina, line 817: You're completely ignoring value passed as parameter.

sssd_test_framework/utils/sssd.py
pbrezina: Yes, but what if you want to provide custom pam target and custom pam stack?

pbrezina (changes requested):

pbrezina:
Hi, we were going in a in circle - one thing is fixed, another issue created. So I took the liberty to fix documentation and parameter handling to move on with this pull request.

Diff: https://github.com/SSSD/sssd-test-framework/compare/fa556449d721aa363c181800e3e51f2f42532738..3240e9b61f7c0c0154a6f98cc3adc54934eea8e0

Sample tests:
```python
@pytest.mark.topology(KnownTopology.Client)
def test_proxy__files(client: Client):
    client.local.user("tuser").add()
    client.sssd.common.proxy()
    client.sssd.start()

    result = client.tools.id(f"tuser@{client.sssd.default_domain}")
    assert result is not None
    assert result.user.name == "tuser"


def test_proxy__ldap(client: Client, ldap: LDAP):
    ldap.user("tuser").add()
    client.sssd.common.proxy("ldap", server_hostname=ldap.host.hostname)
    client.sssd.start()

    result = client.tools.id(f"tuser@{client.sssd.default_domain}")
    assert result is not None
    assert result.user.name == "tuser"
```

Now we need to add nss-pam-ldapd to containers before pushing this. @sidecontrol can you take care of that?

sidecontrol:
@pbrezina , I don't know how to do it yet, but yes, I will figure it out and take care of it.

sidecontrol:
https://github.com/SSSD/sssd-ci-containers/pull/58

sssd_test_framework/utils/sssd.py
```diff
@@ -783,3 +784,74 @@ def autofs(self) -> None:
...
+
+        :param proxy: ``ldap`` or ``files``, defaults to ``files``
+        :type proxy: Literal["files", "ldap"]
+        :param provider: SSSD providers (``id``, ``auth``, ``chpass``, ...), defaults to ``id``
+        :type provider: str | list[str]
+        :param proxy_pam_target: SSSD option proxy_pam_target, defaults to
+            ``None`` (= ``system-auth`` (files), ``sssdproxyldap`` (ldap))
+        :type proxy_pam_target: str | None
+        :param proxy_pam_stack: Custom PAM stack written to
+            /etc/pamd./@proxy_pam_target, defaults to ``None`` (= ignored (files), pam_ldap.so (ldap))
```
ikerexxe, line 808: s/pamd./pam.d

ikerexxe:
Please review the unresolved conversations, if they are already fixed then resolve them. Apart from that rebase on top of master.

pbrezina:
I built nslcd in copr for downstream rhel9 testing and pushed Dan's container patch again. Once it is verified that we didn't break downstream CI again, we can push this.

pbrezina (approved):

ikerexxe (approved):

pbrezina:
@sidecontrol do you agree with latest version?

sidecontrol:
@pbrezina Yes. 

@aborah-sudo Thank you.

sidecontrol (approved):

PR #29, opened 2023-08-07
 tools: adding dnf support

sidecontrol:


aplopez (dismissed):
LGTM. Thanks.

pbrezina:
Hi, why do you need dnf? All packages should be installed by sssd-ci-containers project. In an unlikely case that you need dnf for something very specific, you must make sure to revert changes done by dnf during teardown.

ikerexxe:
> For the authselect tests, there is a test to ensure that the minimal profile doesn't require sssd-ad, sssd-ldap and other packages. How do you propose that we write that test? It is more of a functional system/build/maintainer test.

What's exactly the package that you need to install to test it?

pbrezina:
Me and Dan talked, one use case (removing the sssd packages) can be implemented in other way. Second use case is about removing authselect and verifying that authselect configuration was converted to non-authselect configuration which really requires the authselect package to be removed.

For this Dan needs dnf support, but the changes must be automatically reverted. Using dnf history undo will help here. Perhaps we can get current history ID when we first use dnf and then revert to this id on teardown.

sidecontrol:
> > For the authselect tests, there is a test to ensure that the minimal profile doesn't require sssd-ad, sssd-ldap and other packages. How do you propose that we write that test? It is more of a functional system/build/maintainer test.
> 
> What's exactly the package that you need to install to test it?

The code for this test case is tough to read. We will certainly improve it when porting the test suite over. As far as I can tell, sssd-client, sssd-common, sssd-nfs-idmap

`
        all_delete1 = [i for i in all_rpms
                       if (i.startswith('sssd-client') == False
                           and i.startswith('sssd-common') == False
                           and i.startswith('sssd-nfs-idmap') == False)]
        all_delete2 = [i for i in all_rpms if i.startswith('sssd-common-pac')]
        all_final = all_delete1 + all_delete2
        all_keep1 = [i for i in all_rpms if
                     (i.startswith('sssd-client') == True
                      or i.startswith('sssd-common') == True
                      or i.startswith('sssd-nfs-idmap') == True)]
`

sidecontrol:
@pbrezina @aplopez 

Getting the latest transaction ID isn't sufficient. In the event that a package is already installed, the command exits without errors and the teardown proceeds to undo the wrong ID. So I added a statement to compare the IDs before adding the command to the teardown list.

ikerexxe (dismissed):
LGTM. Thanks for the patch!

sidecontrol:
Added "-y" to the module so we don't have to pass it through each time.

ikerexxe (approved):

PR #34, opened 2023-08-15
requirements: syncing up pytest_mh with sssd.git

sidecontrol:


sidecontrol:
```
Successfully installed pytest-mh-1.0.4

cat .venv/lib/python3.11/site-packages/pytest_mh/utils/fs.py | grep def\ append

# Nothing found
```

```
 pip list | grep pytest-mh
pytest-mh               1.0.5.dev10+g27e27e8

cat .venv/lib/python3.11/site-packages/pytest_mh/utils/fs.py | grep def\ append
    def append(
```

This is failing tox, so I'm not sure what we want to do.

PR #30, opened 2023-08-08
markers: add isbuiltwith marker

pbrezina:
This allows to skip a test if particular SSSD functionality is missing.

PR #35, opened 2023-08-16
Add back 'markers: add builtwith marker' with fix

justin-stephenson:
https://github.com/SSSD/sssd-test-framework/pull/30 was merged and reverted, due to breaking all system tests (`/usr/lib64/sssd/libsss_files.so` did not exist, see https://github.com/SSSD/sssd/actions/runs/5833344764/job/15821158997).

I am re-opening this reverted PR with the fix  `markers: avoid non-zero exit code with unavailable feature `

~~~
            [ -f "/usr/lib64/sssd/libsss_files.so" ] && echo "files-provider"
~~~

Changed to

~~~
            [ -f "/usr/lib64/sssd/libsss_files.so" ] && echo "files-provider" || exit 0
~~~

This PR was tested and CI is green with https://github.com/SSSD/sssd/pull/6884


requirements.txt
```diff
@@ -1,4 +1,4 @@
 jc
 pytest
 python-ldap
-pytest-mh >= 1.0.3
+pytest-mh
```
ikerexxe, line 4: Shouldn't this be `pytest-mh >= 1.0.4`?

sssd_test_framework/hosts/client.py
```diff
@@ -41,7 +41,7 @@ def features(self) -> dict[str, bool]:
             """
             set -ex
 
-            [ -f "/usr/lib64/sssd/libsss_files.so" ] && echo "files-provider"
+            [ -f "/usr/lib64/sssd/libsss_files.so" ] && echo "files-provider" || exit 0
```
ikerexxe, line 44: This will exit the existing session without breaking the tests. But will it detect when the feature is missing?

justin-stephenson: Yes, the following lines of code create a features dict, enabling the feature based on the string echo'd back from this line. If no string is echo'd back then the feature is disabled.

requirements.txt
ikerexxe: Or maybe we need to release 1.0.5 once https://github.com/next-actions/pytest-mh/pull/11 is merged. I mean, there is kind of a tight relationship between this PR and that one.

justin-stephenson: Okay we can wait for Pavel, but I don't think the `requirements.txt` commit needs to part of this PR so I removed it.

pbrezina: Strictly speeking, we need to release pytest-mh 1.0.5 and bump the version here, please use `requirement: bump pytest-mh minimum version` as commit subject.

But our CI should always pull newest pytest-mh from its master branch, so no harm should be done in upstream CI. Downstream CI might get affected though.

justin-stephenson: Okay, I added the commit to bump pytest-mh to >= 1.0.5.

justin-stephenson:
hi @pbrezina  - anything here still pending ?

pbrezina:
I release pytest-mh-1.0.5. There is still unresolved comment in this PR: https://github.com/SSSD/sssd-test-framework/pull/35#discussion_r1305523976

sssd_test_framework/hosts/client.py
justin-stephenson: Updated.

PR #36, opened 2023-08-17
authselect: adding authselect current, enable-feature and disable-feature methods

sidecontrol:


sidecontrol:
This was branched off of https://github.com/SSSD/sssd-test-framework/pull/29 , and needs to be rebased after this is merged.

justin-stephenson:
Hi,

Do you know how the teardown will work with the various authselect backups which are being taken here? 


`authselect current`, `authselect enable-feature` and `authselect disable-feature` are all taking backups (though I think authselect enable and disable feature backups are not being restored as they are not included in the `teardown` method).

For example what if in a test you call, in this order.
~~~
authselect enable-feature featureA
authselect disable-feature featureB
authselect select sssd
~~~
in this case restoring the backup from the `select sssd` command does not sufficiently roll back to the original authselect state.

justin-stephenson:
Woops disregard the "Closed" PR, I click the wrong button :)

sidecontrol:
Thinking about it, I don't think we need the backup on these at all. These methods will be used after the first backup is taken, then when the teardown occurs it will be reverted back to the initial state, we don't need to roll back each and every change. .

sssd_test_framework/utils/authselect.py
```diff
@@ -61,3 +61,29 @@ def select(self, profile: str, features: list[str] = []) -> None:
...
+        :rtype: str
+        """
+        result = self.host.ssh.exec(["authselect", "current"]).stdout
+
+        return result
+
+    def disable_feature(self, features: list[str]) -> None:
+        """
+        Disable Authselect feature.
+        :param features: Authselect features to enable
```
ikerexxe, line 78: s/enable/disable

sidecontrol:
@ikerexxe It happens in select() , which should happen before enable-feature or disable-feature is used.

ikerexxe:
> @ikerexxe It happens in select() , which should happen before enable-feature or disable-feature is used.

If that's a requirement, then I think you should add it as a comment

ikerexxe:
I merged https://github.com/SSSD/sssd-test-framework/pull/29, so you should rebase on master. Moreover, there are two additional commits not related to this PR. Please remove them so that we can move forward.

> @ikerexxe I added the teardown to enable and disable, so it won't matter in what order the command is executed. All methods will create the backup only when the backup doesn't exist.

I think you forgot to add it because I don't see it.

sidecontrol:
Okay, rebased, only relevant code is in this PR

ikerexxe:
> @ikerexxe Try taking a look at the file without the comments. I was seeing some weirdness looking at Anuj's PR not seeing his changes.
> 
> Each of the functions has the following.
> 
> ```
>         if self.__backup is None:
>             self.__backup = "multihost.backup"
>             backup = [f"--backup={self.__backup}"]
>         self.host.ssh.exec(["authselect", "disable-feature", *features, *backup])
> ```

Now it's there. Very strange... LGTM!

ikerexxe (approved):

justin-stephenson (approved):
Ack, Thank you.

PR #39, opened 2023-09-01
Add Keycloak role and topology

spoore1:
Adding role and topology for Keycloak to support new tests that require Keycloak IdP.

### DIFF
--- a/docs/guides/using-roles.rst
+++ b/docs/guides/using-roles.rst
@@ -35,6 +35,7 @@ There are multiple roles available.
 * ``ipa`` -- IPA server
 * ``ldap`` -- 389ds server
 * ``samba`` -- Samba Domain Controller
+* ``keycloak`` -- Keycloak server
 * ``client`` -- SSSD client
 
 Each role is accessible through pytest fixture.
@@ -43,8 +44,8 @@ Using provider roles
 ********************
 
 Provider roles, that is those that represents identity management service (ad,
-samba, ipa, ldap), provide interface to manipulate the service. For example
-managing users and groups. These roles implements a generic interface
+samba, ipa, ldap, keycloak), provide interface to manipulate the service. For
+example managing users and groups. These roles implements a generic interface
 :class:`~sssd_test_framework.roles.generic.GenericProvider` and further extends
 this interface with service specifics.
 :class:`~sssd_test_framework.roles.generic.GenericProvider` can be used when
@@ -118,7 +119,8 @@ different containers). Group management works in the same way but
     :class:`~sssd_test_framework.roles.ad.AD`,
     :class:`~sssd_test_framework.roles.ipa.IPA`,
     :class:`~sssd_test_framework.roles.ldap.LDAP`,
-    :class:`~sssd_test_framework.roles.samba.Samba`
+    :class:`~sssd_test_framework.roles.samba.Samba`,
+    :class:`~sssd_test_framework.roles.keycloak.Keycloak`
 
 Using the client role
 *********************
--- a/sssd_test_framework/hosts/keycloak.py
+++ b/sssd_test_framework/hosts/keycloak.py
@@ -1,0 +1,111 @@
+"""Keycloak multihost host."""
+
+from __future__ import annotations
+
+import time
+
+from pytest_mh.ssh import SSHProcessError
+
+from .base import BaseDomainHost
+
+__all__ = [
+    "KeycloakHost",
+]
+
+
+class KeycloakHost(BaseDomainHost):
+    """
+    Keycloak host object.
+
+    Provides features specific for Keycloak server.
+
+    This class adds ``config.adminpw`` multihost configuration option to set
+    password of the Keycloak admin user so we run kcadm.sh commands to set
+    options in Keycloak.
+
+    .. code-block:: yaml
+        :caption: Example multihost configuration
+        :emphasize-lines: 4
+
+        - hostname: master.keycloak.test
+          role: keycloak
+          config:
+            adminpw: Secret123
+
+    .. note::
+
+        Full backup and restore is supported. However, the operation relies on
+        ``kc.sh export`` and ``kc.sh import`` commands which can take several
+        seconds to finish.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.adminpw = self.config.get("adminpw", "Secret123")
+
+        # Backup of original data
+        self.__backup: str | None = None
+
+    def kclogin(self) -> None:
+        """
+        Obtain ``admin`` user credentials for Keycloak.
+        """
+        for x in range(0, 5):
+            try:
+                str_error = None
+                self.ssh.exec(
+                    [
+                        "/opt/keycloak/bin/kcadm.sh",
+                        "config",
+                        "credentials",
+                        "--server",
+                        f"https://{self.hostname}:8443/auth/",
+                        "--realm",
+                        "master",
+                        "--user",
+                        "admin",
+                        "--password",
+                        self.adminpw,
+                    ]
+                )
+            except SSHProcessError as err:
+                str_error = str(err)
+
+            if str_error:
+                time.sleep(30)
+            else:
+                break
+
+    def backup(self) -> None:
+        """
+        Backup all Keycloak server data.
+
+        This is done by calling ``kc.sh export`` on the server
+        and can take several seconds to finish.
+        """
+        if self.__backup is not None:
+            return
+
+        cmd = self.ssh.run(
+            "set -e; systemctl stop keycloak;"
+            "/opt/keycloak/bin/kc.sh export --dir /tmp/kcbackup"
+            "> /tmp/kcbackup.log;"
+            "systemctl start keycloak;"
+            "ls -1 /tmp/kcbackup| tail -n 1"
+        )
+        self.__backup = cmd.stdout.strip()
+
+    def restore(self) -> None:
+        """
+        Restore all Keycloak server data to its original state.
+
+        This is done by calling ``kc.sh import`` on the server
+        and can take several seconds to finish.
+        """
+        if self.__backup is None:
+            return
+
+        self.ssh.run(
+            "systemctl stop keycloak; /opt/keycloak/bin/kc.sh import --dir /tmp/kcbackup; systemctl start keycloak"
+        )
--- a/sssd_test_framework/roles/ad.py
+++ b/sssd_test_framework/roles/ad.py
@@ -623,6 +623,9 @@ class ADUser(ADObject):
             "OtherAttributes": (self.cli.option.PLAIN, self._attrs_to_hash(unix_attrs)),
             "Enabled": (self.cli.option.PLAIN, "$True"),
             "Path": (self.cli.option.VALUE, self.path),
+            "EmailAddress": (self.cli.option.PLAIN, f'{self.name}@{self.host.client["ad_domain"]}'),
+            "GivenName": (self.cli.option.PLAIN, "dummyfirstname"),
+            "Surname": (self.cli.option.PLAIN, "dummylastname"),
         }
 
         self._add(attrs)
--- a/sssd_test_framework/roles/keycloak.py
+++ b/sssd_test_framework/roles/keycloak.py
@@ -1,0 +1,263 @@
+from __future__ import annotations
+
+import json
+import shlex
+
+from pytest_mh.ssh import SSHProcessResult
+
+from ..hosts.keycloak import KeycloakHost
+from .base import BaseLinuxRole, BaseObject
+
+__all__ = [
+    "Keycloak",
+    "KeycloakUser",
+    "KeycloakGroup",
+]
+
+
+class Keycloak(BaseLinuxRole[KeycloakHost]):
+    """
+    Keycloak service management.
+    """
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+    def setup(self) -> None:
+        """
+        Make sure we have admin's TGT so we can run ipa cli commands.
+        """
+        super().setup()
+        self.host.kclogin()
+
+    def kcadm(self, command: str) -> SSHProcessResult:
+        """
+        Run kcadm command on the Keycloak server.
+
+        This is the CLI to interface with the Keycloak API to perform functions
+        like adding users, deleting groups, etc.
+
+        :param command: kcadm command
+        :type command: str
+        """
+        kcadm = "/opt/keycloak/bin/kcadm.sh"
+        command_split = shlex.split(command)
+        result = self.host.ssh.exec([kcadm] + command_split)
+        return result
+
+    def user(self, name: str) -> KeycloakUser:
+        """
+        Get user object.
+
+        :param name: User name.
+        :type name: str
+        :return: New user object.
+        :rtype: KeycloakUser
+        """
+        return KeycloakUser(self, name)
+
+    def group(self, name: str) -> KeycloakGroup:
+        """
+        Get group object.
+
+        :param name: Group name.
+        :type name: str
+        :return: New group object.
+        :rtype: KeycloakGroup
+        """
+        return KeycloakGroup(self, name)
+
+
+class KeycloakObject(BaseObject[KeycloakHost, Keycloak]):
+    """
+    Keycloak Object.
+    """
+
+    def __init__(self, role: Keycloak, name: str) -> None:
+        """
+        :param role: Keycloak role object.
+        :type role: Keycloak
+        :param name: User name.
+        :type name: str
+        """
+        super().__init__(role)
+        self.name = name
+        self.id: str = ""
+
+
+class KeycloakUser(KeycloakObject):
+    """
+    Keycloak user management.
+    """
+
+    def __init__(self, role: Keycloak, name: str) -> None:
+        """
+        :param role: Keycloak role object.
+        :type role: Keycloak
+        :param name: User name.
+        :type name: str
+        """
+        super().__init__(role, name)
+
+    def add(
+        self,
+        *,
+        password: str | None = "Secret123",
+    ) -> KeycloakUser:
+        """
+        Create new Keycloak user.
+
+        Parameters that are not set are ignored.
+
+        :param password: Password, defaults to 'Secret123'
+        :type password: str | None, optional
+        :return: Self.
+        :rtype: KeycloakUser
+        """
+        create_user = (
+            "create users -r master " f"-s username={self.name} " f"-s email={self.name}@ipa.test " "-s enabled=true"
+        )
+        result = self.role.kcadm(create_user)
+
+        self.id = result.stderr.split()[-1].strip("'")
+
+        set_password = "set-password -r master " f"--username {self.name} " f"--new-password {password}"
+        self.role.kcadm(set_password)
+
+        return self
+
+    def delete(self) -> None:
+        """
+        Delete Keycloak user.
+        """
+        del_user = f"delete users/{self.id}"
+        self.role.kcadm(del_user)
+
+    def get(self) -> dict[str, list[str]]:
+        """
+        Get Keycloak user details.
+
+        :return: Dict of user info
+        :rtype: Dict
+        """
+        get_user = f"get users -q username={self.name}"
+        result = self.role.kcadm(get_user)
+
+        out: dict[str, list[str]] = {}
+        if not result.stdout or result.stdout == "[ ]":
+            return out
+
+        json1 = json.loads(result.stdout)[0]
+        for key in json1.keys():
+            out.setdefault(key, [])
+            out[key].append(json1[key])
+
+        return out
+
+    def get_groups(self) -> list[dict[str, list[str]]]:
+        """
+        Get Keycloak groups that user is a member of.
+
+        :return: Dict of user info
+        :rtype: Dict
+        """
+        get_groups = f"get users/{self.id}/groups"
+        result = self.role.kcadm(get_groups)
+
+        outlist: list[dict[str, list[str]]] = []
+        if not result.stdout:
+            return outlist
+
+        json1 = json.loads(result.stdout)
+        for item in json1:
+            out: dict[str, list[str]] = {}
+            for key in item.keys():
+                out.setdefault(key, [])
+                out[key].append(item[key])
+            outlist.append(out)
+
+        return outlist
+
+
+class KeycloakGroup(KeycloakObject):
+    """
+    Keycloak group management.
+    """
+
+    def __init__(self, role: Keycloak, name: str) -> None:
+        """
+        :param role: Keycloak role object.
+        :type role: Keycloak
+        :param name: Group name.
+        :type name: str
+        """
+        super().__init__(role, name)
+
+    def add(self) -> KeycloakGroup:
+        """
+        Create new Keycloak group.
+
+        Parameters that are not set are ignored.
+
+        :return: Self.
+        :rtype: KeycloakGroup
+        """
+        create_group = f"create groups -r master -s name={self.name}"
+        result = self.role.kcadm(create_group)
+
+        self.id = result.stderr.split()[-1].strip("'")
+        return self
+
+    def add_member(self, member: KeycloakUser | KeycloakGroup) -> KeycloakGroup:
+        """
+        Add group member.
+
+        :param member: User or group to add as a member.
+        :type member: KeycloakUser | KeycloakGroup
+        :return: Self.
+        :rtype: KeycloakGroup
+        """
+        return self.add_members([member])
+
+    def add_members(self, members: list[KeycloakUser | KeycloakGroup]) -> KeycloakGroup:
+        """
+        Add multiple group members.
+
+        :param member: List of users or groups to add as members.
+        :type member: list[KeycloakUser | KeycloakGroup]
+        :return: Self.
+        :rtype: KeycloakGroup
+        """
+        for item in self.__get_member_args(members):
+            update_group = f"update {item}/groups/{self.id} -r master"
+            self.role.kcadm(update_group)
+        return self
+
+    def remove_member(self, member: KeycloakUser | KeycloakGroup) -> KeycloakGroup:
+        """
+        Remove group member.
+
+        :param member: User or group to remove from the group.
+        :type member: KeycloakUser | KeycloakGroup
+        :return: Self.
+        :rtype: KeycloakGroup
+        """
+        return self.remove_members([member])
+
+    def remove_members(self, members: list[KeycloakUser | KeycloakGroup]) -> KeycloakGroup:
+        """
+        Remove multiple group members.
+
+        :param member: List of users or groups to remove from the group.
+        :type member: list[KeycloakUser | KeycloakGroup]
+        :return: Self.
+        :rtype: KeycloakGroup
+        """
+        remove_member = f"remove-member {self.__get_member_args(members)}"
+        self.role.kcadm(remove_member)
+        return self
+
+    def __get_member_args(self, members: list[KeycloakUser | KeycloakGroup]) -> list[str]:
+        users = [x for item in members if isinstance(item, KeycloakUser) for x in (f"users/{item.id}",)]
+        groups = [x for item in members if isinstance(item, KeycloakGroup) for x in (f"groups/{item.id}",)]
+        return [*users, *groups]
--- a/sssd_test_framework/config.py
+++ b/sssd_test_framework/config.py
@@ -43,6 +43,7 @@ class SSSDMultihostDomain(MultihostDomain[SSSDMultihostConfig]):
         from .hosts.client import ClientHost
         from .hosts.ipa import IPAHost
         from .hosts.kdc import KDCHost
+        from .hosts.keycloak import KeycloakHost
         from .hosts.ldap import LDAPHost
         from .hosts.nfs import NFSHost
         from .hosts.samba import SambaHost
@@ -55,6 +56,7 @@ class SSSDMultihostDomain(MultihostDomain[SSSDMultihostConfig]):
             "samba": SambaHost,
             "nfs": NFSHost,
             "kdc": KDCHost,
+            "keycloak": KeycloakHost,
         }
 
     @property
@@ -68,6 +70,7 @@ class SSSDMultihostDomain(MultihostDomain[SSSDMultihostConfig]):
         from .roles.client import Client
         from .roles.ipa import IPA
         from .roles.kdc import KDC
+        from .roles.keycloak import Keycloak
         from .roles.ldap import LDAP
         from .roles.nfs import NFS
         from .roles.samba import Samba
@@ -80,4 +83,5 @@ class SSSDMultihostDomain(MultihostDomain[SSSDMultihostConfig]):
             "samba": Samba,
             "nfs": NFS,
             "kdc": KDC,
+            "keycloak": Keycloak,
         }


sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,299 @@
...
+        dom_name: str,
+        dom_url: str,
+        client_id: str | None = "admin",
+        client_secret: str | None = "Secret123",
+        id_provider: str | None = "ipa",
+        ldap_tls_cacert: str | None = None,
+    ) -> KeycloakDomain:
+        """
+        Create new Keycloak integration domain.
+
```
sidecontrol, line 265: Please add docstrings for each parameter.

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,299 @@
...
+                        f"-s config.cacert=[\"{ldap_tls_cacert}\"] " \
+                        f"-s config.addintgdomain=[\"True\"] " \
+                        f"-s config.delintgdomain=[\"True\"]"
+
+        result = self.role.kcadm(create_domain)
+
+        return result.stdout
+
+    def ipa_fix(self, client: Client) -> None:
+        user_attrs = "mail:mail, sn:sn, givenname:givenname"
```
sidecontrol, line 295: Doc string, I don't know why this function is necessary, can you please add it to the description why this is needed?

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,299 @@
...
+                       f"--username {self.name} " \
+                       f"--new-password {password}"
+        self.role.kcadm(set_password)
+
+        return self
+
+    def delete(self) -> None:
+        """
+        Delete existing Keycloak user.
+        :return: Nothing
```
sidecontrol, line 128: You can take this out, if it returns nothing we don't need it to say that it returns nothing.

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,62 @@
...
+    def kclogin(self) -> None:
+        """
+        Obtain ``admin`` user credentials for Keycloak
+        """
+        #self.ssh.exec(['kinit', 'admin'], input=self.adminpw)
+        self.ssh.exec(['/opt/keycloak/bin/kcadm.sh', 'config', 'credentials',
+                       '--server', f'https://{self.hostname}:8443/auth/',
+                       '--realm', 'master', '--user', 'admin',
+                       '--password', self.adminpw])
+        return
```
sidecontrol, line 36: I think you can remove this, not returning anything.

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,62 @@
...
+    "KeycloakHost",
+]
+
+
+class KeycloakHost(BaseDomainHost):
+    """
+    Keycloak host object.
+
+    Provides features specific for Keycloak server.
+    """
```
sidecontrol, line 17: I know the docs hasn't been updated, but please put your sample code, and usage here.

sidecontrol (changes requested):

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,62 @@
...
+        Backup all IPA server data.
+
+        This is done by calling ``ipa-backup --data --online`` on the server
+        and can take several seconds to finish.
+        """
+        if self.__backup is not None:
+            return
+
+        self.ssh.run('/opt/keycloak/bin/kc.sh export --dir /tmp/kcbackup')
+        cmd = self.ssh.run('ls -1 /tmp/kcbackup | tail -n 1')
```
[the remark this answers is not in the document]
spoore1: I just followed what the IPAHost did here:  https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/hosts/ipa.py#L68

sssd_test_framework/hosts/keycloak.py
spoore1: removing

sssd_test_framework/hosts/keycloak.py
spoore1: updating.

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        Restore all IPA server data to its original state.
+
+        This is done by calling ``ipa-restore --data --online`` on the server
+        and can take several seconds to finish.
+        """
+        if self.__backup is None:
+            return
+
+        self.ssh.run('/opt/keycloak/bin/kc.sh import --dir /tmp/kcbackup')
```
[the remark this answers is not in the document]
spoore1: similar to above, I followed what IPAHost was doing.   So if a backup isn't run or self.__backup is not set, skip running the restore/import.

sssd_test_framework/roles/keycloak.py
spoore1: removing.

sssd_test_framework/roles/keycloak.py
spoore1: adding

sssd_test_framework/hosts/keycloak.py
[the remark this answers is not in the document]
sidecontrol: I was tired when I reviewed this, I it read wrong, LG2M now.

sidecontrol:
I think everything looks good, but docs/using-roles.rst needs to be updated.

sidecontrol (approved):
LGTM

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
+Testing Keycloak integration
+############################
+
+Keycloak integration can be tested using the related methods provided in
+:class:`sssd_test_framework.roles.keycloak.Keycloak`.
+
+Multhost test configuration requirements
```
pbrezina, line 7: Multihost

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
...
+          username: root
+          password: Secret123
+        config:
+          keycloak:
+            keycloak_server: keycloak.ipa.test
+            keycloak_admin_user: admin
+            keycloak_admin_password: Secret123
+
+It should be noted that at this time, the ipa-tuura bridge service does not
+yet have its own role.  Instead it is defined as a client for remote access if
```
pbrezina, line 48: Double space before Instead

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
...
+          host: HOSTNAME_KEYCLOAK
+          username: root
+          password: Secret123
+        config:
+          keycloak:
+            keycloak_server: keycloak.ipa.test
+            keycloak_admin_user: admin
+            keycloak_admin_password: Secret123
+
+It should be noted that at this time, the ipa-tuura bridge service does not
```
pbrezina, line 47: How about create ipa-tuura bridge role and just inherit the class from Client?

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
...
+yet have its own role.  Instead it is defined as a client for remote access if
+necessary and is managed via the Keycloak SCIM storage plugin.
+
+Adding a Keycloak Integration Domain
+------------------------------------
+
+Keycloak integration requires the Keycloak server to add an Integration Domain
+that will point to the identity provider being tested (IPA, AD, LDAP, etc).
+This is accomplished using the method
+:meth:`sssd_test_framework.roles.keycloak.KeycloakDomain.add`.   After Adding
```
pbrezina, line 57: Double space before After
After Adding -> After adding

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
...
+the Integration Domain, you can add users to keycloak and they should be synced
+to the identity provider.  At that point, you should be able to use them like
```
pbrezina, line 59: Double space before At.

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
...
+
+    from pytest_mh import Topology, TopologyDomain
+
+    from sssd_test_framework.roles.client import Client
+    from sssd_test_framework.roles.keycloak import Keycloak
+    from sssd_test_framework.roles.ipa import IPA
+    from sssd_test_framework.roles.generic import GenericProvider
+
+
+    @pytest.mark.topology(
```
pbrezina, line 79: Add Keacloak topology to KnownTopologies?

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
...
+        "ipa",
+        Topology(TopologyDomain("sssd", client=1, keycloak=1, ipa=1)),
+        keycloak="sssd.keycloak[0]", client="sssd.client[0]", ipa="sssd.ipa[0]", provider="sssd.ipa[0]"
```
pbrezina, line 82: provider can be omitted, I think, since this is tightly bound to IPA, therefore you can not use topology parametrization. Or can you?

docs/guides/testing-keycloak.rst
```diff
@@ -0,0 +1,111 @@
...
+            scim_host="bridge.ipa.test",
+            scim_admin="scim",
+            scim_password="Secret123",
+            dom_name="ipa.test",
+            dom_url="http://ipa.test",
+            client_id="admin",
+            client_secret="Secret123",
+            id_provider="ipa",
+            ldap_tls_cacert="/etc/ipaca.crt",
+        )
```
pbrezina, line 104: This would be good to add to the setup of the keycloak role.

Or maybe even better, it should be made a prerequisite for the host. Similar what we have on client - client must be enrolled into domains, so we don't need to reenroll it every time.

What do you think?

pbrezina (changes requested):
Hi, it looks good in general, see some comments inside. Additionally, please check type hints in docstrings since you sometime use for example IPAGroup instead of KeycloakGroup etc.

You also need to modify https://tests.sssd.io/en/latest/config.html (probably lot's of stuff from testing-keycloak should go there)

docs/guides/testing-keycloak.rst
spoore1: fixing

docs/guides/testing-keycloak.rst
spoore1: My original thought was that an ipa-tuura bridge role wouldn't be needed since Keycloak handles most of the interaction with the service through the new storage plugin.   I had added this to use ssh to implement a workaround in a test case but, that has been fixed directly in ipa-tuura now so it's no longer needed.

I'll look into this more today though.

docs/guides/testing-keycloak.rst
spoore1: Sorry, old habit of double spacing after period.   Changing to single.

docs/guides/testing-keycloak.rst
spoore1: fixing.

docs/guides/testing-keycloak.rst
spoore1: fixing.

docs/guides/testing-keycloak.rst
pbrezina: So it may not be needed from API point of view, but you need the host available and the hosts must meet some conditions (SCIMv2 server installed, etc). These conditions are described by the role. So even if you don't access this host, you want to add a specific role to it so you can say that this test requires "keycloak, ipa and ipa-tuura bridge". Now you don't require ipa-tuura bridge and if I pass client container it will still run the test.

If you just put client there, it makes it something different. The requirements are formally described here: https://tests.sssd.io/en/latest/config.html

docs/guides/testing-keycloak.rst
pbrezina: Why would you do that intentionally? :-)

docs/guides/testing-keycloak.rst
spoore1: That's something I wanted to ask about.    A very basic keycloak topology could be created with Keycloak as a provider.   This may be useful later but, I'm not sure yet.  Currently, with ipa-tuura testing, we need something like:

`keycloak, bridge(client), provider(IPA|LDAP|AD)`

docs/guides/testing-keycloak.rst
spoore1: Yes, it could be but, I have some questions/issues with using existing ones for the tests.   

I didn't think I could use the existing topologies because the tests don't need an NFS instance.  Could I just define the nfs server in the mhc.yaml as the client or ipa server and ignore it in the tests?

And if I used existing KnownTopologies, how would I define the keycloak role fixture to be used with the test?

```
@pytest.mark.topology(KnownTopology.AD)
@pytest.mark.topology(KnownTopology.IPA)
def test_example(keycloak: Keycloak, client: Client, provider: GenericProvider):
```

Or should I create new topologies for each like:
```
Keycloak_IPA = (keycloak, client, ipa, provider)
Keycloak_AD = (keycloak, client, ad, provider)
Keycloak_LDAP = (keycloak, client, ldap, provider)
```

docs/guides/testing-keycloak.rst
spoore1: Another thing I was going to ask about.

Keycloak.domain.add() should (almost) always connect with the info from provider (IPA|AD|LDAP).    How could I accomplish this?  In the definition of the KnownTopology somehow?   

Or is it easier to just define that info as new config attributes in the mhc.yaml?   That might also eliminate the need for a bridge role in the mhc.yaml (until we actually need one (if ever)).

docs/guides/testing-keycloak.rst
spoore1: ok, I'll add an ipa-tuura bridge role.

docs/guides/testing-keycloak.rst
pbrezina: So the bridge is running SCIMv2+SSSD, with SSSD connected to the backends, right? The second example is what you want. But the client should be ipa_tuura_bridge or scimclient or something more specific and Keycloak should probably also switch to IPA_TUURA, however IPA_TUURA_AD looks weird and IPA is not involved at all here, right?

Other way that would be maybe better is to define these topologies in ipa-tuura tests only (i.e. within your test code) and then you can use just IPA, AD, Samba, LDAP.  Because you have sligthly different needs then what SSSD has. Then you can freely use client.

spoore1: Yes, SSSD connects to the backends in the ipa-tuura setup.   Also, correct that there is no IPA involved in the ipa-tuura setup with AD.

I see the logic in something like IPA_TUURA_{PROVIDER} name.  Especially considering that ipa-tuura is what is actually connected to the backend provider. (via SSSD).  Part of the reason I was thinking Keycloak_{PROVIDER} was that the goal of using ipa-tuura is to connect Keycloak to the backend providers.   So I was thinking End to End Keycloak to AD, Keycloak to IPA, Keycloak to LDAP.

I could also just define those with the ipa-tuura tests but, how should I document this?  Or should I create a generic, simple Keycloak toplogy that just containers Keycloak as the provider (also will be useful later I think)?

docs/guides/testing-keycloak.rst
spoore1: I realized that I can't define the info needed in the config under bridge or keycloak because it has to come from the parametrized provider.   Is there an easier way to do that in the Keycloak.setup to run a self.domain.add() than to use something like:

```
for p in ['ipa', 'ad', 'ldap']:
    if self.host.domain.hosts_by_role(p):
        break
if not p:
    raiseSomeError

provider = host.domain.hosts_by_role(p)[0]

self.domain.add(
...
client_secret=provider.adminpw
...)
```

Is there a shorter/easier way to find the provider currently in use?

docs/guides/testing-keycloak.rst
pbrezina: So the idea is not to have everything under sssd-test-framework, that's why it is all separated into pytest-mh and sssd-test-framework. pytest-mh can be used by any projects, sssd-test-framework is for sssd project but most idm projects can benefit from it. But extensions specific to particular project should go to the project, not here.

The way I see it, Keycloak role can go here, because SSSD can benefit it when we will include tests for external IdPs. But these topologies containing the bridge are specific to ipa-tuura and should go to this project - you can add it either to the same repository as tests or in another repository and you can use it together with sssd-test-framework.

Therefore you will be able to have custom roles and topologies that makes sense for ipa-tuura but not for sssd and you can provide shorter and meaningful names. Documentation should go to ipa-tuura project.

It's up to you, I won't deny the contribution of course. If you choose to place it all here, IPA_TUURA_(PROVIDER) is in my opinion the way to go. But yes, the idea is to share as much as possible, but sssd-test-framework is not supposed to be the ultimate do-it-all repository, because you will reach these kind of decisions.

docs/guides/testing-keycloak.rst
pbrezina: It depends on what you want to. Is the bridge connected to all backends or always just to one? You can put it to mhc.yaml but you may need to have more configurations.

But if you look what we do with "client", we add sssd.conf options from the provider host's mhc config. This allows us to configure SSSD based on the currently selected topology.

We extend the pytest-mh topology with custom stuff that allows us to do that, see:
* https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/topology.py#L17
* https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/config.py#L17

I would say don't try to montage it all to SSSD project requirements. ipa-tuura requires slightly different things, so maybe you need to make your extensions to allow to test it neatly.

It may be good to have a call and see what needs to be done.

docs/guides/testing-keycloak.rst
spoore1: Ok, I'll put the ipa-tuura specific topologies with those tests and create a generic keycloak one that we can use later for external IdP.

docs/guides/testing-keycloak.rst
spoore1: FYI, I moved the topology definitions for the ipa-tuura stuff to the ipa-tuura test code I'm working on separately.   I do think we might need a separate Keycloak topology in the future but, I need to think about what else is needed there.  Keycloak and client only I think.   Not 100% sure yet.

docs/guides/testing-keycloak.rst
spoore1: I moved the ipa-tuura specific topologies as mentioned above.

pbrezina (changes requested):
Hi, I still see that you are making changes, so I am not sure how  much this is ready to review.

I think the documentation is outdated since it still mentions ipa-tuura a lot but this seems to be gone from the code.

sssd_test_framework/topology.py
```diff
@@ -207,6 +207,16 @@ def test_ldap(client: Client, ldap: LDAP):
     .. topology-mark:: KnownTopology.IPATrustSamba
     """
 
+    Keycloak = SSSDTopologyMark(
+        name="keycloak",
+        topology=Topology(TopologyDomain("sssd", ipa=1, keycloak=1)),
+        domains=dict(test="sssd.keycloak[0]"),
+        fixtures=dict(ipa="sssd.ipa[0]", keycloak="sssd.keycloak[0]", provider="sssd.keycloak[0]"),
+    )
```
pbrezina, line 215: I think that the provider should remain ipa and client should be part of this topology. This way we can use it to test IdP. Ideally, this pull request should allow us to test IdP (sssd -> ipa -> keycloak) but it may be difficult to implement authentication since it requires interaction with web browser. 

If you're not going to do it, then I think it is enough to just provide the keycloak host and role se we can built up on it eventually. In this case, you don't need to provide documentation and topology since we can not use it to test anything yet.

pbrezina:
FYI I closed the old comments since they probably do not apply anymore.

sssd_test_framework/roles/ad.py
```diff
@@ -623,6 +623,9 @@ def add(
             "OtherAttributes": (self.cli.option.PLAIN, self._attrs_to_hash(unix_attrs)),
             "Enabled": (self.cli.option.PLAIN, "$True"),
             "Path": (self.cli.option.VALUE, self.path),
+            "EmailAddress": (self.cli.option.PLAIN, f'{self.name}@{self.host.client["ad_domain"]}'),
+            "GivenName": (self.cli.option.PLAIN, "dummyfirstname"),
+            "Surname": (self.cli.option.PLAIN, "dummylastname"),
```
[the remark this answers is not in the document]
spoore1: Yes, I can move that into a separate commit.

sssd_test_framework/topology.py
spoore1: Yes, I think I'll drop the topology for now.   It can be added later when adding the authentication part for the tests you describe.   I believe that should be coming early this year some time.

docs/guides/testing-keycloak.rst
spoore1: This was fixed but then dropped to move to the ipa-tuura test code.

docs/guides/testing-keycloak.rst
spoore1: This was also fixed then moved elsewhere.

docs/guides/testing-keycloak.rst
spoore1: This was also fixed then moved.

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,108 @@
...
+        This is done by calling ``kc.sh export`` on the server
+        and can take several seconds to finish.
+        """
+        if self.__backup is not None:
+            return
+
+        self.ssh.run("systemctl stop keycloak")
+        self.ssh.run("/opt/keycloak/bin/kc.sh export --dir /tmp/kcbackup")
+        self.ssh.run("systemctl start keycloak")
+        cmd = self.ssh.run("ls -1 /tmp/kcbackup | tail -n 1")
```
pbrezina, line 93: It might be better to put it into single ssh call to safe resources.

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,108 @@
...
+
+        This is done by calling ``kc.sh import`` on the server
+        and can take several seconds to finish.
+        """
+        if self.__backup is None:
+            return
+
+        self.ssh.run("systemctl stop keycloak")
+        self.ssh.run("/opt/keycloak/bin/kc.sh import --dir /tmp/kcbackup")
+        self.ssh.run("systemctl start keycloak")
```
pbrezina, line 108: Same here.

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        Delete existing Keycloak user.
+        """
+        del_user = f"delete users/{self.id}"
+        self.role.kcadm(del_user)
+
+    def get(self) -> dict[str, list[str]]:
+        """
+        Get Keycloak user details and return them.
+        :return: Dict of user info
```
pbrezina, line 139: Add blank line.

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,265 @@
...
+        json1 = json.loads(result.stdout)[0]
+        for key in json1.keys():
+            out.setdefault(key, [])
+            out[key].append(json1[key])
+
+        return out
+
+    def get_groups(self) -> list[dict[str, list[str]]]:
+        """
+        Get Keycloak groups that user is a member of
```
pbrezina, line 158: Add blank line.

pbrezina (changes requested):
Hi Scott,
I would remove to documentation that you had written so far - since at this point, it is not usable for SSSD and therefore I found the documentation vague and misleading. So I'd remove that for now and write it when we actually implement relevant testing framework for IdP. Then we could provide better description and examples.

See minor comments inside, otherwise LGTM. I don't plan to test it, I trust that the code works for use.

sssd_test_framework/hosts/keycloak.py
spoore1: Updating but, should I leave the `ls`  separate to avoid potential issues with output from a previous command affecting the `ls` used to determine _backup path?

sssd_test_framework/hosts/keycloak.py
spoore1: updating

sssd_test_framework/roles/keycloak.py
spoore1: fixing

sssd_test_framework/roles/keycloak.py
spoore1: fixing

sssd_test_framework/hosts/keycloak.py
pbrezina: Is there any output on successful run? If yes, then sure. Otherwise `set -e` should suffice. But I'll leave it up to you.

spoore1: There is output on successful run from the export command but I think I resolved that with a redirect (which may be helpful debugging later too).

docs/guides/testing-keycloak.rst
spoore1: removed the doc so resolving this one.

sssd_test_framework/topology.py
spoore1: dropped the topology for now so resolving this.

spoore1:
> Hi Scott, I would remove to documentation that you had written so far - since at this point, it is not usable for SSSD and therefore I found the documentation vague and misleading. So I'd remove that for now and write it when we actually implement relevant testing framework for IdP. Then we could provide better description and examples.
> 
> See minor comments inside, otherwise LGTM. I don't plan to test it, I trust that the code works for use.

@pbrezina, I removed the doc and cleaned up the hosts run blocks for backup/restore.   Can you take another look at the updates?

Thanks

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        super().__init__(*args, **kwargs)
+
+    def setup(self) -> None:
+        """
+        Make sure we have admin's TGT so we can run ipa cli commands.
+        """
+        super().setup()
+        self.host.kclogin()
+
+    def kcadm(self, command: str) -> SSHProcessResult:
```
danlavu, line 33: 1. Can we use svc / systemctl to manage this?
2. To conform to other classes in the framework, rename this to start? 
3. Do we need stop/restart ?

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        Create new Keycloak user.
+
+        Parameters that are not set are ignored.
+
+        :param password: Password, defaults to 'Secret123'
+        :type password: str | None, optional
+        :return: Self.
+        :rtype: KeycloakUser
+        """
+        # run command to add user
```
danlavu, line 114: Comment is redundant.

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        create_user = (
+            "create users -r master " f"-s username={self.name} " f"-s email={self.name}@ipa.test " "-s enabled=true"
+        )
+        result = self.role.kcadm(create_user)
+
+        # set Keycloak user ID with results of command
```
danlavu, line 120: Redundant, it's readable.

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        self.id = result.stderr.split()[-1].strip("'")
+
+        # run command to set user password
```
danlavu, line 123: Again!

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        set_password = "set-password -r master " f"--username {self.name} " f"--new-password {password}"
+        self.role.kcadm(set_password)
+
+        return self
+
+    def delete(self) -> None:
+        """
+        Delete existing Keycloak user.
```
danlavu, line 131: Delete keycloak user.

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        """
+        del_user = f"delete users/{self.id}"
+        self.role.kcadm(del_user)
+
+    def get(self) -> dict[str, list[str]]:
+        """
+        Get Keycloak user details and return them.
```
danlavu, line 138: Get Keycloak user information/details 

"and return them." isn't necessary

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        json1 = json.loads(result.stdout)[0]
+        for key in json1.keys():
+            out.setdefault(key, [])
+            out[key].append(json1[key])
+
+        return out
+
+    def get_groups(self) -> list[dict[str, list[str]]]:
+        """
+        Get Keycloak groups that user is a member of
```
danlavu, line 159: .

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        """
+        Create new Keycloak group.
+
+        Parameters that are not set are ignored.
+
+        :return: Self.
+        :rtype: KeycloakGroup
+        """
+
+        # run command to add group
```
danlavu, line 206: Redundant.

sssd_test_framework/utils/__init__.py
```diff
@@ -1,2 +1,3 @@
 """SSSD multihost utils used by roles."""
+
```
danlavu, line 2: Looks like your IDE is updating this file?

sssd_test_framework/roles/__init__.py
```diff
@@ -1,2 +1,3 @@
 """SSSD multihost roles."""
+
```
danlavu, line 2: Looks like your IDE is updating this file?

sssd_test_framework/misc/__init__.py
```diff
@@ -1,4 +1,5 @@
 """Miscellaneous functions."""
+
```
danlavu, line 2: Looks like your IDE is updating this file?

sssd_test_framework/hosts/__init__.py
```diff
@@ -1,2 +1,3 @@
 """SSSD multihost hosts."""
+
```
danlavu, line 2: Looks like your IDE is updating this file?

tests/test_misc.py
```diff
@@ -1,4 +1,5 @@
 """Miscellaneous functions."""
+
```
danlavu, line 2: Looks like your IDE is updating this file?

danlavu (changes requested):

docs/guides/using-roles.rst
```diff
@@ -43,8 +44,8 @@ Using provider roles
 ********************
 
 Provider roles, that is those that represents identity management service (ad,
-samba, ipa, ldap), provide interface to manipulate the service. For example
-managing users and groups. These roles implements a generic interface
+samba, ipa, ldap, keycloak), provide interface to manipulate the service. For
+example managing users and groups. These roles implements a generic interface
```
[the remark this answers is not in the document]
spoore1: I think that sentence continues on the next line with a reference to the class.

sssd_test_framework/hosts/__init__.py
spoore1: I think I first saw failures in the tox CI pipeline but, I was able to see the failures locally too.   I suspect it's related to the black version updates that was mentioned recently.   I added the commit with changes in order to correct the failure thinking that everyone would start seeing that.  I can drop those changes if we are changing the config or pinning black to a specific version as the fix.   

@justin-stephenson were you looking into the issue with black throwing errors recently?

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.adminpw = self.config.get("adminpw", "Secret123")
+
+        # Backup of original data
+        self.__backup: str | None = None
+
+    def kclogin(self) -> None:
```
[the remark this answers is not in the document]
spoore1: I modeled this after the IPA hosts class which has a kinit for admin.  This is basically the same.   

That being said, your suggestion would make the login more generally useful as I believe this could be used for login tests against keycloak for any user.   Should that be here or in a utils class like authentication?  Or could that wait for a later update when we add Keycloak specific tests as well?

sssd_test_framework/hosts/keycloak.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        Obtain ``admin`` user credentials for Keycloak
```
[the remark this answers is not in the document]
spoore1: fixing

sssd_test_framework/misc/__init__.py
spoore1: Yes,  this was the same as the hosts/__init__.py.   Let's get an answer there before I drop this one.

sssd_test_framework/roles/__init__.py
spoore1: Yes, this was the same as the hosts/init.py. Let's get an answer there before I drop this one.

sssd_test_framework/hosts/keycloak.py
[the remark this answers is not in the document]
danlavu: Does the function need to happen on the client or keycloak server? keycloak, here. Client, utils, under authentication probably like kerberos.

sssd_test_framework/roles/keycloak.py
```diff
@@ -0,0 +1,267 @@
...
+        :type name: str
+        :return: New group object.
+        :rtype: KeycloakGroup
+        """
+        return KeycloakGroup(self, name)
+
+
+class KeycloakObject(BaseObject[KeycloakHost, Keycloak]):
+    """
+    Keycloak Object
```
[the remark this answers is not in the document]
danlavu: I'm going to find as many ways to say period for the incoming reviews from the team. MWAHAHAHAHA.

sssd_test_framework/roles/keycloak.py
spoore1: 1.  No, kcadm is the command line interface to the Keycloak API.   It's needed for all the user/group commands as well as any other commands directly interfacing with Keycloak.   The service is handled by systemctl on the server in our setup on sssd-ci-containers and other CIs.

2.  As mentioned this does more than just start the process.   It's used by all the supporting methods that interact with Keycloak.   It's more like the IPA _exec I think.   Maybe I should rename it _kcadm?   I had left it without the underscore in case we wanted to allow other keycloak commands to be executed in tests by the role.

3.  Generally, I don't think so but, would that go in the role or host class?   Do other providers have a start/stop/restart build into the role/host?

sssd_test_framework/roles/keycloak.py
spoore1: removing

sssd_test_framework/roles/keycloak.py
spoore1: removing

sssd_test_framework/roles/keycloak.py
spoore1: removing.

sssd_test_framework/roles/keycloak.py
spoore1: fixing

sssd_test_framework/roles/keycloak.py
spoore1: fixing

sssd_test_framework/roles/keycloak.py
spoore1: fixing

sssd_test_framework/roles/keycloak.py
spoore1: deleting

sssd_test_framework/utils/__init__.py
spoore1: Yes, this was the same as the hosts/init.py. Let's get an answer there before I drop this one.

tests/test_misc.py
spoore1: Yes, this was the same as the hosts/init.py. Let's get an answer there before I drop this one.

sssd_test_framework/hosts/keycloak.py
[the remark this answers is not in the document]
spoore1: Keycloak server.

sssd_test_framework/roles/keycloak.py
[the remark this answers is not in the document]
spoore1: > I'm going to find as many ways to say period for the incoming reviews from the team. MWAHAHAHAHA.

LOL,  time to learn "period" in different languages?

sssd_test_framework/hosts/__init__.py
justin-stephenson: https://github.com/SSSD/sssd/pull/7155 was merged to address black changes in sssd tests, but haven't pinned to a specific version yet. Changes to linting requirements are due to https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html

sssd_test_framework/hosts/keycloak.py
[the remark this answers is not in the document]
danlavu: They should be in the keycloak role, then. The host, only really contains backup, restore methods and passes variables to the role.

sssd_test_framework/hosts/__init__.py
spoore1: @justin-stephenson Ok, so we're seeing black failures in other PRs as well that need something like the commit I have here.  Should I pull that our of this PR and add it to a separate one?

justin-stephenson: A separate PR would be nice if it affects files unrelated to this keycloak PR.

sssd_test_framework/hosts/keycloak.py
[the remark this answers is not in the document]
danlavu: I don't know, all the servers we use are containers, so the setup is already done, we're just passing variables to configure them from hosts. 

I think install/setup should be in hosts, under init. When the host is selected the service is already configured. So maybe we remove kclogin and make dump it in init instead?

sssd_test_framework/roles/keycloak.py
danlavu: Okay, so it's like "ipa" command, this is good then, I think. I'd update the docstring to describe that the kcadm command is. The command to interface with the keycloak API: create users, etc.

sssd_test_framework/utils/__init__.py
danlavu: I think you have a ghost in your shell.

sssd_test_framework/hosts/__init__.py
spoore1: ok, I created PR #75 for that an removed the commit from this PR.

sssd_test_framework/utils/__init__.py
spoore1: removed the commit with this so resolving.

sssd_test_framework/roles/__init__.py
spoore1: dropped the commit

sssd_test_framework/misc/__init__.py
spoore1: dropped the commit

sssd_test_framework/hosts/keycloak.py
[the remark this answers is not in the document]
spoore1: It's used in the keycloak role setup() the same way the ipa role uses the ipa hosts kinit in the ipa role setup.  I just followed that pattern.   I do believe it's needed every time the role is instantiated to make sure the login is done before running the kcadm commands against keycloak.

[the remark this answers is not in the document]
danlavu: Scott chatted about this I misunderstood the purpose of this function.

danlavu (approved):
This looks good to me.

pbrezina (approved):

PR #40, opened 2023-09-05
sssctl: user-show command added

patriki01:
I added function to call ``sssctl user-show`` with additional arguments.
Waiting for your comments

### DIFF
--- a/sssd_test_framework/utils/sssctl.py
+++ b/sssd_test_framework/utils/sssctl.py
@@ -188,6 +188,29 @@ class SSSCTLUtils(MultihostUtility[MultihostHost]):
         """
         return self.host.ssh.exec(["sssctl", "user-checks", username, "-a", action, "-s", service])
 
+    def user_show(self, user: str | None = None, sid: str | None = None, uid: int | None = None) -> SSHProcessResult:
+        """
+        Information about cached user
+
+        :param user: User that will be showed, defaults to None
+        :type user: str | None
+        :param sid: Search by SID, defaults to None
+        :type sid: str | None
+        :param uid: Search by user ID, defaults to None
+        :type uid: int | None
+        :return: Result of called command
+        :rtype: SSHProcessResult
+        """
+        options = []
+        if user is not None:
+            options += [user]
+        if sid is not None:
+            options += ["-s", sid]
+        if uid is not None:
+            options += ["-u", str(uid)]
+
+        return self.host.ssh.exec(["sssctl", "user-show", *options])
+
     def config_check(self, config: str | None = None, snippet: str | None = None) -> SSHProcessResult:
         """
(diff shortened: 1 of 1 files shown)


sssd_test_framework/utils/sssctl.py
```diff
@@ -188,6 +188,29 @@ def user_checks(self, username: str, action: str = "acct", service: str = "syste
         """
         return self.host.ssh.exec(["sssctl", "user-checks", username, "-a", action, "-s", service])
 
+    def user_show(self, user: str | None = None, sid: str | None = None, uid: str | None = None) -> SSHProcessResult:
```
pbrezina, line 191: I would also accept uid as int.

pbrezina (changes requested):

sssd_test_framework/utils/sssctl.py
patriki01: That makes sense. Probably SID can be int as well.

ikerexxe: I don't think SID can be an `int` as well as it contains letters and hyphens. Example:
```
S-1-5-21-1004336348-1177238915-682003330-512
```

patriki01: Oh I see. I will update it. Thanks for info

ikerexxe (dismissed):
LGTM. Thanks for the patch!

sssd_test_framework/utils/sssctl.py
```diff
@@ -188,6 +188,29 @@ def user_checks(self, username: str, action: str = "acct", service: str = "syste
...
+        :param sid: Search by SID, defaults to None
+        :type sid: str | None
+        :param uid: Search by user ID, defaults to None
+        :type uid: int | None
+        :return: Result of called command
+        :rtype: SSHProcessResult
+        """
+        options = []
+        if user is not None:
+            options = [user]
```
pbrezina, line 206: `options += [user]`

sssd_test_framework/utils/sssctl.py
```diff
@@ -188,6 +188,29 @@ def user_checks(self, username: str, action: str = "acct", service: str = "syste
...
+        if sid is not None:
+            options = options + ["-s", sid]
```
pbrezina, line 208: `options += ...`

sssd_test_framework/utils/sssctl.py
```diff
@@ -188,6 +188,29 @@ def user_checks(self, username: str, action: str = "acct", service: str = "syste
...
+        if uid is not None:
+            options = options + ["-u", str(uid)]
```
pbrezina, line 210: `options += ...`

pbrezina (approved):

PR #45, opened 2023-09-18
ipa: add support for external group members

pbrezina:


ikerexxe:
Question for testing purposes. `ipa.group("thegrup").add_member("--external sometext")` can be used to test this logic, right?

pbrezina:
`.add_member("user@domain")`, e.g. `.add_member("Administrator@samba.test")`

ikerexxe (approved):
I knew I was doing something wrong, thanks for the examples. LGTM!

pbrezina:
I had it this way originally, then switched. I'm fine both ways. I guess this depends on how familiar are you with IPA, it uses the same command as IPA users and group and you can add all in single command. I.e. `ipa group-add-member group --users ... --groups ... --external ...`. That's why I merged it in the end and just put it into method description. If we go with two methods, then you need to add two calls to do the same thing.

It's your call.

justin-stephenson:
> I had it this way originally, then switched. I'm fine both ways. I guess this depends on how familiar are you with IPA, it uses the same command as IPA users and group and you can add all in single command. I.e. `ipa group-add-member group --users ... --groups ... --external ...`. That's why I merged it in the end and just put it into method description. If we go with two methods, then you need to add two calls to do the same thing.
> 
> It's your call.

Okay let's keep it as is. Thanks for the explanation.

justin-stephenson (approved):
Ack, thank you.

PR #46, opened 2023-09-19
topology: add ipa trust topologies

pbrezina:


pbrezina:
Please, update your pytest-mh before testing this. IPA code triggered infinite recursion on setup_when_called: https://github.com/next-actions/pytest-mh/commit/dc6b5be2cfe91ee9df7169570878285d81eb4767

pbrezina:
Test example requires https://github.com/SSSD/sssd-test-framework/pull/45.

pbrezina:
#45 was pushed, but it also requires https://github.com/SSSD/sssd-test-framework/pull/47 now.

sssd_test_framework/topology.py
```diff
@@ -213,3 +233,8 @@ def test_ldap(client: Client, provider: GenericProvider):
     """
     .. topology-mark:: KnownTopologyGroup.AnyAD
     """
+
+    IPATrust = [KnownTopology.IPATrustAD, KnownTopology.IPATrustSamba]
```
ikerexxe, line 237: To be consistent I'd name this topology `AnyIPATrust`

ikerexxe: If you change it, update the documentation as well

ikerexxe (commented):
One minor comment that affects both the code and the documentation

sidecontrol (approved):

sssd_test_framework/topology.py
pbrezina: Hmm. I don't think we need to follow "Any" as a specific pattern and I'm still trying to figure out if it is helpful here, I don't have problem with either. If it is just to "follow a pattern" I'd say there is not pattern and it is just and we don't require each name in KnownTopologyGroup to start with Any. If you have a reason, I'll change it.

AnyProvider is I think clear, "Any" has its own meaning. For AnyADProvider, I think we could have go with just "ADProvider" but "Any" kind of denotes that there are some limitations of what you can do (Samba is subset of AD). With IPA trust, you can only work with users and groups anyway and there are no limitations here since both AD and Samba behaves the same. So I guess that's why I didn't add "Any" here.

But I can go with either, it's your call.

ikerexxe: It's ok, I just wanted to raise that if we were following a pattern we should also follow it here. Your explanation is clear enough that there isn't any pattern to follow for this case

ikerexxe (approved):
LGTM. Thanks for the patch!

PR #48, opened 2023-09-21
Add topology to feature detection

ikerexxe:
A feature may have dependencies in several hosts, so we need to check
for it in all of them. Adding topology information to the feature
detection mechanism.

If client detection is present it always needs to be checked, other
hosts will be skipped if they aren't present in the topology.

Moreover, add passkey feature detection in ldap.

Depends on https://github.com/next-actions/pytest-mh/pull/18

### DIFF
--- a/sssd_test_framework/hosts/ldap.py
+++ b/sssd_test_framework/hosts/ldap.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import ldap
+from pytest_mh.ssh import SSHLog
 
 from .base import BaseLDAPDomainHost
 
@@ -25,12 +26,43 @@ class LDAPHost(BaseLDAPDomainHost):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
 
+        self._features: dict[str, bool] | None = None
+
         # Additional client configuration
         self.client.setdefault("id_provider", "ldap")
         self.client.setdefault("ldap_uri", f"ldap://{self.hostname}")
 
         # Backup of original data
         self.__backup: dict[str, dict[str, list[bytes]]] = {}
+
+    @property
+    def features(self) -> dict[str, bool]:
+        """
+        Features supported by the host.
+        """
+        if self._features is not None:
+            return self._features
+
+        self.logger.info(f"Detecting features on {self.hostname}")
+
+        result = self.ssh.run(
+            """
+            set -ex
+
+            grep -r "passkey" /etc/dirsrv/ &> /dev/null && echo "passkey" || :
+            """,
+            log_level=SSHLog.Error,
+        )
+
+        # Set default values
+        self._features = {
+            "passkey": False,
+        }
+
+        self._features.update({k: True for k in result.stdout_lines})
+        self.logger.info("Detected features:", extra={"data": {"Features": self._features}})
+
+        return self._features
 
     def backup(self) -> None:
         """
--- a/sssd_test_framework/markers.py
+++ b/sssd_test_framework/markers.py
@@ -5,6 +5,9 @@ from __future__ import annotations
 from functools import partial
 
 import pytest
+from pytest_mh import MultihostItemData, Topology
+
+from sssd_test_framework.topology import KnownTopology, KnownTopologyGroup
 
 from .misc import to_list_of_strings
 from .roles.base import BaseRole
@@ -56,6 +59,8 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
     if not isinstance(item, pytest.Function):
         raise TypeError(f"Unexpected item type: {type(item)}")
 
+    topology: list[Topology] = []
+    mh_item_data: MultihostItemData | None = MultihostItemData.GetData(item)
     for mark in item.iter_markers("builtwith"):
         requirements: dict[str, str] = {}
 
@@ -63,11 +68,34 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
             # @pytest.mark.builtwith("files-provider")
             #  -> check if "files-provider" is supported by client
             requirements["client"] = mark.args[0]
+            topology = []
         elif not mark.args and mark.kwargs:
             # @pytest.mark.builtwith(client="passkey", ipa="passkey") ->
             # -> check if "passkey" is supported by both client and ipa
             requirements = dict(mark.kwargs)
+            topology = []
+        elif (
+            len(mark.args) == 1
+            and isinstance(mark.args[0], (Topology, KnownTopology, KnownTopologyGroup))
+            and mark.kwargs
+        ):
+            # @pytest.mark.builtwith(KnownTopology.IPA, ipa="passkey") ->
+            # -> check if "passkey" is supported by ipa only if the test runs on IPA topology
+            requirements = dict(mark.kwargs)
+            if isinstance(mark.args[0], Topology):
+                topology = [mark.args[0]]
+            elif isinstance(mark.args[0], KnownTopology):
+                topology = [mark.args[0].value.topology]
+            elif isinstance(mark.args[0], KnownTopologyGroup):
+                topology = [x.value.topology for x in mark.args[0].value]
         else:
             raise ValueError(f"{item.nodeid}::{item.originalname}: invalid arguments for @pytest.mark.builtwith")
 
-        item.add_marker(pytest.mark.require(partial(builtwith, item=item, requirements=requirements)))
+        if mh_item_data is None:
+            raise ValueError(f"{item.nodeid}::{item.originalname}: multihost item data is not set")
+
+        if mh_item_data.topology_mark is None:
+            raise ValueError(f"{item.nodeid}::{item.originalname}: multihost topology mark is not set")
+
+        if not topology or mh_item_data.topology_mark.topology in topology:
+            item.add_marker(pytest.mark.require(partial(builtwith, item=item, requirements=requirements)))


justin-stephenson:
The conditional checks in this loop are quite confusing to me. Can you briefly explain how it works?  How will it scale for when we have several more markers added ?

```
    for mark in item.iter_markers("builtwith"):
        requirements: dict[str, str] = {}

        if len(mark.args) == 1 and not mark.kwargs:
            # @pytest.mark.builtwith("files-provider")
            #  -> check if "files-provider" is supported by client
            requirements["client"] = mark.args[0]
        elif not mark.args and mark.kwargs:
            # @pytest.mark.builtwith(client="passkey", ipa="passkey") ->
            # -> check if "passkey" is supported by both client and ipa
            requirements = dict(mark.kwargs)
        elif len(mark.args) == 1 and mark.args:
            # @pytest.mark.builtwith(KnownTopology.IPA, ipa="passkey") ->
            # -> check if "passkey" is supported by ipa
            for iterargs in mark.args:
                topology = iterargs
        else:
            raise ValueError(f"{item.nodeid}::{item.originalname}: invalid arguments for @pytest.mark.builtwith")

```

ikerexxe:
> The conditional checks in this loop are quite confusing to me. Can you briefly explain how it works? How will it scale for when we have several more markers added ?

Do you have any idea of a marker that we could add in the future? I mean, I don't expect any other marker to be added, but you never know.

justin-stephenson:
> > The conditional checks in this loop are quite confusing to me. Can you briefly explain how it works? How will it scale for when we have several more markers added ?
> 
> Do you have any idea of a marker that we could add in the future? I mean, I don't expect any other marker to be added, but you never know.

I'm not sure but if Pavel is okay with it then I don't oppose, mostly I just found the following quite difficult to understand how it relates to various `builtwith` features:

        if len(mark.args) == 1 and not mark.kwargs:
        ...
        elif not mark.args and mark.kwargs:
        ...
        elif len(mark.args) == 1 and mark.args:

ikerexxe:
> I think markers and features will be needed to distinguish different AD topologies. I'll worry about that though, just an FYI.

I think that this is already covered with the current proposal. Let's say that you want to test something that need one feature available in the client and two in AD. That can be done with:
```
@pytest.mark.builtwith(client="feature1")
@pytest.mark.builtwith(KnownTopology.IPATrustAD, ad="feature1")
@pytest.mark.builtwith(KnownTopology.IPATrustAD, ad="feature2")
``` 

> I'm not sure but if Pavel is okay with it then I don't oppose, mostly I just found the following quite difficult to understand how it relates to various `builtwith` features:

I also had problems understanding it at the beginning. The comments after each condition help a lot.


By the way, CI reports an error that in my limited knowledge of python I don't know how to fix.

pbrezina:
> The conditional checks in this loop are quite confusing to me. Can you briefly explain how it works? How will it scale for when we have several more markers added ?

I didn't review the code yet, I'm just replying to your question. Don't be confused with "different methods to setup single marker" (i.e. different types of arguments) and "two distinguished markers".

We can add any marker we need, independent on `builtswith` and as long as we can keep distinguishing between different parameter types for `builtswith` we can keep extending it.


> ```
> @pytest.mark.builtwith(client="feature1")
> @pytest.mark.builtwith(KnownTopology.IPATrustAD, ad="feature1")
> @pytest.mark.builtwith(KnownTopology.IPATrustAD, ad="feature2")
> ```

This can be also written as:

```python
@pytest.mark.builtwith(client="feature1")
@pytest.mark.builtwith(KnownTopology.IPATrustAD, ad=["feature1", "feature2"])
```

> By the way, CI reports an error that in my limited knowledge of python I don't know how to fix.

This is because pytest-mh pull request was not yet merged.

sssd_test_framework/markers.py
```diff
@@ -22,10 +24,17 @@ def pytest_configure(config: pytest.Config):
     )
 
 
-def builtwith(item: pytest.Function, requirements: dict[str, str], **kwargs: BaseRole):
+def builtwith(item: pytest.Function, requirements: dict[str, str], topology: KnownTopology, **kwargs: BaseRole):
```
pbrezina, line 27: this should take type `Topology | None`, not `KnownTopology`

sssd_test_framework/markers.py
```diff
@@ -22,10 +24,17 @@ def pytest_configure(config: pytest.Config):
...
     def value_error(msg: str) -> ValueError:
         return ValueError(f"{item.nodeid}::{item.originalname}: @pytest.mark.builtwith: {msg}")
 
+    mh = item.funcargs["mh"]
+
+    # Skip feature detection in topologies that are not running
+    # Always check feature detection in client
+    if "client" not in requirements and topology and topology.value.topology != mh.topology:
```
pbrezina, line 35: Do not specialize client. Rather check that `topology is None or not matching`

sssd_test_framework/markers.py
```diff
@@ -67,7 +77,13 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
             # @pytest.mark.builtwith(client="passkey", ipa="passkey") ->
             # -> check if "passkey" is supported by both client and ipa
             requirements = dict(mark.kwargs)
+        elif len(mark.args) == 1 and mark.args:
```
pbrezina, line 80: `elif len(mark.args) == 1 and isintance(mark.args[0], (Topology, KnownTopology, KnownTopologyGroup) and mark.kwargs`

sssd_test_framework/markers.py
```diff
@@ -67,7 +77,13 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
...
+            # @pytest.mark.builtwith(KnownTopology.IPA, ipa="passkey") ->
+            # -> check if "passkey" is supported by ipa
+            for iterargs in mark.args:
```
pbrezina, line 83:
```python
if isinstance(mark.args[0], Topology):
    topology = [mark.args[0]]
elif isinstance(mark.args[0], KnownTopology):
    topology = mark.arkgs[0].value
elif isintance(marks.args[0], KnownTopologyGroup):
    topology = [x.value for x in marks.args[0]]
```

sssd_test_framework/markers.py
```diff
@@ -56,6 +65,7 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
     if not isinstance(item, pytest.Function):
         raise TypeError(f"Unexpected item type: {type(item)}")
 
+    topology = None
```
pbrezina, line 68: `topology: list[Topology] = []`

sssd_test_framework/markers.py
```diff
@@ -67,7 +77,13 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
             # @pytest.mark.builtwith(client="passkey", ipa="passkey") ->
             # -> check if "passkey" is supported by both client and ipa
             requirements = dict(mark.kwargs)
+        elif len(mark.args) == 1 and mark.args:
+            # @pytest.mark.builtwith(KnownTopology.IPA, ipa="passkey") ->
+            # -> check if "passkey" is supported by ipa
```
pbrezina, line 82: ... only if the test runs on IPA topology

sssd_test_framework/markers.py
```diff
@@ -67,7 +77,13 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
...
+            for iterargs in mark.args:
+                topology = iterargs
         else:
             raise ValueError(f"{item.nodeid}::{item.originalname}: invalid arguments for @pytest.mark.builtwith")
 
-        item.add_marker(pytest.mark.require(partial(builtwith, item=item, requirements=requirements)))
+        item.add_marker(pytest.mark.require(partial(
```
pbrezina, line 88:
```python
if not topology:
    topology = [None]

for t in topology:
   iteam.add_marker(...)
```

pbrezina (changes requested):
Hi, see comments inside to fix your code.

I think we can also simplify it a lot more, because the topology information is actually available in `item`  parameter from `pytest_runtest_setup`. `pytest_runtest_setup` is already called on the test case with associated topology so don't have to propage this information to the marker but we can add the marker only if the topology match.

sssd_test_framework/markers.py
pbrezina: Also add `mh: MultihostFixture` to parameters and ditch `mh = item.funcargs["mh"]`

ikerexxe:
> I think we can also simplify it a lot more, because the topology information is actually available in `item` parameter from `pytest_runtest_setup`. `pytest_runtest_setup` is already called on the test case with associated topology so don't have to propage this information to the marker but we can add the marker only if the topology match. API to access multihost data is private currently, we would need to make `MultihostItemData` public in pytest-mh.

You mean, checking the topology in `pytest_runtest_setup` instead of `builtwith` function, right?

I think I was able to make that API public, but check it just in case: https://github.com/next-actions/pytest-mh/pull/19

sssd_test_framework/markers.py
pbrezina: args of course. args are positional arguments (list), kwargs are keyword arguments (dictionary)

pbrezina:
> > I think we can also simplify it a lot more, because the topology information is actually available in `item` parameter from `pytest_runtest_setup`. `pytest_runtest_setup` is already called on the test case with associated topology so don't have to propage this information to the marker but we can add the marker only if the topology match. API to access multihost data is private currently, we would need to make `MultihostItemData` public in pytest-mh.
> 
> You mean, checking the topology in `pytest_runtest_setup` instead of `builtwith` function, right?

Right. So the way it works is that pytest_mh takes the test case, reads topology marker and duplicates it per each topology. That is you get `test * topologies` number of tests, each has one topology assigned. pytest_runtest_setup is called for each of these duplicates individually.

So we can just read the topology of the item and add `require` marker only if it actually applies to this topology directly in pytest_runtest_setup.

> I think I was able to make that API public, but check it just in case: [next-actions/pytest-mh#19](https://github.com/next-actions/pytest-mh/pull/19)

Should work.

> By the way, which member of item contains the topology data?

You need to access it with this function: https://github.com/next-actions/pytest-mh/blob/master/pytest_mh/_private/data.py#L35

Example: https://github.com/next-actions/pytest-mh/blob/master/pytest_mh/_private/fixtures.py#L311

sssd_test_framework/markers.py
ikerexxe: I was asking because in your [previous comment](https://github.com/SSSD/sssd-test-framework/pull/48#discussion_r1334272951) you requested me to check `mark.kwargs` in the condition, but this variable isn't used anywhere.

ikerexxe: Disregard my last comment

sssd_test_framework/hosts/ldap.py
```diff
@@ -25,13 +27,45 @@ class LDAPHost(BaseLDAPDomainHost):
...
+        if self._features is not None:
+            return self._features
+
+        self.logger.info(f"Detecting features on {self.hostname}")
+
+        result = self.ssh.run(
+            """
+            set -ex
+
+            grep "passkey" /etc/dirsrv/slapd-localhost/schema/99user.ldif &> /dev/null && \
```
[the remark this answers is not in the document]
ikerexxe: I removed the instance part, so fixed

sssd_test_framework/markers.py
ikerexxe: I kept this part as it was before the PR because I moved all the checks to `pytest_runtest_setup()`, as you suggested in another comment

sssd_test_framework/markers.py
ikerexxe: Done

sssd_test_framework/markers.py
ikerexxe: Well, not exactly the same but I implemented something similar, so I'm closing this one also

sssd_test_framework/markers.py
```diff
@@ -56,18 +61,39 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
...
+                topology = [mark.args[0]]
+            elif isinstance(mark.args[0], KnownTopology):
+                topology = [mark.args[0].value]
+            elif isinstance(mark.args[0], KnownTopologyGroup):
+                topology = [x.value for x in mark.args[0]]
         else:
             raise ValueError(f"{item.nodeid}::{item.originalname}: invalid arguments for @pytest.mark.builtwith")
 
-        item.add_marker(pytest.mark.require(partial(builtwith, item=item, requirements=requirements)))
+        if not topology:
```
pbrezina, line 92: I think you can simplify it as:

```python
if not topology or mh.topology_mark in topology:
    item.add_marker(...)
```

sssd_test_framework/markers.py
```diff
@@ -56,18 +61,39 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
     if not isinstance(item, pytest.Function):
         raise TypeError(f"Unexpected item type: {type(item)}")
 
+    topology: list[Topology] = []
+    mh: MultihostItemData | None = MultihostItemData.GetData(item)
```
pbrezina, line 65: mh_item_data so one does not confuse it with mh which is the MultihostFixture

sssd_test_framework/markers.py
ikerexxe: Fixed

sssd_test_framework/markers.py
ikerexxe: Fixed

ikerexxe:
Depends on https://github.com/next-actions/pytest-mh/pull/20 to use:
```
@pytest.mark.builtwith(KnownTopology.LDAP, ldap="passkey")
```

Everything else fixed in this PR.

pbrezina:
I applied the following diff to fix mypy complains that actually revealed a bug in how we handled KnownTopologyGroup. Now it is functional. I took the liberty and pushed this into your branch.

```python
diff --git a/sssd_test_framework/markers.py b/sssd_test_framework/markers.py
index 1dffe55..67fdc87 100644
--- a/sssd_test_framework/markers.py
+++ b/sssd_test_framework/markers.py
@@ -85,11 +85,17 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
             if isinstance(mark.args[0], Topology):
                 topology = [mark.args[0]]
             elif isinstance(mark.args[0], KnownTopology):
-                topology = [mark.args[0].value]
+                topology = [mark.args[0].value.topology]
             elif isinstance(mark.args[0], KnownTopologyGroup):
-                topology = [x.value for x in mark.args[0]]
+                topology = [x.value.topology for x in mark.args[0].value]
         else:
             raise ValueError(f"{item.nodeid}::{item.originalname}: invalid arguments for @pytest.mark.builtwith")
 
-        if not topology or mh_item_data.topology_mark in topology:
+        if mh_item_data is None:
+            raise ValueError(f"{item.nodeid}::{item.originalname}: multihost item data is not set")
+
+        if mh_item_data.topology_mark is None:
+            raise ValueError(f"{item.nodeid}::{item.originalname}: multihost topology mark is not set")
+
+        if not topology or mh_item_data.topology_mark.topology in topology:
             item.add_marker(pytest.mark.require(partial(builtwith, item=item, requirements=requirements)))

```

pbrezina (approved):

ikerexxe:
Awesome! Thank you for the fixes :smile:

PR #53, opened 2023-10-12
MISC: attrs_parse() fixed error when multi-line value appeared

patriki01:
When LDAP attributes values are too long, they split into multiple lines. It can look like the example provided, where ``originalDN`` is the problematic attribute:
```
# record 11
dn: name=ae3c2c8c-4b8f-11ee-ac71-c6e1a1a33042,cn=hbac_rules,cn=custom,cn=test,cn=sysdb
originalDN: ipaUniqueID=ae3c2c8c-4b8f-11ee-ac71-c6e1a1a33042,cn=hbac,dc=ipa,dc
 =test
serviceCategory: all
```

And this caused an error in ``attrs_parse()`` function. I added two tests to check this error, so look there for better understanding of the problem.
This is not a very nice solution, but I couldn't think of a better one. Feel free to suggest changes.

### DIFF
--- a/sssd_test_framework/misc/__init__.py
+++ b/sssd_test_framework/misc/__init__.py
@@ -17,16 +17,25 @@ def attrs_parse(lines: list[str], attrs: list[str] | None = None) -> dict[str, l
     :rtype: dict[str, list[str]]
     """
     out: dict[str, list[str]] = {}
-    for line in lines:
-        line = line.strip()
+    i = 0
+    while i < len(lines):
+        line = lines[i]
         if not line:
+            i += 1
             continue
 
-        (key, value) = map(lambda x: x.strip(), line.split(":", 1))
+        (key, value) = map(lambda x: x.lstrip(), line.split(":", 1))
+        while i < len(lines) - 1:
+            if lines[i + 1].startswith(" "):
+                value += lines[i + 1][1:]
+                i += 1
+            else:
+                break
+
         if attrs is None or key in attrs:
             out.setdefault(key, [])
             out[key].append(value)
-
+        i += 1
     return out
 
 
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -20,9 +20,8 @@ def test_attrs_parse__nofilter():
     param1: value1
     param2: value2
     param3: value3
-    """.split(
-        "\n"
-    )
+    """
+    lines = textwrap.dedent(lines).strip().split("\n")
 
     expected = {
         "param1": ["value1"],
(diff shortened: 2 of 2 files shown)


sssd_test_framework/misc/__init__.py
```diff
@@ -17,16 +17,24 @@ def attrs_parse(lines: list[str], attrs: list[str] | None = None) -> dict[str, l
     :rtype: dict[str, list[str]]
     """
     out: dict[str, list[str]] = {}
-    for line in lines:
-        line = line.strip()
+    i = 0
```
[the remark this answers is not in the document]
patriki01: Thanks Tomas for info, I did not know that. I will work on this and let you know when it is done.

patriki01:
I improved the code and also added the test case you mentioned.

sssd_test_framework/misc/__init__.py
```diff
@@ -17,16 +17,24 @@ def attrs_parse(lines: list[str], attrs: list[str] | None = None) -> dict[str, l
...
+    while i < len(lines):
+        line = lines[i].strip()
```
thalman, line 22: This way it is much better. But you are still loosing trailing spaces. Do not strip here.

tests/test_misc.py
```diff
@@ -51,6 +48,34 @@ def test_attrs_parse__filter():
...
+    [
+        (
+            ["cn: sudorules", "distinguishedName: objectSID=123,cn=id_m", " appings,cn=test,cn=sysdb"],
+            {"cn": ["sudorules"], "distinguishedName": ["objectSID=123,cn=id_mappings,cn=test,cn=sysdb"]},
+        ),
+        (
+            [
+                "description: My teacher is a goo",
+                " d one but I do not like him",
+                "  very much: he wears a dirty coat.",
```
thalman, line 65: Can you update the string like this to cover trailing spaces?

```
            [
                "description: My teacher is a goo",
                " d one but I do not like him",
                "  very much: he wears a dirty ",
                " coat.",
            ],
```

sssd_test_framework/misc/__init__.py
```diff
@@ -17,16 +17,24 @@ def attrs_parse(lines: list[str], attrs: list[str] | None = None) -> dict[str, l
...
         if not line:
+            i += 1
             continue
 
         (key, value) = map(lambda x: x.strip(), line.split(":", 1))
```
thalman, line 27: use x.lstrip() so we do not loose trailing space. Leading space here is not allowed so that can be stripped.

JFI String that starts with space must be encoded (base64?) and there will be the double colon like this:

    dn:: SFSDIHDKFIEH=

tests/test_misc.py
patriki01: Yes, but this way test still passes, because in inner while there is no strip. I change it a bit more so it fails on this implementation.

patriki01:
Feedback resolved

sssd_test_framework/misc/__init__.py
thalman: ack

thalman (approved):
Thanks @patriki01 for the effort. Looks great now :)
ACK

pbrezina (approved):

PR #54, opened 2023-10-23
Tools: journalctl added

patriki01:
I added command for ``journalctl``.  I included only what I thought are most commonly used parameters. Feel free to suggest change.

### DIFF
--- a/sssd_test_framework/utils/tools.py
+++ b/sssd_test_framework/utils/tools.py
@@ -6,6 +6,7 @@ from typing import Any
 
 import jc
 from pytest_mh import MultihostHost, MultihostUtility
+from pytest_mh.cli import CLIBuilder, CLIBuilderArgs
 from pytest_mh.ssh import SSHProcess, SSHProcessResult
 from pytest_mh.utils.fs import LinuxFileSystem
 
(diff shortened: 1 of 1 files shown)


danlavu:
It'd be nice to add a doc entry for this and provide the following examples, 

1. Typical usage, getting output from journalctl
2. Assertion usage, how to use it and search for a string in the journal for a test case 

assert client.utils.tools.journalctl(grep="search string").rc == 0
assert client.utils.tools.journal(grep="bad string").rc not 0

patriki01:
I added example usage as you proposed.

ikerexxe (approved):
I'd also add the `--system` and `--user` options, I think I'm using them quite often.

However, this is only an optional improvement so I'm approving this PR.

patriki01:
I added those options @ikerexxe

ikerexxe:
@danlavu, @pbrezina do you mind reviewing this PR?

danlavu:
@pbrezina  ack

@patriki01 I suppose we can take the comment about the assertions out then.

patriki01:
Okay, I move it there: https://github.com/next-actions/pytest-mh/pull/23

PR #55, opened 2023-10-25
passkey: avoid an infinite loop in cares in authentication of passkey

madhuriupadhye:
The issue is an infinite loop in cares.
generate_unique_id() caused by 'LD_PRELOAD=/opt/random.so'.
generate_unique_id() is calling arc4random_buf() and the loop in cares
is keeping a list of old ids to avoid those. But arc4random_buf() is
overwritten by random.so and always returns the same value and as a
result the same id is always used and causes the infinite loop.

To make the environment only available to
passkey_child not to add those environment variable to
/etc/sysconfig/sssd but rename passkey_child.

Signed-off-by: Madhuri Upadhye <mupadhye@redhat.com>

### DIFF
--- a/sssd_test_framework/utils/authentication.py
+++ b/sssd_test_framework/utils/authentication.py
@@ -320,22 +320,28 @@ class SUAuthenticationUtils(MultihostUtility[MultihostHost]):
         :return: True if authentication was successful, False otherwise.
         :rtype: bool
         """
-        self.fs.backup("/etc/sysconfig/sssd")
+        self.fs.backup("/usr/libexec/sssd/passkey_child")
+        self.fs.copy("/usr/libexec/sssd/passkey_child", "/usr/libexec/sssd/passkey_child.orig")
+
         device_path = self.fs.upload_to_tmp(device, mode="a=r")
         ioctl_path = self.fs.upload_to_tmp(ioctl, mode="a=r")
         script_path = self.fs.upload_to_tmp(script, mode="a=r")
 
         run_su = self.fs.mktmp(
             rf"""
-            #!/bin/bash
-            set -ex
-            env | grep ^UMOCKDEV_ > /etc/sysconfig/sssd
-            printf "LD_PRELOAD=$LD_PRELOAD" >> /etc/sysconfig/sssd
-            systemctl restart sssd
-            chmod -R a+rwx $UMOCKDEV_DIR
+                #!/bin/bash
+                set -ex
+                echo '#!/bin/bash' > /usr/libexec/sssd/passkey_child
+                echo -n 'export ' >> /usr/libexec/sssd/passkey_child
+                env | grep ^UMOCKDEV_ >> /usr/libexec/sssd/passkey_child
+                echo -n 'export ' >> /usr/libexec/sssd/passkey_child
+                printf "LD_PRELOAD=$LD_PRELOAD\n" >> /usr/libexec/sssd/passkey_child
+                echo 'exec /usr/libexec/sssd/passkey_child.orig $@' >> /usr/libexec/sssd/passkey_child
+                chmod 755 /usr/libexec/sssd/passkey_child
+                chmod -R a+rwx $UMOCKDEV_DIR
 
-            su --shell /bin/sh nobody -c "su - '{username}'"
-            """,
+                su --shell /bin/sh nobody -c "su - '{username}'"
+                """,
             mode="a=rx",
         )
 
@@ -383,6 +389,8 @@ class SUAuthenticationUtils(MultihostUtility[MultihostHost]):
             exit 203
             """
         )
+
+        self.fs.restore("/usr/libexec/sssd/passkey_child")
 
         if result.rc > 200:
             raise ExpectScriptError(result.rc)
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 jc
 pytest
 python-ldap
-pytest-mh >= 1.0.5
+pytest-mh >= 1.0.7


sssd_test_framework/utils/authentication.py
```diff
@@ -324,18 +324,26 @@ def passkey(self, username: str, *, pin: str | int, device: str, ioctl: str, scr
         device_path = self.fs.upload_to_tmp(device, mode="a=r")
         ioctl_path = self.fs.upload_to_tmp(ioctl, mode="a=r")
         script_path = self.fs.upload_to_tmp(script, mode="a=r")
+        self.host.ssh.run(
+            "mv -f /usr/libexec/sssd/passkey_child /usr/libexec/sssd/passkey_child.orig", raise_on_error=False
```
sumit-bose, line 328: Hi,

you might run into issues if this is called multiple times in a test run because at the second call the script will be moved to `/usr/libexec/sssd/passkey_child.orig` which will then call itself recursively. Maybe it would be possible to use the backup/restore functionality for this, i.e. add `/usr/libexec/sssd/passkey_child` to the `backup()` and `restore()` calsl in `sssd_test_framework/hosts/client.py`. Then the `/usr/libexec/sssd/passkey_child` would be restored after each test. @pbrezina, is this a proper use of the backup/restore functionality or should the test do the cleanup itself, i.e. make sure `/usr/libexec/sssd/passkey_child.orig"` is moved back to `/usr/libexec/sssd/passkey_child`?

bye,
Sumit

sumit-bose: Hi,

thanks for the update, I'm fine with the change but please remove the old, commented-out part. @pbrezina, is it ok to use the backup/restore functionality this way?

bye,
Sumit

sssd_test_framework/hosts/client.py
```diff
@@ -140,6 +140,7 @@ def backup(self) -> None:
             backup /etc/sssd "{location}/config"
             backup /var/log/sssd "{location}/logs"
             backup /var/lib/sss "{location}/lib"
+            backup /usr/libexec/sssd/passkey_child "{location}/passkey_child.orig"
```
spoore1, line 143: Is passkey_child always present or should the backup/restore check that the feature is enabled?

sssd_test_framework/utils/authentication.py
madhuriupadhye: Hello,

I am facing another issue, 
like if I backup and restore the from using methods form file `hosts/client.py` then tests failing, as I added only one thing is either truncate the `/usr/libexec/sssd/passkey_child` or completely remove it before entering in method  using `rm -f /usr/libexec/sssd/passkey_child` but still tests failed,
but If I add simple `mv -f` before and after the methods its working.

@pbrezina can you help here for appropriate approach, I tested this locally.

pbrezina:
Hi, I added `copy` and `restore` methods to pytest-mh fs. If I understand it correctly we want to:

1. Copy passkey_child -> passkey_child.orig
2. Mock passkey_child (write custom contents, exec passkey_child.orig)
3. Text authentication
4. Restore paskey_child and passkey_child.orig

The code should look like:

```python
 def passkey(...):
    ...
    self.fs.copy("passkey_child", "passkey_child.orig")
    self.fs.write("passkey_child", "...passkey child contents")
    run_su = ... (we probably do not need to restart sssd anymore)
    playback_umockdev = ...
    run expect ...

    # Now this is probably not needed in passkey case, but if we want to restore the files immediately so original files are used for the rest of the test
    self.fs.restore("passkey_child")
```

ikerexxe:
@pbrezina the restore function doesn't work, and I think that's the issue Madhuri is facing.

I created a simple test case to test this functionality:
```python
def test_files_provider__example(client: Client, provider: GenericProvider):
    client.fs.copy("/usr/libexec/sssd/passkey_child", "/usr/libexec/sssd/passkey_child.orig")
    client.host.ssh.exec(["rm", "/usr/libexec/sssd/passkey_child"])
    client.fs.restore("/usr/libexec/sssd/passkey_child")
    pass
```

The copy works correctly, and I can find a file called `/usr/libexec/sssd/passkey_child.orig` with the content of `/usr/libexec/sssd/passkey_child`. But the restoration doesn't work and `/usr/libexec/sssd/passkey_child` never comes back. I reviewed the code and I thought that restoring `/usr/libexec/sssd/passkey_child.orig` would help, but that doesn't work either. Can you take a look to see what's happening?

pbrezina:
> @pbrezina the restore function doesn't work, and I think that's the issue Madhuri is facing.
> 
> I created a simple test case to test this functionality:
> 
> ```python
> @pytest.mark.topology(KnownTopology.LDAP)
> def test_files_provider__example(client: Client, provider: GenericProvider):
>     client.fs.copy("/usr/libexec/sssd/passkey_child", "/usr/libexec/sssd/passkey_child.orig")
>     client.host.ssh.exec(["rm", "/usr/libexec/sssd/passkey_child"])
>     client.fs.restore("/usr/libexec/sssd/passkey_child")
>     pass
> ```
> 
> The copy works correctly, and I can find a file called `/usr/libexec/sssd/passkey_child.orig` with the content of `/usr/libexec/sssd/passkey_child`. But the restoration doesn't work and `/usr/libexec/sssd/passkey_child` never comes back. I reviewed the code and I thought that restoring `/usr/libexec/sssd/passkey_child.orig` would help, but that doesn't work either. Can you take a look to see what's happening?

You did not backup `/usr/libexec/sssd/passkey_child`


 ```diff
   @pytest.mark.topology(KnownTopology.LDAP)
   def test_files_provider__example(client: Client, provider: GenericProvider):
+     client.fs.backup("/usr/libexec/sssd/passkey_child")
       client.fs.copy("/usr/libexec/sssd/passkey_child", "/usr/libexec/sssd/passkey_child.orig")
       client.host.ssh.exec(["rm", "/usr/libexec/sssd/passkey_child"])
       client.fs.restore("/usr/libexec/sssd/passkey_child")
       pass
 ```

`copy` only creates backup of the destination file because the source file is not touched

sssd_test_framework/utils/authentication.py
madhuriupadhye: Updated the code.

sssd_test_framework/hosts/client.py
madhuriupadhye: now, we are not updating the client.py.

madhuriupadhye:
> In the example from [#55 (comment)](https://github.com/SSSD/sssd-test-framework/pull/55#issuecomment-1792663813) `write` creates backup of `/usr/libexec/sssd/passkey_child`. In your code, there is no `write` so you have to call `backup`.

After the backup, it worked as expected. Thank you!

sssd_test_framework/utils/authentication.py
sumit-bose: Hi,

thanks, I'm fine with the latest version and will mark this conversation as resolved.

bye,
Sumit

sumit-bose (approved):

sssd_test_framework/utils/authentication.py
```diff
@@ -320,22 +320,30 @@ def passkey(self, username: str, *, pin: str | int, device: str, ioctl: str, scr
         :return: True if authentication was successful, False otherwise.
         :rtype: bool
         """
-        self.fs.backup("/etc/sysconfig/sssd")
+        self.fs.backup("/usr/libexec/sssd/passkey_child")
+        self.fs.copy("/usr/libexec/sssd/passkey_child", "/usr/libexec/sssd/passkey_child.orig")
+        self.host.ssh.run("rm -f /usr/libexec/sssd/passkey_child", raise_on_error=False)
```
pbrezina, line 325: You can remove this, it is not needed.

sssd_test_framework/utils/authentication.py
```diff
@@ -320,22 +320,30 @@ def passkey(self, username: str, *, pin: str | int, device: str, ioctl: str, scr
...
-            """,
+                #!/bin/bash
+                set -ex
+                echo '#!/bin/bash' > /usr/libexec/sssd/passkey_child
+                echo -n 'export ' >> /usr/libexec/sssd/passkey_child
+                env | grep ^UMOCKDEV_ >> /usr/libexec/sssd/passkey_child
+                echo -n 'export ' >> /usr/libexec/sssd/passkey_child
+                printf "LD_PRELOAD=$LD_PRELOAD\n" >> /usr/libexec/sssd/passkey_child
+                echo 'exec /usr/libexec/sssd/passkey_child.orig $@' >> /usr/libexec/sssd/passkey_child
+                chmod 755 /usr/libexec/sssd/passkey_child
```
pbrezina, line 341: You can move this in `self.fs.write(...)` and remove `self.fs.backup("/usr/libexec/sssd/passkey_child")`

sssd_test_framework/utils/authentication.py
```diff
@@ -320,22 +320,30 @@ def passkey(self, username: str, *, pin: str | int, device: str, ioctl: str, scr
...
+                systemctl restart sssd
```
pbrezina, line 342: Restarting sssd should not be needed anymore since passkey_child is not running at this moment.

sssd_test_framework/utils/authentication.py
```diff
@@ -320,22 +320,30 @@ def passkey(self, username: str, *, pin: str | int, device: str, ioctl: str, scr
...
+                chmod -R a+rwx $UMOCKDEV_DIR
```
pbrezina, line 343: What's in UMOCKDEV_DIR? Is this really needed?

pbrezina (changes requested):

sssd_test_framework/utils/authentication.py
spoore1: UMOCKDEV_DIR is a temporary working directory where it does some of it's emulation.   It was needed with the original solution with su to make it work as expected so I expect it is needed here as well.

sssd_test_framework/utils/authentication.py
spoore1: @pbrezina we might be able to predict what some of the UMOCKDEV_* vars will be based on the device info but, I'm not sure we can predict UMOCKDEV_DIR.   That will differ everytime umockdev-run is executed.

Without creating the passkey_child every time with the var for that execution like above, how can we make sure the passkey_child contains the correct value for UMOCKDEV_DIR?    I'm guessing since it's SSSD that executes passkey_child, that it inherits it's env from there right?

sssd_test_framework/utils/authentication.py
ikerexxe: The UMOCKDEV_DIR name is generated randomly every time umockdev is run. Well, maybe there's some logic behind it but to me it looks completely random.

sssd_test_framework/utils/authentication.py
madhuriupadhye: content_run_su = """
                #!/bin/bash
                export LD_PRELOAD=/opt/random.so:libumockdev-preload.so.0
                exec /usr/libexec/sssd/passkey_child.orig $@ \n
                """
        self.fs.write("/usr/libexec/sssd/passkey_child", content_run_su, mode="0755")

        run_su = self.fs.mktmp(
            rf"""
                #!/bin/bash
                set -ex
                echo "" >> /usr/libexec/sssd/passkey_child
                echo -n 'export ' >> /usr/libexec/sssd/passkey_child
                env | grep ^UMOCKDEV_ >> /usr/libexec/sssd/passkey_child
                #systemctl restart sssd
                chmod -R a+rwx $UMOCKDEV_DIR
                su --shell /bin/sh nobody -c "su - '{username}'"
                """,
            mode="a=rx",
        )

hello @pbrezina, I try this way, but its not working, as half part with `fs.write` and other with `self.fs.mktmp` getting auth failure.

pbrezina: hmm, ok. Then disregard this comment.

sssd_test_framework/utils/authentication.py
madhuriupadhye: yes, we need it and every time it create the random directory under /tmp that starts with` umockdev` prefix
like
export UMOCKDEV_DIR=/tmp/umockdev.8LORD2
export UMOCKDEV_DIR=/tmp/umockdev.DVC2D2

pbrezina:
Please, let me know when it is ready for final review.

sssd_test_framework/utils/authentication.py
madhuriupadhye: updated.

sssd_test_framework/utils/authentication.py
madhuriupadhye: yes, removed.

madhuriupadhye:
> Please, let me know when it is ready for final review.

Yes, Done with my changes, please check again.

pbrezina (commented):
Ack, just please change the commit message, e.g. `passkey: avoid infinite loop in cares`

spoore1 (dismissed):
LGTM but, I think we need a new version of pytest-mh built and the requirements updated to fix the tox error:  `LinuxFileSystem" has no attribute "copy"  [attr-defined`

pbrezina:
> LGTM but, I think we need a new version of pytest-mh built and the requirements updated to fix the tox error: `LinuxFileSystem" has no attribute "copy" [attr-defined`

Yes. @madhuriupadhye can you add https://github.com/SSSD/sssd-test-framework/commit/18b47de186164d7c70a7f8807ff0b86e71e565e1 but bump the version to 1.0.7 please?

pbrezina (approved):

PR #56, opened 2023-10-27
LDAP: inetOrgPerson scheme added

patriki01:
I added only attributes that I needed i.e. ``mail`` and ``sn`` (which is required by inetOrgPerson). I can add more attributes if necessary.


justin-stephenson (approved):
Ack, CI failure looks unrelated.

pbrezina (approved):

PR #58, opened 2023-11-06
Tools: wait_for() added

patriki01:
Function for waiting for bash commands. If something is not clear, please suggest changes.

### DIFF
--- a/sssd_test_framework/utils/tools.py
+++ b/sssd_test_framework/utils/tools.py
@@ -586,6 +586,41 @@ class LinuxToolsUtils(MultihostUtility[MultihostHost]):
 
         super().teardown()
 
+    def wait_for_condition(self, condition: str, body: str = "", timeout: int = 60) -> SSHProcessResult:
+        """
+        Wait at maximum ``timeout`` seconds until the ``condition`` is true. Execute ``body`` after each attempt.
+        The condition is a bash expression, usually it is a single bash command that must succeed before a test
+        can continue.
+
+        .. note::
+
+            Internally, this expands to ``timeout {time}s bash -c 'until {condition}; do : {body}; done'``.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.LDAP)
+            def test_just_condition(client: Client):
+                client.sssd.domain["ldap_uri"] = "ldap://typo"
+                client.sssd.start(debug_level=None, raise_on_error=False)
+
+                assert client.sssd.default_domain
+                r = client.tools.wait_for_condition(condition=f"sssctl domain-status {client.sssd.default_domain}")
+                assert r.rc == 0
+                assert "LDAP: not connected" in r.stdout
+
+        :param condition: Command that is awaited
+        :type condition: str
+        :param body: Body to be executed while waiting for condition, defaults to ""
+        :type body: str, optional
+        :param timeout: How long should we try the command in seconds, defaults to 60
+        :type timeout: int, optional
+        :return: Proccess result
+        :rtype: SSHProcessResult
+        """
+
+        return self.host.ssh.run(f"timeout {timeout}s bash -c 'until {condition}; do : {body}; done'")
+
 
 class KillCommand(object):
     def __init__(self, host: MultihostHost, process: SSHProcess, pid: int) -> None:


justin-stephenson:
Hi Patrik, Thanks for the PR. Can you elaborate on the use-case here? I understand the `timeout` part for a command that does not return immediately, but I am not quite understanding the need for the bash `until` loop.

patriki01:
The main purpose of this function is not long running commands, although it can be used for them. **This function is intended to replace fixed sleep time in tests, which almost always have some arbitrary number inside.**
See the example below. Why is there a fixed sleep time? Probably because otherwise the call under the sleep command would fail. If we use this construct, it will attempt to run the command for 60 seconds, and if the command fails, sleep(1) (by default) and then try to run the command again until it completes successfully. In this example, it should definitely be under 15 seconds. As a result, we have reduced the sleep time, saving some resources.
On top of that, different distributions might need different sleep time for particular command, which is solved by this constructor as well.
I hope this helps.
```
def test_refresh_after_cleanup_task(ldap_conn, refresh_after_cleanup_task):
    ent.assert_group_by_name(
        "group2",
        dict(mem=ent.contains_only("user1")))

    ent.assert_passwd_by_name(
        'user1',
        dict(name='user1', passwd='*', uid=1001, gid=2001,
             gecos='1001', shell='/bin/bash'))

    time.sleep(15)

    ent.assert_group_by_name(
        "group2",
        dict(mem=ent.contains_only("user1")))
```

sssd_test_framework/utils/tools.py
```diff
@@ -586,6 +586,35 @@ def teardown(self):
 
         super().teardown()
 
+    def wait_for(self, condition: str, body: str = "sleep 1", time: int = 60) -> SSHProcessResult:
```
pbrezina, line 589: I wouldn't use `sleep 1` as default, empty string will be better. Our use case is mostly to "wait until command succeeds" to avoid potential race condition, but usually it should succeed very quickly. By putting sleep there, we would just delay the test. 

We can even omit the sleep completely, or use something like 5us (`sleep 0.000005`). Omitting it might be better.

pbrezina: Rename `time` -> `timeout`

sssd_test_framework/utils/tools.py
```diff
@@ -586,6 +586,35 @@ def teardown(self):
...
+        """
+        For a period of ``time`` seconds try the ``condition`` command to finish successfully.
+        After each attempt process sleeps for 1 second by default.
```
pbrezina, line 592:
```
Wait at maximum ``timeout`` seconds until the ``condition`` is true. Execute ``body`` after each attempt. The condition is a bash expression, usually it is a single bash command that must succeed before a test can continue.

.. note::
    Internally, this expands to ``timeout {time}s bash -c 'until {condition}; do : {body}; done'``.
```

pbrezina (changes requested):

sssd_test_framework/utils/tools.py
patriki01: Fixed

patriki01: The way I see this function is that it just waits for the command until it is executed successfully. But we have to stop the command if it takes just to long to succeed. Basically, the main purpose is to wait for success not to wait for timeout. But of course we can change the name.

I think the optional argument ``body`` could come in handy and it doesn't matter if we leave it there.

sssd_test_framework/utils/tools.py
```diff
@@ -586,6 +586,40 @@ def teardown(self):
...
         super().teardown()
 
+    def wait_for(self, condition: str, body: str = "", timeout: int = 60) -> SSHProcessResult:
+        """
+        Wait at maximum ``timeout`` seconds until the ``condition`` is true. Execute ``body`` after each attempt.
+        The condition is a bash expression, usually it is a single bash command that must succeed before a test
+        can continue.
+
+        .. note::
+            Internally, this expands to ``timeout {time}s bash -c 'until {condition}; do : {body}; done'``.
```
pbrezina, line 596: You need to add a blank line before the paragraph.

sssd_test_framework/utils/tools.py
```diff
@@ -586,6 +586,40 @@ def teardown(self):
...
+            def test_just_condition(client: Client):
+                client.sssd.domain["ldap_uri"] = "ldap://typo"
+                client.sssd.start(debug_level=None, raise_on_error=False)
+
+                assert client.sssd.default_domain
+                r = client.tools.wait_for(condition=f"sssctl domain-status {client.sssd.default_domain}")
+                assert r.rc == 0
+                assert "LDAP: not connected" in r.stdout
+
+        :param condition: command that is awaited
```
pbrezina, line 611: Please, use uppercase first letters.

pbrezina (changes requested):

sssd_test_framework/utils/tools.py
pbrezina: Body can be helpful at some cases, for example if you known that if the command failed, you want to wait some time before trying again because it does not make sense sooner. But we will not use it much probably, since our main use case is to avoid race conditions.

`loop_until_timeout` suggests that we really wait for a timeout to happen, which is not the case. Maybe we can just use `until` which is the bash command? Or `wait_for_condition`, `wait_for_success`?

sssd_test_framework/utils/tools.py
patriki01: Okay and why? Docs seems to look okay

sssd_test_framework/utils/tools.py
justin-stephenson: > Body can be helpful at some cases, for example if you known that if the command failed, you want to wait some time before trying again because it does not make sense sooner. But we will not use it much probably, since our main use case is to avoid race conditions.

Okay sure, either way I don't insist.
> 
> `loop_until_timeout` suggests that we really wait for a timeout to happen, which is not the case. Maybe we can just use `until` which is the bash command? Or `wait_for_condition`, `wait_for_success`?

I'm good with `wait_for_condition` or `wait_for_success`

patriki01: Okay, renamed

sssd_test_framework/utils/tools.py
pbrezina: That is a surprise to me :-) RST syntax is:

```
.. directive:: arguments
    options

    content
```

pbrezina (changes requested):
Thank you, ack. Just please, amend the commit message with the new name, i.e. `tools: add wait_for_condition`

justin-stephenson (approved):
Ack, thank you.

patriki01:
Commit renamed. Thanks for cooperation

PR #59, opened 2023-11-13
sssctl: sssctl analyze command added

patriki01:
Ask, if something is not clear.

### DIFF
--- a/sssd_test_framework/utils/sssctl.py
+++ b/sssd_test_framework/utils/sssctl.py
@@ -263,3 +263,25 @@ class SSSCTLUtils(MultihostUtility[MultihostHost]):
         }
 
         return self.host.ssh.exec(["sssctl", "domain-status"] + self.cli.args(args), raise_on_error=False)
+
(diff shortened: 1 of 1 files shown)


sssd_test_framework/utils/sssctl.py
```diff
@@ -263,3 +263,26 @@ def domain_status(
         }
 
         return self.host.ssh.exec(["sssctl", "domain-status"] + self.cli.args(args), raise_on_error=False)
+
+    def analyze(
```
justin-stephenson, line 267: could be better to call this `analyze_request` and always execute request subcommands? We have an analyze `error` module (https://github.com/SSSD/sssd/commit/ed3726c37fe07aab788404bfa2f9003db15f4210) but no need to test it at the moment

patriki01: Good point. Fixed

justin-stephenson (approved):
Ack thank you.

thalman (approved):
ACK

PR #61, opened 2023-11-22
CLIENT: add detection of "--with-sssd-user" feature

alexey-tikhonov:


sssd_test_framework/hosts/client.py
```diff
@@ -43,6 +43,7 @@ def features(self) -> dict[str, bool]:
 
             [ -f "/usr/lib64/sssd/libsss_files.so" ] && echo "files-provider" || :
             [ -f "/usr/libexec/sssd/passkey_child" ] && echo "passkey" || :
+            man sssd.conf | grep -q "The user to drop the privileges to" && echo "non-privileged" :
```
ikerexxe, line 46: Add `||` at the end, like other features. If you don't do it and a test case checks whether this functionality is provided, then the test environment will fail miserably.

alexey-tikhonov: Right, it would set feature value to `non-privileged :`
Fixed.

alexey-tikhonov:
Added a second patch to makes use of new feature.
Initially I tried to add a method to `SSSDUtils` class but looks like there is no sane way to access client features from there.

alexey-tikhonov:
```
INFO     2023-11-25 15:14:25,803      client.test Detected features:
                                                  Features: 
                                                    files-provider: False
                                                    passkey: False
                                                    non-privileged: False
```

alexey-tikhonov:
FWIW, https://github.com/SSSD/sssd/pull/7052 is an usage example.
On centos-8:
```
tests/test_id.py::test_id__getpwnam[root] (ipa) PASSED                   [ 16%]
tests/test_id.py::test_id__getpwnam[root] (ldap) PASSED                  [ 16%]
tests/test_id.py::test_id__getpwnam[root] (samba) PASSED                 [ 17%]
tests/test_id.py::test_id__getpwnam[sssd] (ipa) PASSED                   [ 17%]
tests/test_id.py::test_id__getpwnam[sssd] (ldap) PASSED                  [ 17%]
tests/test_id.py::test_id__getpwnam[sssd] (samba) PASSED                 [ 18%]
```
but due to above I'm not sure if it really runs under 'sssd' or SSSD was actually built without-sssd-user and sssd.conf::user option is simply ignored? Since there are no fails, SSSD logs were not captured.

ikerexxe:
IIUC, the second commit configures SSSD to run under `sssd` user by default when it's available. Am I right? The commit message isn't very clear, so maybe it's worth rewriting it.

By the way, you wrote `Cleint` instead of `Client` in the commit message.

alexey-tikhonov:
> IIUC, the second commit configures SSSD to run under `sssd` user by default when it's available. Am I right? The commit message isn't very clear, so maybe it's worth rewriting it.

Not "by default", but when `set_sssd_user("sssd")` is called explicitly.
It just refuses to set "sssd.conf::user" option if feature isn't supported.


> By the way, you wrote `Cleint` instead of `Client` in the commit message.

Fixed.

alexey-tikhonov:
And btw I got logs from https://github.com/SSSD/sssd/pull/7052 and it works as expected - 'test_id__getpwnam[sssd] (ldap)':
```
(2023-11-27 10:42:33): [sssd] [get_provider_config] (0x0100): Formed command '/usr/libexec/sssd/sssd_be --domain test --uid 996 --gid 990 --logger=files' for provider '%BE_test'
...
(2023-11-27 10:42:33): [be[test]] [become_user] (0x0200): Trying to become user [996][990].
```

alexey-tikhonov:
What I eventually would like  - to be able to do things like:
```
@pytest.mark.parametrize("sssd_service_user", ("root", "sssd"))
```
Ideally it should be a dynamic list, based on feature availability, but I didn't figure how to check client's properties in the fixture.

sssd_test_framework/roles/client.py
```diff
@@ -94,3 +94,18 @@ def setup(self) -> None:
                     raise ValueError("List is not expected")
 
                 self.sssd.import_domain(domain, role)
+
+    def set_sssd_user(self, user: str) -> bool:
```
pbrezina, line 98: Can you move it to SSSDUtils? This way, it will be also usable on the IPA server, not just the client.

sssd_test_framework/roles/client.py
```diff
@@ -94,3 +94,18 @@ def setup(self) -> None:
...
+        """
+        Set [sssd]/user option.
+
+        :param user: Option value to set.
+        :type user: str
+        :return: True if success, false otherwise
+        :rtype: bool
```
pbrezina, line 105: Wouldn't it make more sense to raise an error instead of returning bool?

pbrezina:
> Added a second patch that makes use of new feature. Initially I tried to add a method to `SSSDUtils` class but looks like there is no sane way to access client features from there.

You can use `self.host.features` since client.features is the same client.host.features. In the long run, I think the MultihostUtility should the MultihostRole instead of MultihostHost as an argument.

sssd_test_framework/roles/client.py
alexey-tikhonov: Done.

alexey-tikhonov:
> > Ideally it should be a dynamic list, based on feature availability, but I didn't figure how to check client's properties in the fixture.
> 
> Hi, it is possible.
> 
> Essentially, pytest.mark.builtwith just creates callback that lookups the parameters in `client.features` and then add pytest.mark.require(callback). What you want to do, is use pytest.mark.require directly, with custom callback,

Done.

I think I addressed all comments and also updated https://github.com/SSSD/sssd/pull/7052

ikerexxe (approved):
LGTM!

pbrezina (approved):
Ack.

@alexey-tikhonov Before merging, we discussed the possibility of adding the parametrization into PR CI workflow instead of into the framework so all tests are run both with root and non-root. What's your plan here?

alexey-tikhonov:
> Before we merge, I'd like to clarify: We discussed the possibility of adding the parametrization into PR CI workflow instead of into the framework so all tests are run both with root and non-root. What's your plan here?

 - currently some tests ('passkey', for example) do not work when SSSD runs under non-privileged user (issue with the test, not with SSSD)
 - this would double time PR CI spends in 'system' tests, I'm not sure it's worth to run **all** the tests in both modes
 - this would require to keep "knowledge" of what platforms build with "--with-sssd-user=sssd" in additional place
 - there are known issues yet to address in SSSD code (even if tests currently doesn't cover those cases)
 - ...

For a start I'd like to select very basic set of core tests (NSS and auth) and run those in both modes where supported (practically only C9S currently). I think it's a good start to have a kind of regression testing for the feature.

Maybe eventually we will build "--with-sssd-user=sssd" everywhere and will change defaults to "user=sssd" (ideally directly in systemd configs). Then we can remove this helpers as "not needed anymore".

pbrezina:
Thank you. Merging...

PR #64, opened 2023-12-07
ad: avoid error if backup file does not exist

pbrezina:
Currently, there is an error in the command output.


thalman (approved):
LGTM, ACK

PR #68, opened 2024-01-11
ipa: fix get methods

pbrezina:
Recent changes to attrs_parse broke ipa.get since the output of ipa
*-show command prepends the attributes with single space.

### DIFF
--- a/sssd_test_framework/roles/ipa.py
+++ b/sssd_test_framework/roles/ipa.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from textwrap import dedent
 from typing import Any
 
 from pytest_mh.cli import CLIBuilderArgs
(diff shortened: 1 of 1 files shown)


spoore1:
This fix resolves the IPA user data issue I was seeing.   Thank you.

The previous change appears to have broken AD user get as well except differently.   The command returns output formatted with spaces between the key names and the colons:

```
> /usr/local/lib/python3.11/site-packages/sssd_test_framework/roles/ad.py(508)get()
-> return attrs_parse(cmd.stdout_lines, attrs)
(Pdb) p cmd.stdout_lines
['', '', 'DistinguishedName : CN=kcuser111551,CN=Users,DC=da,DC=test', 'Enabled           : True', 'GivenName         : dummyfirstname', 'Name              : kcuser111551', 'ObjectClass       : user', 'ObjectGUID        : 6e27b8c8-77fc-447c-a19d-49b9aec9503c', 'SamAccountName    : kcuser111551', 'SID               : S-1-5-21-3858741270-958828843-232487239-1114', 'Surname           : dummylastname', 'UserPrincipalName : ', '', '', '']
```

I believe this can be fixed similar to the IPA fix with something like this in the ad.py role module:

```
        lines = re.sub('[ ]*:', ':', cmd.stdout).splitlines()
        return attrs_parse(lines, attrs)
```

Can you add something like this here as well?

spoore1 (approved):
LGTM.   My tests work again with this change.

spoore1:
FYI, I opened a separate PR as suggested for the AD data here:

https://github.com/SSSD/sssd-test-framework/pull/70

PR #71, opened 2024-01-19
utils: added sshd config and service utilities

danlavu:


sssd_test_framework/utils/sshd.py
```diff
@@ -0,0 +1,126 @@
...
+            # Select authselect profile and feature
+            client.authselect.select("sssd", ["with-gssapi", "with-mkhomedir"])
+
+            # Start sssd
+            client.sssd.start()
+
+            # Configure ssh client and daemon
+            client.sshd.config_set([{"GSSAPIAuthentication": "yes"}])
+
+            # Start sshd
```
ikerexxe, line 36: I think this should be `Restart sshd`

ikerexxe (commented):
Just a minor comment, everything else looks good.

sssd_test_framework/utils/sshd.py
danlavu: Yes, yes, it should.

ikerexxe (dismissed):
LGTM! Thanks for the patch

sssd_test_framework/utils/sshd.py
```diff
@@ -0,0 +1,126 @@
...
+        self.args: str = f'--transform "sshd.lns incl {self.file}"'
+        self.cmd: str = ""
+
+    def setup_when_used(self) -> None:
+        super().setup_when_used()
+        self.fs.backup(self.file)
+
+    def config_read(self) -> str:
+        """
+        Read SSH daemon configuration as Augeas tree.
```
pbrezina, line 69: Add blank line. Same on other places.

pbrezina (changes requested):
Hi Dan, please, see minor change in the doc string. Otherwise LGTM. I don't plan to test it, I trust that the code works for you.

danlavu:
It works for me. Moving to approved.

pbrezina:
@danlavu Remove the merge commit please.

danlavu:
> @danlavu Remove the merge commit please.

Done.

sssd_test_framework/utils/sshd.py
```diff
@@ -0,0 +1,129 @@
...
+        Reload the SSH daemon.
+
+        :param service: Service to start, defaults to 'sshd'
+        :type service: str, optional
+        :param raise_on_error: Raise exception on error, defaults to True
+        :type raise_on_error: bool, optional
+        :return: SSH process result.
+        :rtype: SSHProcessResult
+        """
+
```
pbrezina, line 128: This blank line should not be present.

sssd_test_framework/utils/sshd.py
```diff
@@ -0,0 +1,128 @@
...
+        self.svc: SystemdServices = svc
+
+        self.file: str = file
+        self.path: str = "/files" + self.file
+        self.args: str = f'--transform "sshd.lns incl {self.file}"'
+        self.cmd: str = ""
+
+    def setup_when_used(self) -> None:
+        super().setup_when_used()
+        self.fs.backup(self.file)
```
pbrezina, line 65: I was about to push it, but this also requires a teardown to reload original configuration. Something like:

```
def teardown_when_used(self) -> None:
    self.fs.restore(self.file)
    self.svc.reload("sshd")
```

pbrezina (changes requested):
Also how is this going to be used? Wouldn't it be maybe better to create new instance of sshd listening on different port so you don't meddle with ssh connections that we use to talk to the host?

sssd_test_framework/utils/sshd.py
danlavu: I thought pytest-mh , when using self.fs.backup restored the file. This needs to be updated for the Pam util class I wrote to then.

danlavu: Moving to draft, something isn't right.

danlavu: Okay, self.fs.restore(self.file) isn't necessary, the file does get restored.

danlavu: https://pastebin.com/TiNs0Q3S

I put a print after the config_set and another in the teardown, the changes have already been reverted.

danlavu: I don't for see many tests using this function at all, this is mainly to enable GSSAPIAuthentication on the host which can be done safely with little impact. I updated the comment with a warning when using this. There're only a few tests that need to edit the sshd config.

pbrezina: Yes, at least for the moment.



> Okay, self.fs.restore(self.file) isn't necessary, the file does get restored.

This is something we need to work on in pytest-mh. To build a dependency tree so it is clear in which order are the MultihostUtility objects destroyed. But it works for us at the moment.

pbrezina (changes requested):

PR #72, opened 2024-01-21
utils: added password_with_output function to capture ssh output

danlavu:
some test cases will need to check the ssh output

### DIFF
--- a/sssd_test_framework/utils/authentication.py
+++ b/sssd_test_framework/utils/authentication.py
@@ -413,6 +413,80 @@ class SSHAuthenticationUtils(MultihostUtility[MultihostHost]):
         self.opts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
         """SSH CLI options."""
 
+    def password_with_output(self, username: str, password: str) -> tuple[int, int, str, str]:
+        """
+        SSH to the remote host and authenticate the user with password and captures standard output and error.
+
+        :param username: Username.
+        :type username: str
+        :param password: User password.
+        :type password: str
+        :return: Tuple containing [except return code, command exit code, stdout, stderr].
+        :rtype: Tuple[int, int, str, str]
+        """
+
+        result = self.host.ssh.expect_nobody(
+            rf"""
+            # Disable debug output
+            exp_internal 0
+
+            proc exitmsg {{ msg code }} {{
+                # Send EOF, if we are in the prompt
+                send \x04;
+
+                # Wait for the exit code
+                lassign [wait] pid spawnid os_error_flag rc
+
+                puts ""
+                puts "expect result: $msg"
+                puts "expect exit code: $code"
+                puts "expect spawn exit code: $rc"
+                exit $code
+            }}
+
+            # It takes some time to get authentication failure
+            set timeout {DEFAULT_AUTHENTICATION_TIMEOUT}
+            set prompt "\n.*\[#\$>\] $"
+            log_user 1
+            log_file /tmp/expect.log
+
+            spawn ssh {self.opts} \
+                -o PreferredAuthentications=password \
+                -o NumberOfPasswordPrompts=1 \
+                -l "{username}" localhost
+
+            expect {{
+                "password:" {{send "{password}\n"}}
+                timeout {{exitmsg "Unexpected output" 201}}
+                eof {{exitmsg "Unexpected end of file" 202}}
+            }}
+
+            expect {{
+                -re $prompt {{exitmsg "Password authentication successful" 0}}
+                "{username}@localhost: Permission denied" {{exitmsg "Authentication failure" 1}}
+                "Connection closed by UNKNOWN port 65535" {{exitmsg "Connection closed" 2}}
+                timeout {{exitmsg "Unexpected output" 201}}
+                eof {{exitmsg "Unexpected end of file" 202}}
+            }}
+
+            exitmsg "Unexpected code path" 203
+            """,
+            verbose=False,
+        )
+
+        if result.rc > 200:
+            raise ExpectScriptError(result.rc)
+
+        expect_data = result.stdout_lines[-3:]
+
+        # Get command exit code.
+        cmdrc = int(expect_data[2].split(":")[1].strip())
+
+        # Alter stdout, first line is spawned command, the last three are our expect output.
+        stdout = "\n".join(result.stdout_lines[1:-3])
+
+        return result.rc, cmdrc, stdout, result.stderr
+
     def password(self, username: str, password: str) -> bool:
         """
         SSH to the remote host and authenticate the user with password.
@@ -424,41 +498,8 @@ class SSHAuthenticationUtils(MultihostUtility[MultihostHost]):
         :return: True if authentication was successful, False otherwise.
         :rtype: bool
         """
-
-        result = self.host.ssh.expect_nobody(
-            rf"""
-            # It takes some time to get authentication failure
-            set timeout {DEFAULT_AUTHENTICATION_TIMEOUT}
-            set prompt "\n.*\[#\$>\] $"
-
-            spawn ssh {self.opts} \
-                -o PreferredAuthentications=password \
-                -o NumberOfPasswordPrompts=1 \
-                -l "{username}" localhost
-
-            expect {{
-                "password:" {{send "{password}\n"}}
-                timeout {{puts "expect result: Unexpected output"; exit 201}}
-                eof {{puts "expect result: Unexpected end of file"; exit 202}}
-            }}
-
-            expect {{
-                -re $prompt {{puts "expect result: Password authentication successful"; exit 0}}
-                "{username}@localhost: Permission denied" {{puts "expect result: Authentication failure"; exit 1}}
-                "Connection closed by UNKNOWN port 65535" {{puts "expect result: Connection closed"; exit 2}}
-                timeout {{puts "expect result: Unexpected output"; exit 201}}
-                eof {{puts "expect result: Unexpected end of file"; exit 202}}
-            }}
-
-            puts "expect result: Unexpected code path"
-            exit 203
-        """
-        )
-
-        if result.rc > 200:
-            raise ExpectScriptError(result.rc)
-
-        return result.rc == 0
+        rc, _, _, _ = self.password_with_output(username, password)
+        return rc == 0
 
     def password_expired(self, username: str, password: str, new_password: str) -> bool:
         """


danlavu:
I don't know how I feel about this, yet. Using expect to issue a double SSH connection. This is what the test is looking like. 

```
@pytest.mark.importance("critical")
@pytest.mark.ticket(bz=2077893)
@pytest.mark.topology(KnownTopology.AD)
def test_sssd__enable_and_disable_with_gssapi(client: Client, provider: GenericProvider):
    """
    :title: Authselect sssd profile with-gssapi
    :setup:
        1. Create provider user "user-1"
        2. Select SSSD profile with-gssapi and with-mkhomedir
        3. Start SSSD
        4. Configure sshd and allow gssapi authentication
        5. Reload sshd
    :steps:
        1. SSH as "user-1" using gssapi
        2. Disable with-gssapi feature and SSH as "user-1" using gssapi
    :expectedresults:
        2. Authentication is successful for "user-1"
        2. Feature is disabled and authentication is unsuccessful for "user-1"
    :customerscenario: False
    :requirement: IDM-SSSD-TC: sssd: bz2077893 add with-gssapi (pam_sss_gss.so)
    """
    provider.user("user-1").add()
    client.authselect.select("sssd", ["with-gssapi", "with-mkhomedir"])
    client.sssd.start()

    client.sshd.config_set([
            {"GSSAPIAuthentication": "yes"},
            {"PasswordAuthentication": "yes"},
            {"UsePAM": "yes"}
    ])
    client.sshd.reload()

    assert client.auth.ssh.gssapi("user-1", password="Secret123")
```

danlavu:
I don't like it, taking it out of "ready for review"

danlavu:
I removed the gssapi stuff for now and just made this PR about capturing the ssh output for some test cases.

danlavu:
`print(client.auth.ssh.password("user-1", password="GOODPASS", output=True))`

`spawn ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PreferredAuthentications=password -o NumberOfPasswordPrompts=1 -l user-1 localhost
Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
user-1@localhost's password: 
Last failed login: Wed Jan 31 04:08:06 UTC 2024 from 172.16.100.1 on ssh:notty
There were 4 failed login attempts since the last successful login.
Could not chdir to home directory /home/test/user-1: Permission denied
-sh: /home/test/user-1/.profile: Permission denied
-sh-5.2$ expect result: Password authentication successful`

`print(client.auth.ssh.password("user-1", password="GOODPASS", output=False)) : TRUE`

`print(client.auth.ssh.password("user-1", password="GOODPASS")) : TRUE`

`print(client.auth.ssh.password("user-1", password="BADPASS")) : FALSE`

pbrezina:
Is this ready for review?

danlavu:
> Is this ready for review?

Yup.

sssd_test_framework/utils/authentication.py
```diff
@@ -413,23 +412,29 @@ def __init__(self, host: MultihostHost) -> None:
...
-        :type name: str
+        :param username: Username.
+        :type username: str
         :param password: User password.
         :type password: str
-        :return: True if authentication was successful, False otherwise.
-        :rtype: bool
+        :param output: Log output.
+        :type output: bool
+        :return: True if authentication was successful, False otherwise. If output is true, return output.
```
ikerexxe, line 425: Wouldn't it be better to return the two values (auth success/failure and output)? Is that possible in python?

sssd_test_framework/utils/authentication.py
```diff
@@ -413,23 +412,29 @@ def __init__(self, host: MultihostHost) -> None:
...
+        :rtype: bool | str
         """
+        log = "0"
+        if output is True:
+            log = "1"
```
ikerexxe, line 430: I don't understand why you need to convert the boolean to a numeric value. Can you explain?

danlavu: I think its nicer to use True/False instead if "1" | "0".

sssd_test_framework/utils/authentication.py
danlavu: It's not used very often and what tests will use it isn't concerned about it passing.  Wanted to ensure that no code needed to change from this PR.

sssd_test_framework/utils/authentication.py
```diff
@@ -192,14 +192,13 @@ def password(self, username: str, password: str) -> bool:
         """
         Call ``su - $username`` and authenticate the user with password.
 
-        :param name: User name.
-        :type name: str
+        :param username: Username.
+        :type username: str
```
pbrezina, line 196: Please, put this into separate commit.

sssd_test_framework/utils/authentication.py
pbrezina: Create method `password_with_output`, that returns `tuple[bool, str, str]` (result, stdout, stderr) so you get all information. `password` will call `password_with_output`.

This way, it is clear what is return, you get all information that you might need and it will remain type safe.

pbrezina (changes requested):

sssd_test_framework/utils/authentication.py
```diff
@@ -538,6 +538,54 @@ def password_expired(self, username: str, password: str, new_password: str) -> b
 
         return result.rc == 0
 
+    def password_with_output(self, username: str, password: str) -> tuple[int, str, str]:
```
pbrezina, line 541: This is exactly the same as password, right? If yes, we should use it in the password method:

```python
def password(self, username: str, password: str) -> bool:
    rc, _, _ = self.password_with_output(username, password)
    return rc
```

pbrezina (changes requested):

sssd_test_framework/utils/authentication.py
danlavu: The only difference is setting log_user 1 in expect.

ikerexxe (approved):
This is much better! Thanks for taking into account the feedback.

danlavu:
I had to fix the return type on the password function, @ikerexxe  that has been updated.

ikerexxe (approved):

pbrezina:
Ack. But since we are doing this, I would like to add one more thing to the return value, to avoid changing it later again.

It should return (expect return code, spawned command return code, stdout, stderr).

I also added https://github.com/next-actions/pytest-mh/commit/ee357a95a41f075c5019ec0a8805662092e0f2e2 so we can capture only the ssh output. Please, consider the following change:

```diff
diff --git a/sssd_test_framework/utils/authentication.py b/sssd_test_framework/utils/authentication.py
index 8727791..aaa0b68 100644
--- a/sssd_test_framework/utils/authentication.py
+++ b/sssd_test_framework/utils/authentication.py
@@ -413,7 +413,7 @@ class SSHAuthenticationUtils(MultihostUtility[MultihostHost]):
         self.opts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
         """SSH CLI options."""
 
-    def password_with_output(self, username: str, password: str) -> tuple[int, str, str]:
+    def password_with_output(self, username: str, password: str) -> tuple[int, int, str, str]:
         """
         SSH to the remote host and authenticate the user with password and captures standard output and error.
 
@@ -421,16 +421,34 @@ class SSHAuthenticationUtils(MultihostUtility[MultihostHost]):
         :type username: str
         :param password: User password.
         :type password: str
-        :return: Tuple containing [return code, standard out, standard error].
-        :rtype: Tuple[int, str, str]
+        :return: Tuple containing [except return code, command exit code, stdout, stderr].
+        :rtype: Tuple[int, int, str, str]
         """
 
         result = self.host.ssh.expect_nobody(
             rf"""
+            # Disable debug output
+            exp_internal 0
+
+            proc exitmsg {{ msg code }} {{
+                # Send EOF, if we are in the prompt
+                send \x04;
+
+                # Wait for the exit code
+                lassign [wait] pid spawnid os_error_flag rc
+
+                puts ""
+                puts "expect result: $msg"
+                puts "expect exit code: $code"
+                puts "expect spawn exit code: $rc"
+                exit $code
+            }}
+
             # It takes some time to get authentication failure
             set timeout {DEFAULT_AUTHENTICATION_TIMEOUT}
             set prompt "\n.*\[#\$>\] $"
             log_user 1
+            log_file /tmp/expect.log
 
             spawn ssh {self.opts} \
                 -o PreferredAuthentications=password \
@@ -439,27 +457,35 @@ class SSHAuthenticationUtils(MultihostUtility[MultihostHost]):
 
             expect {{
                 "password:" {{send "{password}\n"}}
-                timeout {{puts "expect result: Unexpected output"; exit 201}}
-                eof {{puts "expect result: Unexpected end of file"; exit 202}}
+                timeout {{exitmsg "Unexpected output" 201}}
+                eof {{exitmsg "Unexpected end of file" 202}}
             }}
 
             expect {{
-                -re $prompt {{puts "expect result: Password authentication successful"; exit 0}}
-                "{username}@localhost: Permission denied" {{puts "expect result: Authentication failure"; exit 1}}
-                "Connection closed by UNKNOWN port 65535" {{puts "expect result: Connection closed"; exit 2}}
-                timeout {{puts "expect result: Unexpected output"; exit 201}}
-                eof {{puts "expect result: Unexpected end of file"; exit 202}}
+                -re $prompt {{exitmsg "Password authentication successful" 0}}
+                "{username}@localhost: Permission denied" {{exitmsg "Authentication failure" 1}}
+                "Connection closed by UNKNOWN port 65535" {{exitmsg "Connection closed" 2}}
+                timeout {{exitmsg "Unexpected output" 201}}
+                eof {{exitmsg "Unexpected end of file" 202}}
             }}
 
-            puts "expect result: Unexpected code path"
-            exit 203
-        """
+            exitmsg "Unexpected code path" 203
+            """,
+            verbose=False,
         )
 
         if result.rc > 200:
             raise ExpectScriptError(result.rc)
 
-        return result.rc, result.stdout, result.stderr
+        expect_data = result.stdout_lines[-3:]
+
+        # Get command exit code.
+        cmdrc = int(expect_data[2].split(":")[1].strip())
+
+        # Alter stdout, first line is spawned command, the last three are our expect output.
+        stdout = "\n".join(result.stdout_lines[1:-3])
+
+        return result.rc, cmdrc, stdout, result.stderr
 
     def password(self, username: str, password: str) -> bool:
         """
@@ -472,7 +498,7 @@ class SSHAuthenticationUtils(MultihostUtility[MultihostHost]):
         :return: True if authentication was successful, False otherwise.
         :rtype: bool
         """
-        rc, _, _ = self.password_with_output(username, password)
+        rc, _, _, _ = self.password_with_output(username, password)
         return rc == 0
 
     def password_expired(self, username: str, password: str, new_password: str) -> bool:
```

danlavu:
@pbrezina  That's fine with me, I applied your patch.

danlavu:
@pbrezina https://github.com/SSSD/sssd-test-framework/pull/85  Same changes to SU

pbrezina (approved):

pbrezina:
@ikerexxe Do you want to look at latest changes?

ikerexxe:
No need. I'm sure they've improved the code.

PR #74, opened 2024-01-31
hosts: adding domain and realm variables

danlavu:
- adding a generic domain and realm variable that can be used in tests, as provider.host.domain/realm
 - updated existing functions to find the domain to use the new variable instead

PR #83, opened 2024-02-15
roles: add domain and realm to providers

pbrezina:
This is a follow up of https://github.com/SSSD/sssd-test-framework/pull/74

@danlavu  It adds domain and realm to role objects as well as a shortcut.


danlavu (approved):
LG2M

PR #85, opened 2024-02-16
utils: adding output capture for su password authentication

danlavu:


danlavu:
Turns out that pam_lastlog only works with su by default, so just like auth.ssh.password, this is for auth.su.password

danlavu:
Redid the pull request without branching it on top of ssh, this can be looked at independently now.

sssd_test_framework/utils/authentication.py
```diff
@@ -188,16 +188,16 @@ def __init__(self, host: MultihostHost, fs: LinuxFileSystem) -> None:
         super().__init__(host)
         self.fs: LinuxFileSystem = fs
 
-    def password(self, username: str, password: str) -> bool:
+    def password_with_output(self, username: str, password: str) -> tuple[int, str, str]:
```
pbrezina, line 191: returns tuple[int, int, str, str] same in docstring

pbrezina: Seems like you are missing the latest changes from ssh password?

pbrezina (changes requested):

sssd_test_framework/utils/authentication.py
danlavu: @pbrezina I tried it with the latest changes, but there was an index error and the code failed to work. SU is less complicated than SSH and I don't think it's necessary to have the command exit codes included in this function. Having just the tuple[int, str, str] I think it's fine?

pbrezina:
Works like a charm:

```diff
diff --git a/sssd_test_framework/utils/authentication.py b/sssd_test_framework/utils/authentication.py
index 22eeed9..04163aa 100644
--- a/sssd_test_framework/utils/authentication.py
+++ b/sssd_test_framework/utils/authentication.py
@@ -188,7 +188,7 @@ class SUAuthenticationUtils(MultihostUtility[MultihostHost]):
         super().__init__(host)
         self.fs: LinuxFileSystem = fs
 
-    def password_with_output(self, username: str, password: str) -> tuple[int, str, str]:
+    def password_with_output(self, username: str, password: str) -> tuple[int, int, str, str]:
         """
         Call ``su - $username`` and authenticate the user with password and captures standard output and error.
 
@@ -202,6 +202,23 @@ class SUAuthenticationUtils(MultihostUtility[MultihostHost]):
 
         result = self.host.ssh.expect_nobody(
             rf"""
+            # Disable debug output
+            # exp_internal 0
+
+            proc exitmsg {{ msg code }} {{
+                # Close spawned program, if we are in the prompt
+                catch close
+
+                # Wait for the exit code
+                lassign [wait] pid spawnid os_error_flag rc
+
+                puts ""
+                puts "expect result: $msg"
+                puts "expect exit code: $code"
+                puts "expect spawn exit code: $rc"
+                exit $code
+            }}
+
             # It takes some time to get authentication failure
             set timeout {DEFAULT_AUTHENTICATION_TIMEOUT}
             set prompt "\n.*\[#\$>\] $"
@@ -210,26 +227,34 @@ class SUAuthenticationUtils(MultihostUtility[MultihostHost]):
 
             expect {{
                 "Password:" {{send "{password}\n"}}
-                timeout {{puts "expect result: Unexpected output"; exit 201}}
-                eof {{puts "expect result: Unexpected end of file"; exit 202}}
+                timeout {{exitmsg "Unexpected output" 201}}
+                eof {{exitmsg "Unexpected end of file" 202}}
             }}
 
             expect {{
-                -re $prompt {{puts "expect result: Password authentication successful"; exit 0}}
-                "Authentication failure" {{puts "expect result: Authentication failure"; exit 1}}
-                timeout {{puts "expect result: Unexpected output"; exit 201}}
-                eof {{puts "expect result: Unexpected end of file"; exit 202}}
+                -re $prompt {{exitmsg "Password authentication successful" 0}}
+                "Authentication failure" {{exitmsg "Authentication failure" 1}}
+                timeout {{exitmsg "Unexpected output" 201}}
+                eof {{exitmsg "Unexpected end of file" 202}}
             }}
 
-            puts "expect result: Unexpected code path"
-            exit 203
-        """
+            exitmsg "Unexpected code path" 203
+        """,
+        verbose=False
         )
 
         if result.rc > 200:
             raise ExpectScriptError(result.rc)
 
-        return result.rc, result.stdout, result.stderr
+        expect_data = result.stdout_lines[-3:]
+
+        # Get command exit code.
+        cmdrc = int(expect_data[2].split(":")[1].strip())
+
+        # Alter stdout, first line is spawned command, the last three are our expect output.
+        stdout = "\n".join(result.stdout_lines[1:-3])
+
+        return result.rc, cmdrc, stdout, result.stderr
 
     def password(self, username: str, password: str) -> bool:
         """
@@ -242,7 +267,7 @@ class SUAuthenticationUtils(MultihostUtility[MultihostHost]):
         :return: True if authentication was successful, False otherwise.
         :rtype: bool
         """
-        rc, _, _ = self.password_with_output(username, password)
+        rc, _, _, _ = self.password_with_output(username, password)
         return rc == 0
 
     def password_expired(self, username: str, password: str, new_password: str) -> bool:

```

I had to change `send \x04;` to `catch close`, otherwise I got an error when EOF was already sent/received. Same mistake is in ssh as well, I'll fix that by force push.

pbrezina:
The `catch` is important, `close` is just better then sending EOF in unicode.

danlavu:
Rebased, updated the doc string, it was missing the second tuple value.

PR #87, opened 2024-02-20
CLIENT: fix "non-privileged" feature detection

alexey-tikhonov:
Using option name makes it less prone to formatting / wording changes.


pbrezina:
It works... but maybe it would be better to detect a symbol using nm? Is there any specific symbol to use?

For long term, I opened https://github.com/SSSD/sssd/issues/7202

alexey-tikhonov:
> It works... but maybe it would be better to detect a symbol using nm? Is there any specific symbol to use?

I think option name is better. Internal symbols (not part of exported ABI) is an implementation detail, subject to change.
And actually I can't think of any, or there is nothing "fundamental" that is built conditionally (maybe a couple of micro helpers).

PR #88, opened 2024-02-21
roles:  adding optional email parameter to active directory user add

danlavu:


danlavu:
We have a test case worth automating https://github.com/SSSD/sssd/pull/7174 , using the email instead of the userid to login. To adequately test this, we need to be able to change the email value of the user created in AD.

danlavu:
Test case https://github.com/SSSD/sssd/pull/7206

pbrezina:
LGTM, can you please add it also to Samba? There's `--mail-address` parameter. It would also be nice if we are able to change it in `modify()`.

sssd_test_framework/roles/ad.py
```diff
@@ -610,9 +611,14 @@ def add(
         :type gecos: str | None, optional
         :param shell: Login shell, defaults to None
         :type shell: str | None, optional
+        :param email: Email, defaults to None
```
pbrezina, line 614: `defaults to None (= user@domain)`, same in samba

pbrezina (changes requested):

PR #89, opened 2024-02-26
roles: adding gpo functions to AD role

danlavu:
- create gpo class to manage group policy objects; add(), link(), unlink(), permissions(), policy(), remove(), 
- moved attrs_to_hash to the roles class so other classes can inherit it
- updated adobject with get_sid method 
- added specific gpo backup and restore methods
- updated authentication expect catch for users not in a policy
- create ad computer class to move the computer object
- increased default authentication timeout in utils.authentication.py

### DIFF
--- a/docs/guides/testing-gpo.rst
+++ b/docs/guides/testing-gpo.rst
@@ -1,0 +1,66 @@
+Testing AD GPO HBAC
+====================
+
+:class:`~sssd_test_framework.roles.ad.AD.GPO`
+provides Group Policy Objects (GPO) management to configure GPO policies on AD. Allowing us
+to manage Host Based Access Control(HBAC) from the domain controller. GPOs are windows policies,
+consisting of computer and user configurations, registry keys, administrative template, security
+settings that are deployed to computers on the domain.
+
+.. code-block:: python
+    :caption: Test allow and deny users using GPOs
+
+    @pytest.mark.topology(KnownTopology.AD)
+    def test_ad__gpo_is_set_to_enforcing(client: Client, ad: AD):
+
+        allow_user = ad.user("allow_user").add()
+        deny_user = ad.user("deny_user").add()
+
+        ad.gpo("test policy").add().policy(
+            {
+                "SeInteractiveLogonRight": [allow_user, ad.group("Domain Admins")],
+                "SeRemoteInteractiveLogonRight": [allow_user, ad.group("Domain Admins")],
+                "SeDenyInteractiveLogonRight": [deny_user],
+                "SeRemoteDenyInteractiveLogonRight": [deny_user],
+            }
+        ).link()
+
+        client.sssd.domain["ad_gpo_access_control"] = "enforcing"
+        client.sssd.start()
+
+        assert client.auth.ssh.password(username="allow_user", password="Secret123")
+        assert not client.auth.ssh.password(username="deny_user", password="Secret123")
+
+Policies
+========
+SSSD uses the following security keys to manage access control.
+
+* SeInteractiveLogonRight: User or group is permitted to log in locally, TTY sessions and su (required).
+* SeRemoteInteractiveLogonRight: User or group is permitted to log in through SSH.
+* SeDenyInteractiveLogonRight: User or group is denied to log in locally. (required)
+* SeDenyRemoteInteractiveLogonRight: User or group is denied to log in through SSH.
+
+The *Remote* keys can be omitted, in which case the value is copied from the other similar key. The above
+policy examples are the same.
+
+.. code-block:: python
+    :caption: Remote keys
+
+        ad.gpo("test policy").add().policy(
+            {
+                "SeInteractiveLogonRight": [allow_user, ad.group("Domain Admins")],
+                "SeDenyInteractiveLogonRight": [deny_user],
+            }
+        ).link()
+
+.. note::
+   When configuring the policy, an Administrators group must be added to login.
+
+
+Inheritance
+===========
+GPOs are created and linked to a targets, sites, domains, or OUs. Local settings are processed first,
+then sites, domains and lastly OUs. Making GPOs linked to the OU take highest precedence. The values
+are keys, so the values will not be constructed from all the policies, it will be replaced by the
+policy with the higher priority. GPOs can be set to enforcing, which then no policy will override
+the enforced policy. Lastly the GPO does have an order, so the order can be defined manually.
--- a/sssd_test_framework/hosts/ad.py
+++ b/sssd_test_framework/hosts/ad.py
@@ -52,6 +52,12 @@ class ADHost(BaseDomainHost):
         # Lazy properties
         self.__naming_context: str | None = None
 
+        # GPO policy file location
+        self._backup_gpo_location: str | None = None
+
+        # Computers file location
+        self._backup_computers_location: str | None = None
+
     @property
     def features(self) -> dict[str, bool]:
         """
@@ -94,19 +100,18 @@ class ADHost(BaseDomainHost):
 
     def backup(self) -> None:
         """
-        Perform limited backup of the domain controller data. Currently only
-        content under ``$default_naming_context`` is backed up.
-
-        This is done by performing simple LDAP search on the base dn. This
-        operation is usually very fast.
+        Perform limited backup of the domain controller data. Content under
+        ``$default_naming_context``. Site, groupPolicyContainer and computer
+        objects are explicitly exported so GPO setup can be undone. These
+        operations are usually very fast.
         """
         self.ssh.run(
             rf"""
-                $backup = "C:\multihost_backup.txt"
+                $backup = "C:\\multihost_backup.txt"
                 if (Test-Path $backup) {{
                     Remove-Item $backup
                 }}
-                $result = Get-ADObject -SearchBase '{self.naming_context}' -Filter "*"
+                $result = Get-ADObject -SearchBase "{self.naming_context}" -Filter "*"
                 foreach ($r in $result) {{
                     $r.DistinguishedName | Add-Content -Path $backup
                 }}
@@ -114,32 +119,126 @@ class ADHost(BaseDomainHost):
         )
         self._backup_location = "C:\\multihost_backup.txt"
 
+        self.ssh.run(
+            rf"""
+                $backup = "C:\\multihost_gpo_backup.txt"
+                if (Test-Path $backup) {{
+                    Remove-Item $backup
+                }}
+                $filter = "objectClass=GroupPolicyContainer"
+                $result = Get-ADObject -SearchBase "{self.naming_context}" -LDAPFilter ($filter)
+                foreach ($r in $result) {{
+                    $r.DistinguishedName | Add-Content -Path $backup
+                }}
+            """
+        )
+        self._backup_gpo_location = "C:\\multihost_gpo_backup.txt"
+
+        self.ssh.run(
+            rf"""
+                $backup = "C:\\multihost_computers_backup.txt"
+                $basedn = '{self.naming_context}'
+                $computersdn = "cn=computers,$basedn"
+                if (Test-Path $backup) {{
+                    Remove-Item $backup
+                }}
+                $filter = "objectClass=computer"
+                $result = Get-ADObject -SearchBase "$computersdn" -LDAPFilter ($filter)
+                foreach ($r in $result) {{
+                    $r.DistinguishedName | Add-Content -Path $backup
+                }}
+            """
+        )
+        self._backup_computers_location = "C:\\multihost_computers_backup.txt"
+
     def restore(self) -> None:
         """
         Perform limited restoration of the domain controller state.
 
         This is done by removing all records under ``$default_naming_context``
-        that are not present in the original state.
-        """
-        if not self._backup_location:
-            return
+        and that are not present in the original state.
 
-        self.ssh.run(
-            rf"""
-        $backup = Get-Content "{self._backup_location}"
-        $result = Get-ADObject -SearchBase '{self.naming_context}' -Filter "*"
-        foreach ($r in $result) {{
-            if (!$backup.contains($r.DistinguishedName)) {{
-                Write-Host "Removing: $r"
-                Try {{
-                   Remove-ADObject -Identity $r.DistinguishedName -Recursive -Confirm:$False
-                }} Catch [Microsoft.ActiveDirectory.Management.ADIdentityNotFoundException] {{
-                    # Ignore not found error as the object may have been deleted by recursion
-                }}
-            }}
-        }}
+        If GPOs are found, some additional steps are performed. The policy directory
+        located at 'C:\\Windows\\SYSVOL\\domain\\Policies\\{{GUID}}' is deleted.
+        Before removing the GPO, the GPO needs to be unlinked from the target object.
+        There is logic to run through the GPOs that were not present, unlink them and
+        then removed.
 
-        # If we got here, make sure we exit with 0
-        Exit 0
+        The client computer object may move to a different location during a test.
+        There is a check to ensure that the object is in 'cn=computers' otherwise
+        the object will be deleted when attempting to restore the computer state.
         """
-        )
+        if self._backup_computers_location:
+            self.ssh.run(
+                rf"""
+                    $basedn = '{self.naming_context}'
+                    $backup = Get-Content '{self._backup_computers_location}'
+                    $result = Get-ADObject -SearchBase $basedn -Filter "*"
+                    $computersdn = "cn=computers,$basedn"
+                    foreach ($b in $backup) {{
+                        if ($b -Like "*$computersdn*") {{
+                            $cn = $b.split(",")[0].split("=")[1].ToUpper()
+                            $computer = (Get-ADComputer $cn).DistinguishedName
+                            if ($computer -NotLike "*$computersdn*") {{
+                                Write-Host Moving: $computer : $computersdn
+                                Move-ADObject -Identity "$computer" -TargetPath "$computersdn" -Confirm:$false
+                            }}
+                        }}
+                    }}
+                    Exit 0
+                """
+            )
+
+        if self._backup_gpo_location:
+            self.ssh.run(
+                rf"""
+                    $basedn = '{self.naming_context}'
+                    $sitesdn = "cn=Default-First-Site-Name,cn=sites,cn=configuration,$basedn"
+                    $backup = Get-Content '{self._backup_gpo_location}'
+                    $filter = ("objectClass=GroupPolicyContainer")
+                    $gpo = Get-ADObject -SearchBase $basedn -Properties "*" -LDAPFilter $filter
+                    $sites = Get-ADObject -Identity $sitesdn -Properties "*"
+                    $link = Get-ADObject -SearchBase $basedn -Properties "*" -LDAPFilter ("gplink=*")
+                    foreach ($r in $gpo) {{
+                        if (!$backup.contains($r.DistinguishedName)) {{
+                            if ($sites.gplink -Like "*$r*") {{
+                                Remove-GPLink -GUID $r.Name -Target "Default-First-Site-Name"
+                                Write-Host Removing: "*$r*" from Default-First-Site-Name
+                            }}
+                            foreach ($g in $link) {{
+                                if ($g.gplink -Like "*$r*") {{
+                                    Remove-GPLink -GUID $r.Name -Target $g
+                                    Write-Host Removing: "*$r*" from $g
+                                }}
+                            }}
+                            Remove-GPO -GUID $r.Name
+                            $path = Join-Path C:\Windows\SYSVOL\domain\Policies\ $r.Name
+                            Remove-Item $path -Recurse -Confirm:$false -Force
+                            Write-Host "Deleting directory: $path"
+                        }}
+                    }}
+                    # An extra gplink clear on the sites target.
+                    Set-ADObject -Identity $sitesdn -Clear gPLink
+                    Exit 0
+                """
+            )
+
+        if self._backup_location:
+            self.ssh.run(
+                rf"""
+                    $backup = Get-Content '{self._backup_location}'
+                    $result = Get-ADObject -SearchBase '{self.naming_context}' -Filter "*"
+                    foreach ($r in $result) {{
+                        if (!$backup.contains($r.DistinguishedName)) {{
+                            Write-Host "Removing: $r"
+                            Try {{
+                                Remove-ADObject -Identity $r.DistinguishedName -Recursive -Confirm:$false
+                            }} Catch [Microsoft.ActiveDirectory.Management.ADIdentityNotFoundException] {{
+                                # Ignore not found error as the object may have been deleted by recursion
+                            }}
+                        }}
+                    }}
+                    # If we got here, make sure we exit with 0
+                    Exit 0
+                """
+            )
--- a/sssd_test_framework/roles/ad.py
+++ b/sssd_test_framework/roles/ad.py
@@ -8,7 +8,7 @@ from pytest_mh.cli import CLIBuilderArgs
 from pytest_mh.ssh import SSHClient, SSHPowerShellProcess, SSHProcessResult
 
 from ..hosts.ad import ADHost
-from ..misc import attrs_include_value, attrs_parse
+from ..misc import attrs_include_value, attrs_parse, attrs_to_hash
 from .base import BaseObject, BaseWindowsRole, DeleteAttribute
 from .ldap import LDAPNetgroupMember
 from .nfs import NFSExport
@@ -20,8 +20,10 @@ __all__ = [
     "ADGroup",
     "ADObject",
     "ADOrganizationalUnit",
+    "ADComputer",
     "ADSudoRule",
     "ADUser",
+    "GPO",
 ]
 
 
@@ -252,7 +254,7 @@ class AD(BaseWindowsRole[ADHost]):
         """
         Get organizational unit object.
 
-        .. code-blocK:: python
+        .. code-block:: python
             :caption: Example usage
 
             @pytest.mark.topology(KnownTopology.AD)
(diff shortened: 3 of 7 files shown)


sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1500,309 @@ def __get_info(
...
+
+    Provides shortcuts for command execution and implementation of :meth:`get`
+    and :meth:`delete` methods.
+    """
+
+    def __init__(
+        self,
+        role: AD,
+        name: str,
+        target: str | None = "Default-First-Site-Name",
```
pbrezina, line 1515: target is only used in GPO class, it should not be here.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1500,309 @@ def __get_info(
...
+        :param target: Group policy target.
+        :type target: str
+        """
+        super().__init__(role, name, target)
+
+        self.permissions: GPPermissions = GPPermissions(role, self.name)
+
+        self.key: GPKey = GPKey(role, self.name)
+
+    def link(self) -> None:
```
pbrezina, line 1779: I would move target as a parameter of link instead of setting it in constructor.

pbrezina (changes requested):
Hi Dan, see some comments inside.

I think it would be for the best to start with docs/guides/testing-gpo.rst and add some thorough examples so I can see how do you want it to be used. The code is currently incomplete so I can't always tell what is the intention.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1500,309 @@ def __get_info(
...
+            "Key": (self.cli.option.VALUE, self.key),
+            "ValueName": (self.cli.option.VALUE, self.key + value_name),
+            "Value": (self.cli.option.VALUE, value),
+            "Type": (self.cli.option.VALUE, value_type),
+            "Action": (self.cli.option.VALUE, action),
+        }
+
+        self._add("Set-GPPrefRegistryValue", params)
+
+    def set_ssh(self, users: list[dict[str, bool, bool]]) -> None:
```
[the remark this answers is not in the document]
danlavu: Hahaha... Neither do I yet.  ;)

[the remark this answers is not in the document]
danlavu: Okay, I had to take some time and learn more Python, in particular classes, inheritance and instantiation to implement this.  So, this won't be used at all. The following "keys" that we need to edit, 'SeInteractiveLogon, SeRemoteInteractiveLogon, SeDenyInteractiveLogon, SeDenyRemoteInteractiveLogon' cannot be modified using PowerShell, because they are not in the registry. They are stored in the security policy database. This is why, I had to previously, backup the GPO, migrate it to the new domain, and import them. There is a lot of negatives implementing it this way but we may have to. 

There is something I am going to try first to avoid using templates, hopefully it'll work.

danlavu:
@pbrezina this is ready for a cursory review. It's mostly working, the files and policy is created but it's not created correctly. I've been looking at this https://www.tenaka.net/post/how-to-create-gpos-with-restricted-groups-using-powershell.

sssd_test_framework/hosts/ad.py
```diff
@@ -94,33 +94,75 @@ def disconnect(self) -> None:
 
     def backup(self) -> None:
         """
-        Perform limited backup of the domain controller data. Currently only
-        content under ``$default_naming_context`` is backed up.
+        Perform limited backup of the domain controller data. Content under
+        ``$default_naming_context`` is backed up and list of GPOs.
```
pbrezina, line 98: $default_naming_context and list of GPOs is backed up.

sssd_test_framework/hosts/ad.py
```diff
@@ -94,33 +94,75 @@ def disconnect(self) -> None:
...
-            """
+        $backup = "C:\multihost_backup.txt"
+        if (Test-Path $backup) {{
+              Remove-Item $backup
+            }}
+        $result = Get-ADObject -SearchBase '{self.naming_context}' -Filter "*"
+        foreach ($r in $result) {{
+           $r.DistinguishedName | Add-Content -Path $backup
+            }}
+        """
```
pbrezina, line 113: Seems like you only changed the indentation? Can you avoid this change?

sssd_test_framework/hosts/ad.py
```diff
@@ -94,33 +94,75 @@ def disconnect(self) -> None:
...
+        if (Test-Path $backup) {{
+            Remove-Item $backup
+        }}
+        $search_base = "cn=policies,cn=system,{self.naming_context}"
+        $object_class = "groupPolicyContainer"
+        $result = Get-ADObject -SearchBase "$search_base" -Filter "ObjectClass -eq '$object_class'"
+            foreach ($r in $result) {{
+            $r.ObjectGUID | Add-Content -Path $backup
+            }}
+        """
```
pbrezina, line 129: Please, indent it correctly.

black does not indent multiline scripts nicely as it must keep the blank lines intact. But we want nice indentation.

sssd_test_framework/hosts/ad.py
```diff
@@ -94,33 +94,75 @@ def disconnect(self) -> None:
...
+            rf"""
+        $backup = "C:\multihost_gpo_backup.txt"
+        if (Test-Path $backup) {{
+            Remove-Item $backup
+        }}
+        $search_base = "cn=policies,cn=system,{self.naming_context}"
+        $object_class = "groupPolicyContainer"
+        $result = Get-ADObject -SearchBase "$search_base" -Filter "ObjectClass -eq '$object_class'"
+            foreach ($r in $result) {{
+            $r.ObjectGUID | Add-Content -Path $backup
```
pbrezina, line 127: Indent.

sssd_test_framework/hosts/ad.py
```diff
@@ -94,33 +94,75 @@ def disconnect(self) -> None:
...
+        foreach ($r in $result) {{
+            if (!$backup.contains($r.ObjectGUID)) {{
+                Write-Host "Removing:" $r.ObjectGUID
+                Remove-GPO -GUID $r.ObjectGUID -Confirm:$False
+                Remove-Item (Join-Path C:\Windows\SYSVOL\domain\Policies\ $r.Name) -Recurse -Confirm:$false -Force
+            }}
+        }}
+
+        Exit 0
+        """
```
pbrezina, line 163: Reindent.

sssd_test_framework/hosts/base.py
```diff
@@ -57,15 +57,19 @@ def __init__(self, *args, **kwargs) -> None:
         self._backup_location: str | None = None
         """Backup file or folder location."""
 
+        self._backup_gpo_location: str | None = None
```
pbrezina, line 60: This should go to ADHost.

sssd_test_framework/hosts/base.py
```diff
@@ -57,15 +57,19 @@ def __init__(self, *args, **kwargs) -> None:
...
     def pytest_teardown(self) -> None:
         """
         Called once after all tests are finished.
         """
-        if self._backup_location is not None:
-            if self.ssh.shell is SSHPowerShellProcess:
-                self.ssh.exec(["Remove-Item", "-Force", "-Recurse", self._backup_location])
-            else:
-                self.ssh.exec(["rm", "-fr", self._backup_location])
+        for i in [self._backup_location, self._backup_gpo_location]:
```
pbrezina, line 67: backup_gpo_location should be handled in ADHost only.

sssd_test_framework/roles/ad.py
```diff
@@ -281,6 +284,51 @@ def test_example(client: Client, ad: AD):
...
+    ) -> GPO:
+        """
+        Get group policy object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.AD)
+            def test_ad__gpo_is_set_to_enforcing(client: Client, ad: AD, sssd_service_user: str):
+
```
pbrezina, line 302: Avoid blank line.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1565,313 @@ def __get_info(
...
+        Execute cmdlet.
+
+        .. code-block:: console
+
+            $ $cmdlet $command_group $args
+            for example >>> New-GPO
+                            New-GPLink
+
+        :description: This function is modeled after ADObject._exec with one key difference.
+        The cmdlet is taken as a parameter instead of being concatenated and constructed, because
```
pbrezina, line 1607: Indent.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1565,313 @@ def __get_info(
...
+        elif args is None:
+            args = ""
+
+        formatted = "" if format_with is None else f"| {format_with}"
+
+        return self.role.host.ssh.run(
+            f"""
+            Import-Module GroupPolicy, PSIni
+            {cmdlet} {args} {formatted}
+        """,
```
pbrezina, line 1633: Indent.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1565,313 @@ def __get_info(
...
+            **kwargs,
+        )
+
+    def get(self, attrs: list[str]) -> dict[str, list[str]]:
+        """Get group policy object.
```
pbrezina, line 1638: Put description to new line.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1565,313 @@ def __get_info(
...
+        return str(self.get(["Id"]).get("Id")[0]).strip()
+
+    def get_gpo_path(self) -> str:
+        """
+        Get group policy file path.
+
+        :return: Group policy path.
+        :rtype: str
+        """
+        gpo_path = path.join(
```
pbrezina, line 1684: Using `WindowsPath` from pathlib will be better.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1565,313 @@ def __get_info(
...
+            self.id: str = self.get_id()
+            """Group policy id."""
+
+            self.path: str = self.get_gpo_path()
+            """Group policy file path."""
+
+            self.file_path = '"' + path.join(self.path, "GptTmpl.inf") + '"'
+            """Group policy GptTmpl file path."""
+
+        self.link() if link is True else None
```
pbrezina, line 1748: Yak.

pbrezina (changes requested):
Hi, please, see some minor comments.

I have some questions about the usage, we'll discuss it later on the call today.

danlavu:
@pbrezina @thalman, this is ready for review and not a cursory one.

danlavu:
https://github.com/SSSD/sssd/pull/7249/files 

The tests need to be written, but here is the PR, and it does contain one working test case. 

`
============================= test session starts ==============================
collecting ... 

Selected tests will use the following hosts:
  client: client.test
  ad: dc.ad.test
  nfs: nfs.test

collected 2 items

tests/test_ad.py::test_ad__gpo_is_set_to_enforcing[root] (ad) 
tests/test_ad.py::test_ad__gpo_is_set_to_enforcing[sssd] (ad) 

======================== 1 passed, 1 skipped in 35.61s =========================
PASSED     [ 50%]SKIPPED    [100%]
Skipped: SSSD was built without support for running under non-root

Process finished with exit code 0
`

sssd_test_framework/roles/ad.py
```diff
@@ -281,6 +282,40 @@ def test_example(client: Client, ad: AD):
...
 
+    def gpo(self, name: str) -> GPO:
+        """
+        Get group policy object.
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.AD)
+            def test_ad__gpo_is_set_to_enforcing(client: Client, ad: AD, sssd_service_user: str):
```
pbrezina, line 293: sssd_servise_user is not defined nor used, you can omit it

sssd_test_framework/roles/ad.py
```diff
@@ -309,6 +344,29 @@ def test_example(client: Client, ad: AD):
 
         return ADSudoRule(self, name, basedn)
 
+    def attrs_to_hash(self, attrs: dict[str, Any]) -> str | None:
```
pbrezina, line 347: Can you move it into misc/__init__.py and add unit test for it to tests/test_misc.py?

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1545,216 @@ def __get_info(
...
+        """Group policy search base."""
+
+        self._dn: str = self.get("DistinguishedName")
+        """Group policy dn."""
+
+        self._cn: str = self.get("CN")
+        """Group policy cn."""
+
+    def get(self, key: str) -> str:
+        """Get group policy attributes.
```
pbrezina, line 1575: Add subject to new line.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1545,216 @@ def __get_info(
...
+        self._search_base: str = f"cn=policies,cn=system,{self.role.host.naming_context}"
+        """Group policy search base."""
+
+        self._dn: str = self.get("DistinguishedName")
+        """Group policy dn."""
+
+        self._cn: str = self.get("CN")
+        """Group policy cn."""
+
+    def get(self, key: str) -> str:
```
pbrezina, line 1574: What is the use case for this method? Why does it allow to retrieve only a single key instead of set of attributes?

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1545,216 @@ def __get_info(
...
+                else:
+                    value = result[i].split(":")
+                    return value[1].strip()
+            i += 1
+
+    def delete(self) -> None:
+        """
+        Delete group policy object.
+
+        Note, GPLinks and GPO files are removed test during cleanup.
```
pbrezina, line 1604: `are removed test during cleanup` incorrect wording?

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1545,216 @@ def __get_info(
...
+        """
+        self.role.host.ssh.run(f'Remove-GPO -Guid "{self._cn}" -Confirm:$false')
+
+    def add(self) -> GPO:
+        """
+        Add a group policy object.
+
+        This creates the GPO and to update the security database, the SecEdit path is then
```
pbrezina, line 1612: `and to' -> `and`?

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1545,216 @@ def __get_info(
...
+        created with the GptTempl.inf file. The file populated with the generic headers but
```
pbrezina, line 1613: The file **is** populated?

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1545,216 @@ def __get_info(
...
+        :return: Group policy object
+        :rtype: GPO
+        """
+        self.role.host.ssh.run(f'Remove-GPLink -Guid "{self._cn}" -Target "{target}"')
+        self.host._backup_gpo.remove(f'Remove-GPLink -Guid "{self._cn}" -Target "{target}"')
+
+        return self
+
+    def permissions(self, level: str, target_type: str | None = "User") -> GPO:
+        """Configure group policy object permissions.
```
pbrezina, line 1670: Add to new line.

sssd_test_framework/roles/ad.py
```diff
@@ -1490,4 +1545,216 @@ def __get_info(
...
+        _keys: list[str] = [
+            "SeInteractiveLogonRight",
+            "SeRemoteInteractiveLogonRight",
+            "SeDenyInteractiveLogonRight",
+            "SeDenyRemoteInteractiveLogonRight",
+        ]
+
+        for i in _keys:
+            if i not in logon_rights.keys():
+                raise KeyError
```
pbrezina, line 1715: Please, provide message to the KeyError. E.g. `f"Expected {_keys} but got {i}`.

You can also use the Literal type instead of str, e.g. `Literal["SeInteractiveLogonRight", ...]` this way it will be checked by mypy.

sssd_test_framework/hosts/ad.py
pbrezina: Still there. I am not native English speaker but I believe the wording should be `Content under ``$default_naming_context`` and list of GPOs is backed up.`

Aren't GPOs actually mostly covered by existing backup/restore? `cn=policies,cn=system,{self.naming_context}` is under the naming context so this should be already taken care of? Then you only  need to unlink it (since we're going to use Default-First-Site-Name which is already present) and remove the ini file, right? Maybe we can use something along these lines to existing restore:

```
if (isgpo($r)) {
    unlink gpo
    remove ini file
}
```

pbrezina: Doesn't removing the GPO object automatically trigger unlink and ini file removal?

docs/guides/testing-gpo.rst
```diff
@@ -0,0 +1,44 @@
...
+====================
+
+Class :class:`sssd_test_framework.roles.ad.gpo`
+provides Group Policy Objects (GPO) management to test Host Based Access Control (HBAC) in Active Directory (AD).
+GPOs are windows settings, configurations, registry keys, security settings that is managed by the domain controllers.
+They are applied to an OU, domain or the site.
+
+SSSD uses the following security keys to manage access control.
+
+* SeInteractiveLogonRight : User or group is permitted to log in locally, TTY sessions and su.
```
pbrezina, line 11: Remove space from left side of the colon?

sssd_test_framework/roles/ad.py
```diff
@@ -518,19 +576,16 @@ def _attrs_to_hash(self, attrs: dict[str, Any]) -> str | None:
...
-                if isinstance(value, list):
-                    values = [f'"{x}"' for x in value]
-                    out += f'"{key}"={",".join(values)};'
-                else:
-                    out += f'"{key}"="{value}";'
+        return self.role.attrs_to_hash(attrs)
 
-        if not out:
-            return None
+    def get_sid(self) -> str:
```
pbrezina, line 581: Make it a property? E.g.:

```python
@property
def sid(self) -> str:
    if self.__sid is not None:
        self.__sid = str(self.get(["SID"]).get("SID")[0]).strip()

    return self.__sid
```

pbrezina (changes requested):
Hi Dan, I think this looks really great. Great job!

See some minor comments inside. Additionally, I would like to check if we can simplify backup and restore since I don't particularly like you touching a property (`_backup_gpo`) of the Host object, it is something I'd like to avoid if possible.

I think removing a GPO should automatically trigger at least unlink operation, if not ini file deletion.

pbrezina:
And of course PR CI is failing. Please, make it green.

Would it be possible to write this also for Samba?

sssd_test_framework/roles/ad.py
pbrezina: Haha, you are right. Yes, it makes sense to keep it in AD because we use it only there. However, putting it to misc will allow you to unit test it and this time my mind leaned to this :-) It's your call, whatever you prefer.

sssd_test_framework/hosts/ad.py
danlavu: understand them to be the same, my wording is a little less formal, fixed. 

You'd think that removing the GPO would, but it does not. The GPOs reside in three locations, and one doesn't really matter. 

1. grouppolicycontainer object object, cn=policies,cn=system, removing the LDAP object is covered by self.naming_context still. The GPO just looks at the object to find the GUID, so it can construct the path to remove the gpo directory. Running the Remove-GPO in _backup_gpo_location isn't necessary.  
 
2. This is the challenging part, the target object, by default cn=sites,cn=configuration has an attribute 'gplink' that references the linked GPO. This isn't restored by self.naming_context because the object is not new. The target can also change to the domain or a specific OU. I couldn't find anything in the GPO object that tells me what is linked to it. Doing the adverse command when called is easiest, just for gplink though.

3. the inf file that has the gpo id in the path. We can leave this, we'll have artifacts directories in c:\windows\sysvol\domain\policies\{GUID}

sssd_test_framework/roles/ad.py
danlavu: Fewer lines of code and a simple way to get the value was the goal. 

I tried using parse_attrs, right now on the multiline values, the string isn't stripped correctly, so the values look like "...dc=ad,dc                                                     =test", which doesn't work injecting the values into powershell code. I'm going to fix that later.

sssd_test_framework/roles/ad.py
danlavu: That note should be here, it should be on the restore function, so I'm removing it.

sssd_test_framework/roles/ad.py
danlavu: Updated the description. Should make more sense now.

sssd_test_framework/roles/ad.py
danlavu: It is. 
`
$content = @{{'Unicode'=@{{'Unicode'='yes'}};'Version'=@{{'signature'='"$CHICAGO$"';'Revision'='1'}}}}
Out-IniFile -InputObject $content -FilePath $file
`

`
[Unicode]
Unicode=yes

[Version]
signature=$CHICAGO$
Revision=1
`

Relies on this https://github.com/SSSD/sssd-ci-containers/pull/83

sssd_test_framework/roles/ad.py
danlavu: I spent a couple of hours learning and trying to implement Literal, then started poking at TypedDict. I updated it to provide a message for now.

danlavu:
> And of course PR CI is failing. Please, make it green.
> 
> Would it be possible to write this also for Samba?

Sure, I'd like to understand the abstracted methods better so we can execute the gpo tests using GenericAD topology. Right now it's not a priority though, but I'll create a task for it.

pbrezina:
> > And of course PR CI is failing. Please, make it green.
> > Would it be possible to write this also for Samba?
> 
> Sure, I'd like to understand the abstracted methods better so we can execute the gpo tests using GenericAD topology. Right now it's not a priority though, but I'll create a task for it.

It would be better to do this right away for three reasons: 1) We can run GPO tests in PR CI, 2) Sometimes making the API generic requires some changes and sacrifices. 3) But it is usually very easy. We can work on this together if you want.

You want to start by copy pasting what you have now into the Samba role, then find the alternative commands that you need to run. In this case, it will be `samba-tool gpo create/link`, policy will probably stay the same. Backup restore - we do full backup/restore in Samba so this is for free, but you need to make sure to also restore the original ini files on the disk.

sssd_test_framework/hosts/ad.py
danlavu: I've entirely changed this. Instead of backing line items, and using a rollback variable, all the gpo related objects are exported to an csv and imported back in.

docs/guides/testing-gpo.rst
```diff
@@ -0,0 +1,50 @@
+Testing AD GPO HBAC
+====================
+
+:class:`~sssd_test_framework.roles.ad.AD.GPO`
+provides Group Policy Objects (GPO) management to configure GPO policies on AD. Allowing us
+to manage Host Based Access Control(HBAC) from the domain controller. GPOs are windows policies,
+consisting of compute rand user configurations, registry keys, administrative template, security
```
thalman, line 7: compute rand -> computer and

docs/guides/testing-gpo.rst
```diff
@@ -0,0 +1,50 @@
...
+* SeRemoteInteractiveLogonRight: User or group is permitted to log in through SSH.
+* SeDenyInteractiveLogonRight: User or group is denied to log in locally.
+* SeDenyRemoteInteractiveLogonRight: User or group is denied to log in through SSH.
+
+.. note:: When configuring the policy, an Administrators group must be added to login.
+
+
+Inheritance
+===========
+GPOs are created and linked to a target; sites, domains and OUs. Local settings are processed first,
```
thalman, line 48: `target; sites` should be probably `target sites`? (extra semicolon)

thalman (changes requested):
Great job, just minor typos in doc. Thanks Dan!

danlavu:
@pbrezina, I changed a few things that should probably be in another pull request or requests. I hope we can be a little flexible about this. 

* ADComputer class, contains one method move.
* auth.ssh.password() with additional expect responses  

@pbrezina  Okay, I'll start working on it. In this pull request?

danlavu:
I had to rework the restore code again. The challenge has been and is, you cannot unlink the GPO with just the name, you need the target object DN. So there is an extra step that searches for the gplink attribute and makes sure it's found before its removed. 

"cn=Default-First-Site-Name,cn=sites,cn=configuration" is not included in the ldap query unless the searchbase is explicitly set to cn=configuration. These notes will help understand the code a little better.

sssd_test_framework/hosts/ad.py
danlavu: Okay, I'm wrong, well my code was wrong, which led me to a poor conclusion. Remove-GPO does remove the link and all the time spent trying to remove the gplink before removing the GPO was a waste of nyRemote
and add time. As long as I learn something right?

danlavu:
Okay, all debugging is done, especially the clean up code is working.

danlavu:
Sorry for all the pushes, I was using it to get the results from mypy, I have since started running mypy locally.

danlavu:
Okay... no more changes, no more debugging, tox checks are green. 

The following test code works. https://github.com/SSSD/sssd/pull/7249/files  

```
tests/test_gpo.py::test_gpo__is_set_to_enforcing[root] (ad) PASSED                                                                                             [ 12%]
tests/test_gpo.py::test_gpo__is_set_to_enforcing[sssd] (ad) SKIPPED (SSSD was built without support for running under non-root)                                [ 25%]
tests/test_gpo.py::test_gpo__is_set_to_permissive[root] (ad) PASSED                                                                                            [ 37%]
tests/test_gpo.py::test_gpo__is_set_to_permissive[sssd] (ad) SKIPPED (SSSD was built without support for running under non-root)                               [ 50%]
tests/test_gpo.py::test_gpo__is_set_to_disabled[root] (ad) PASSED                                                                                              [ 62%]
tests/test_gpo.py::test_gpo__is_set_to_disabled[sssd] (ad) SKIPPED (SSSD was built without support for running under non-root)                                 [ 75%]
tests/test_gpo.py::test_gpo__site_over_domain_over_ou_inheritance[root] (ad) PASSED                                                                            [ 87%]
tests/test_gpo.py::test_gpo__site_over_domain_over_ou_inheritance[sssd] (ad) SKIPPED (SSSD was built without support for running under non-root)               [100%]

============================================================== 4 passed, 4 skipped in 128.50s (0:02:08) ==============================================================
```

sssd_test_framework/roles/ad.py
pbrezina: Oh, it was not a question, but suggestion to the text :-) There is "The file populated" (without "is")

danlavu:
@pbrezina  it is.

danlavu:
The PR for the tests is here, https://github.com/SSSD/sssd/pull/7249/files , only some code is written, but all the docstrings are done and the test cases/steps are up for review.

danlavu:
I added one additional change, writing 'test_gpo__works_in_offline_mode', the default expect timeout was too short and threw 203 errors asserting denied users. 

```
[root@client /]# time ssh deny_user1@ad.test@client.test
deny_user1@ad.test@client.test's password: 
Connection closed by 172.16.100.40 port 22

real    0m34.815s
user    0m0.006s
sys     0m0.003s
```

pbrezina (commented):
Hi, thank you. The code looks good, I will run some tests and then ack.

pbrezina (approved):

PR #92, opened 2024-03-21
local: Use -f to execute groupdel

justin-stephenson:


pbrezina:
Why? `--force` is valid option. If there is some problem, please, add description to the commit message.

alexey-tikhonov:
> Why? `--force` is valid option. If there is some problem, please, add description to the commit message.

'shadow-utils' tool doesn't support '--force' on centos-8 package.

alexey-tikhonov:
It's an addition to incomplete 66528081e63193d1e66840e8816c26d4d40f4e08

justin-stephenson:
> Why? `--force` is valid option. If there is some problem, please, add description to the commit message.

Updated.

PR #93, opened 2024-03-22
hosts: add ipa restart after restore

jakub-vavra-cz:
IPA got stuck after restoring data in some cases, adding restart to alleviate the issue.


madhuriupadhye (approved):
LGTM

PR #94, opened 2024-03-22
ci: Remove the broken 3.x from ci.

jakub-vavra-cz:
The ci for 3.x is clearly broken and it is blocking other fixes.


jakub-vavra-cz:
Did not help, the checks that are switched off are still required.

pbrezina:
Yes, this is about https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule

PR #100, opened 2024-04-22
roles: keycloak role user mgt and kcadm updates

spoore1:
Adding new methods to KeycloakUser class to extend user management support:  delete, modify, set_password

Adding raise_on_error parameter to KeycloakUser methods so that they can be called for expected failure scenarios.  This required adding raise_on_error to kcadm exec call.


sssd_test_framework/roles/keycloak.py
```diff
@@ -126,12 +129,66 @@ def add(
...
+        self,
+        password: str | None = "Secret123",
+    ) -> KeycloakUser:
+        """
+        Set Keycloak user password.
+
+        :param password: Password, defaults to 'Secret123'
+        :type password: str | None, optional
+        :return: Self,
+        :rtype: KeycloakUser
```
danlavu, line 187: remove the comma please

sssd_test_framework/roles/keycloak.py
```diff
@@ -126,12 +129,66 @@ def add(
...
+        *,
+        firstName: str | None = None,
+        lastName: str | None = None,
+        email: str | None = None,
+        enabled: bool | None = None,
+    ) -> KeycloakUser:
+        """
+        Modify keycloak user attributes.
+
+        Parameters that are not set are ignored.
```
danlavu, line 156: I think this is enough, like the rest of the code, that it doesn't need to be said.

sssd_test_framework/roles/keycloak.py
```diff
@@ -126,12 +129,66 @@ def add(
...
+    ) -> KeycloakUser:
+        """
+        Set Keycloak user password.
+
+        :param password: Password, defaults to 'Secret123'
+        :type password: str | None, optional
+        :return: Self,
+        :rtype: KeycloakUser
+        """
+        set_password = "set-password -r master " f"--username {self.name} " f"--new-password {password}"
```
danlavu, line 189: f"set-password -r master --username {self.name} --new-password {password}" 

instead?

sssd_test_framework/roles/keycloak.py
```diff
@@ -126,12 +129,66 @@ def add(
...
+        :return: SSH exec result
+        :rtype: SSHProcessResult
+        """
+
+        attrs = {"firstName": firstName, "lastName": lastName, "email": email, "enabled": enabled}
+
+        update_attrs = " ".join([f'-s "{k}={v}"' for k, v in attrs.items() if v is not None])
+        update_user = f"update users/{self.id} {update_attrs}"
+        self.role.kcadm(update_user)
+        return self
```
danlavu, line 175: empty line before the return

sssd_test_framework/roles/keycloak.py
```diff
@@ -126,12 +129,66 @@ def add(
...
+        :param lastName: User's last name
+        :type lastName: str | None, optional
+        :param email: User's email address
+        :type email: str | None, optional
+        :param enabled: User account enabled/disabled status
+        :type enabled: bool | None, optional
+        :return: SSH exec result
+        :rtype: SSHProcessResult
+        """
+
```
danlavu, line 169: No empty line after the comments, sorry to be nitpicky but just following what most of our code looks like.

sssd_test_framework/roles/keycloak.py
```diff
@@ -126,12 +129,66 @@ def add(
...
         """
         Delete Keycloak user.
+
+        :return: SSH exec result
+        :rtype: SSHProcessResult
         """
         del_user = f"delete users/{self.id}"
-        self.role.kcadm(del_user)
+        result = self.role.kcadm(del_user)
+        return result
```
danlavu, line 143: empty line before the return

danlavu (changes requested):

sssd_test_framework/roles/keycloak.py
spoore1: added

sssd_test_framework/roles/keycloak.py
spoore1: added

sssd_test_framework/roles/keycloak.py
spoore1: removed

sssd_test_framework/roles/keycloak.py
spoore1: removed

sssd_test_framework/roles/keycloak.py
spoore1: yes, I think this must have be a reformat mix-up on my part.  I believe I originally had this spread across multiple lines and one of the linter/format tools combined on one line.  I changed it to reflect your suggestion.

sssd_test_framework/roles/keycloak.py
spoore1: removed

danlavu (dismissed):
LG2M

sssd_test_framework/roles/keycloak.py
```diff
@@ -152,6 +209,8 @@ def get(self) -> dict[str, list[str]]:
             out.setdefault(key, [])
             out[key].append(json1[key])
 
+        self.id = out["id"][0]
```
pbrezina, line 212: Please, don't create fields outside of the constructor, it is a bad habit. If you do this, you never know what properties are available (because they can be set anywhere) and you can't document them. Instead, add this line to the constructor.

sssd_test_framework/roles/keycloak.py
```diff
@@ -126,12 +129,66 @@ def add(
 
         return self
 
-    def delete(self) -> None:
+    def delete(
+        self,
+    ) -> SSHProcessResult:
```
pbrezina, line 134: What is the intention behind this?

pbrezina (changes requested):
Hi, see comments inside.

By the way, you say you add raise_on_error parameter, but I don't see it anywhere. Did you change your mind?

sssd_test_framework/roles/keycloak.py
danlavu: Yes, it looked strange and I proposed returning a tuple like password authentication, and then Scott said he didn't need it after all.

sssd_test_framework/roles/keycloak.py
spoore1: I think this was left over from an earlier update to return the result to check the error returned.   I have since changed to use this:

```
    with pytest.raises(SSHProcessError, match="User couldn't be deleted"):
        kcuser1.delete()
```

I left this to return the result if needed later but, TBH I can't think of a reason now.   The kcadm delete command typically returns nothing except on failure which I can check like this.

Is checking the raised error like above a good way to accomplish this?   So I can change this back to returning None.  Or would I be better off returning the SSHProcessResult and processing it that way?

pbrezina: It really depends on what you want to accomplish this.

Current implementation of `kcadm` raises error on `rc != 0`.

* If you want to assert that a command failed, then you don't need to return the result because kcadm will raise `SSHProccessError` anyway.
* If you want to check something in the result of successful run then you need to return the result.

Another way would be to set `raise_on_error=False` and return the result but then you always have to assert the return code by yourself (`assert result.rc == 0`).

From what you describe, I'd switch back to return `None` and rely on the exception. With that, `pytest.raises` is a good way to assert exceptions.

spoore1: ok, reverting this back to the original returning None.

sssd_test_framework/roles/keycloak.py
spoore1: @danlavu I think I was confused over the point you were making earlier..

@pbrezina isn't self.id inherited from the KeycloakObject parent class?

pbrezina: > @danlavu I think I was confused over the point you were making earlier..
> 
> @pbrezina isn't self.id inherited from the KeycloakObject parent class?

Ah, yes, you are right. It wasn't clear from the diff. In that case, you still should put it to the User class because it is not used anywhere else, right?

spoore1: Several Keycloak commands require ID so it's also used in the KeycloakGroup class.  Should I move it from KeycloakObject into each class where it's needed?   Right now KeycloakObject only defines `name` and `id`.

pbrezina: Ok, I went through the whole file and see what you are doing now. Keep it as is, but maybe make it little bit safer to use to avoid hunting potential bugs:

```
class KeycloakObject(BaseObject[KeycloakHost, Keycloak]):
    """
    Keycloak Object.
    """

    def __init__(self, role: Keycloak, name: str) -> None:
        """
        :param role: Keycloak role object.
        :type role: Keycloak
        :param name: User name.
        :type name: str
        """
        super().__init__(role)
        self.name = name

        # This is set as a side effect of certain operations like add or get.
        self._id: str | None = None

    @property
    def id(self) -> str:
        """
        Keycloak object ID.

        :raises ValueError: If ID is not set.
        :return: Keycloak object ID.
        :rtype: str
        """
        if self._id is None:
            raise ValueError("Keycloak object ID is unset.")

        return self._id

    @id.setter
    def id(self, value: str):
        self._id = value
```

sssd_test_framework/roles/ad.py
```diff
@@ -800,6 +806,10 @@ def modify(
         :type shell: str | DeleteAttribute | None, optional
         :param email: Email address, defaults to None
         :type email: str | DeleteAttribute | None, optional
+        :param given_name: Given name of user, defaults to None
```
danlavu, line 809: Variable names look strange. I'd prefer 

givenname instead given_name and surname instead of sur_name

It's just because they point to an LDAP attribute. It's a small detail; reconsider it. Otherwise, it looks good. Thanks!

spoore1: ok.  fixed.  Not sure why I added the underscore in the names there.

sssd_test_framework/roles/keycloak.py
spoore1: Forgot to mention earlier that I made the suggested change.

danlavu (dismissed):

pbrezina (changes requested):
Would it be possible to implement get() also for groups and unify the id property for both objects? I.e.:

```python
@property
def id(self) -> str:
    if self._id is None:
        obj = self.get()
        self._id = obj["id"]

    return self._id

@id.setter
def id(self, value: str) --> None:
    self._id = value
```
This way, you can avoid calling get() from the user's constructor when it is not needed for new users and both classes would handle it the same way.

sssd_test_framework/roles/keycloak.py
```diff
@@ -39,6 +39,8 @@ def kcadm(self, command: str) -> ProcessResult:
 
         :param command: kcadm command
         :type command: str
+        :return: SSH exec result
+        :rtype: SSHProcessResult
```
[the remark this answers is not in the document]
spoore1: updated

sssd_test_framework/roles/keycloak.py
```diff
@@ -39,6 +39,8 @@ def kcadm(self, command: str) -> ProcessResult:
 
         :param command: kcadm command
         :type command: str
+        :return: SSH exec result
```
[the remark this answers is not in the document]
spoore1: updated

spoore1:
@pbrezina I updated the KeycloakGroup to add a get() method and modified the id setting.   Let me know what you think.

pbrezina (changes requested):
Great, now we need one more thing - since you call get() from KeycloackObject, you need to add it as abstract method there. E.g.

```python
from abc import ABC, abstractmethod

class KeycloakObject(ABC, BaseObject[KeycloakHost, Keycloak]):
    @abstractmethod
    def get(self) -> dict[str, list[str]]:
        """docstring"""
        pass
```

Remove the commented lines and we can merge.

spoore1:
@pbrezina, I added abstract get() method to KeycloakObject class and removed comments I found.   Anything else for this one?   I just pushed a fix for the tox error also.

pbrezina (approved):
Ack, thank you.

sssd_test_framework/roles/ad.py
```diff
@@ -833,6 +847,65 @@ def modify(
...
+        return self
+
+    def reset(
+        self,
+        password: str = "Secret123",
+    ) -> ADUser:
+        """
+        Reset user password
+
+        Parameters that are not set are ignored.
```
danlavu, line 865: Redundant, the only parameter is password with a default value.

sssd_test_framework/hosts/ldap.py
```diff
@@ -35,6 +35,14 @@ def __init__(self, *args, **kwargs) -> None:
         self.client.setdefault("id_provider", "ldap")
         self.client.setdefault("ldap_uri", f"ldap://{self.hostname}")
 
+        # Use different default for domain
```
danlavu, line 38: Nitpick: Most of the comments have a period at the end. It could be more consistent, just going with what Pavel set.

sssd_test_framework/roles/ad.py
```diff
@@ -833,6 +847,65 @@ def modify(
         }
 
         self._modify(attrs)
+
+        # Password changes require a different command group so special
+        # handling is needed
```
danlavu, line 852: .

sssd_test_framework/roles/ad.py
```diff
@@ -833,6 +847,65 @@ def modify(
...
+            self.reset(str(password))
+
+        return self
+
+    def reset(
+        self,
+        password: str = "Secret123",
+    ) -> ADUser:
+        """
+        Reset user password
```
danlavu, line 863: .

sssd_test_framework/roles/ad.py
```diff
@@ -833,6 +847,65 @@ def modify(
...
+            Import-Module ActiveDirectory
+            Set-ADAccountPassword {args}
+            """
+        )
+
+        return self
+
+    def expire(self, expiration: str = "19700101000000") -> ADUser:
+        """
+        Set user password expiration date and time
```
danlavu, line 890: .

sssd_test_framework/roles/generic.py
```diff
@@ -508,6 +508,30 @@ def modify(
         """
         pass
 
+    @abstractmethod
+    def reset(self, password: str | None = "Secret123") -> GenericUser:
+        """
+        Reset user password
```
danlavu, line 514: .

sssd_test_framework/roles/generic.py
```diff
@@ -508,6 +508,30 @@ def modify(
...
+        :type password: str, optional
+        :return: Self.
+        :rtype: IPAUser
+        """
+        pass
+
+    @abstractmethod
+    def expire(self, expiration: str | None = "19700101000000") -> GenericUser:
+        """
+        Set user password expiration date and time
```
danlavu, line 526: .

sssd_test_framework/roles/ipa.py
```diff
@@ -455,19 +455,28 @@ def add(
...
         shell: str | None = None,
         user_auth_type: str | list[str] | None = None,
+        idp: str | None = None,
+        idp_user_id: str | None = None,
+        password_expiration: str | None = None,
     ) -> IPAUser:
         """
         Modify existing IPA user.
 
         Parameters that are not set are ignored.
```
danlavu, line 474: Redundant means a little more in the Generic class, I think you should remove this comment.

sssd_test_framework/roles/ipa.py
```diff
@@ -482,22 +491,61 @@ def modify(
...
+        """
+        pwinput = f"{password}\n{password}"
+        self.role.host.conn.run(f"ipa passwd {self.name}", input=pwinput)
+        self.expire("20380101120000Z")
+
+        return self
+
+    def expire(self, expiration: str | None = "19700101000000Z") -> IPAUser:
+        """
+        Set user password expiration date and time
```
danlavu, line 538: .

sssd_test_framework/roles/ipa.py
```diff
@@ -482,22 +491,61 @@ def modify(
...
+            "idp-user-id": (self.cli.option.VALUE, idp_user_id),
+            "password-expiration": (self.cli.option.VALUE, password_expiration),
         }
 
         self._modify(attrs, input=password)
         return self
 
+    def reset(self, password: str | None = "Secret123") -> IPAUser:
+        """
+        Reset user password
```
danlavu, line 523: .

sssd_test_framework/roles/ipa.py
```diff
@@ -455,19 +455,28 @@ def add(
...
+        idp: str | None = None,
+        idp_user_id: str | None = None,
+        password_expiration: str | None = None,
     ) -> IPAUser:
         """
         Modify existing IPA user.
 
         Parameters that are not set are ignored.
 
+        :param first: First name of user
```
danlavu, line 476: .

Seems like if there is not no comma, a period at the end, with a comma, period is dropped. 

:param first:  First name.
:param first: First name, defaults to None
:type or rtype: no periods

sssd_test_framework/roles/keycloak.py
```diff
@@ -208,6 +292,27 @@ def add(self) -> KeycloakGroup:
         self.id = result.stderr.split()[-1].strip("'")
         return self
 
+    def get(self) -> dict[str, list[str]]:
+        """
+        Get Keycloak group details.
+
+        :return: Dict of group info
```
danlavu, line 299: .

danlavu (changes requested):

sssd_test_framework/roles/keycloak.py
spoore1: updating

sssd_test_framework/roles/ipa.py
spoore1: removing

sssd_test_framework/roles/ipa.py
spoore1: updating

sssd_test_framework/roles/ad.py
```diff
@@ -833,6 +847,65 @@ def modify(
         }
 
         self._modify(attrs)
+
+        # Password changes require a different command group so special
```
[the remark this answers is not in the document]
spoore1: consolidating lines

sssd_test_framework/roles/ad.py
spoore1: adding

sssd_test_framework/roles/ad.py
spoore1: removing

sssd_test_framework/roles/ad.py
spoore1: adding

sssd_test_framework/roles/ad.py
```diff
@@ -833,6 +847,65 @@ def modify(
...
+        """
+        expire = datetime.strptime(expiration, "%Y%m%d%H%M%S")
+        expire_format = expire.strftime("%m/%d/%Y %H:%M:%S")
+
+        attrs: CLIBuilderArgs = {**self._identity, "DateTime": (self.cli.option.VALUE, f"{expire_format}")}
+
+        args = " ".join(self.cli.args(attrs, quote_value=True))
+        self.role.host.conn.run(
+            f"""
+            Import-Module ActiveDirectory
```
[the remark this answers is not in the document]
spoore1: squashing

sssd_test_framework/roles/generic.py
spoore1: adding

sssd_test_framework/roles/generic.py
spoore1: adding

sssd_test_framework/roles/ipa.py
spoore1: adding

sssd_test_framework/roles/ipa.py
spoore1: adding

danlavu (approved):
Looks good to me, thanks!

PR #103, opened 2024-05-06
topology: provision topologies if not already using per-topology setup

pbrezina:
Currently, we provide machines that are already fully setup with the
    client enrolled into all domains and trusts established. This is no
    longer necessary. With this patch, we will enroll into the domains
    using per-topology setup if needed.

    It adds also provisioned_topologies field to mhc.yaml to avoid
    additional provisioning attempt if the topology is already fully setup.

---

Please note that this needs to be closed together with changes to other repositories:
* https://github.com/SSSD/sssd/pull/7349
* https://github.com/SSSD/sssd-ci-containers/pull/95
* https://github.com/next-actions/pytest-mh/pull/54

### DIFF
--- a/sssd_test_framework/hosts/ldap.py
+++ b/sssd_test_framework/hosts/ldap.py
@@ -7,14 +7,14 @@ from typing import Any
 import ldap
 from pytest_mh.ssh import SSHLog
 
-from .base import BaseLDAPDomainHost
+from .base import BaseLDAPDomainHost, BaseLinuxHost
 
 __all__ = [
     "LDAPHost",
 ]
 
 
-class LDAPHost(BaseLDAPDomainHost):
+class LDAPHost(BaseLDAPDomainHost, BaseLinuxHost):
     """
     LDAP host object.
 
--- a/sssd_test_framework/hosts/ldap.py
+++ b/sssd_test_framework/hosts/ldap.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from typing import Any
+
 import ldap
 from pytest_mh.ssh import SSHLog
 
@@ -37,13 +39,6 @@ class LDAPHost(BaseLDAPDomainHost):
         self.client.setdefault("id_provider", "ldap")
         self.client.setdefault("ldap_uri", f"ldap://{self.hostname}")
 
-        # Backup of original data
-        self.__backup: dict[str, dict[str, list[bytes]]] = {}
-
-    def pytest_setup(self) -> None:
-        # Start ldap before properties are enumerated in MultihostUtility
-        self._start()
-
     def setup(self) -> None:
         # Make sure ldap is running for each test
         super().setup()
@@ -78,13 +73,26 @@ class LDAPHost(BaseLDAPDomainHost):
 
         return self._features
 
-    def backup(self) -> None:
+    def remove_backup(self, backup_data: Any | None) -> None:
+        """
+        Remove backup data from the host.
+
+        :param backup_data: Backup data.
+        :type backup_data: Any | None
+        """
+        # Nothing to do since we store backup in memory
+        pass
+
+    def backup(self) -> Any:
         """
         Backup all directory server data.
 
         Full backup of ``cn=config`` and default naming context is performed.
         This is done by simple LDAP search on given base dn and remembering the
         contents. The operation is usually very fast.
+
+        :return: Backup data.
+        :rtype: Any
         """
         self._start()
         data = self.conn.search_s(self.naming_context, ldap.SCOPE_SUBTREE)
@@ -94,9 +102,10 @@ class LDAPHost(BaseLDAPDomainHost):
         dct = self.ldap_result_to_dict(data)
         dct.update(self.ldap_result_to_dict(config))
         dct.update(self.ldap_result_to_dict(nc))
-        self.__backup = dct
 
-    def restore(self) -> None:
+        return dct
+
+    def restore(self, backup_data: Any | None) -> None:
         """
         Restore directory server data.
 
@@ -105,7 +114,16 @@ class LDAPHost(BaseLDAPDomainHost):
         difference between original data obtained by :func:`backup` and then
         calling add, delete and modify operations to convert current state to
         the original state. This operation is usually very fast.
+
+        :return: Backup data.
+        :rtype: Any
         """
+        if backup_data is None:
+            return
+
+        if not isinstance(backup_data, dict):
+            raise TypeError(f"Expected dict, got {type(backup_data)}")
+
         self._start()
         data = self.conn.search_s(self.naming_context, ldap.SCOPE_SUBTREE)
         config = self.conn.search_s("cn=config", ldap.SCOPE_BASE)
@@ -118,8 +136,8 @@ class LDAPHost(BaseLDAPDomainHost):
 
         for dn, attrs in reversed(data.items()):
             # Restore records that were modified
-            if dn in self.__backup:
-                original_attrs = self.__backup[dn]
+            if dn in backup_data:
+                original_attrs = backup_data[dn]
                 modlist = ldap.modlist.modifyModlist(attrs, original_attrs)
                 modlist = self.__filter_modlist(dn, modlist)
                 if modlist:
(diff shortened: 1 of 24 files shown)


sssd_test_framework/misc/__init__.py
```diff
@@ -153,3 +155,49 @@ def attrs_to_hash(attrs: dict[str, Any]) -> str | None:
         return None
 
     return "@{" + out.rstrip(";") + "}"
+
```
danlavu, line 158: This is nice, this is very nice.

sssd_test_framework/misc/ssh.py
```diff
@@ -92,3 +94,74 @@ def __enter__(self) -> SSHKillableProcess:
     def __exit__(self, exception_type, exception_value, traceback) -> None:
         self.kill()
         self.process.wait()
+
+
+Param = ParamSpec("Param")
+
+
+def retry_command(
```
danlavu, line 102: How is this going to be used? with retry() as a decorator we can do something like 

```
@retry()
def test_run_this_5x():
```

Some examples in the doc string for retry and retry_command will be nice.

sssd_test_framework/utils/pam.py
```diff
@@ -169,63 +142,57 @@ def test_example(client: Client, provider: GenericProvider):
...
-
-            client.sssd.common.pam(["with-faillock"])
+            with mh_utility(PAMFaillockUtils(client.host, client.fs)) as faillock:
+                # Setup faillock
+                faillock.config_set({"deny": "3", "unlock_time": "300"})
+                client.sssd.common.pam(["with-faillock"])
 
-            # Start SSSD
-            client.sssd.start()
+                # Start SSSD
```
danlavu, line 150: replace tabs with spaces

sssd_test_framework/topology_controllers.py
```diff
@@ -79,10 +202,42 @@ class IPATrustADTopologyController(BackupTopologyController):
...
+
+        # Create trust
+        self.logger.info(f"Establishing trust between {ipa.domain} and {trusted.domain}")
+        ipa.kinit()
+        ipa.ssh.exec(
+            ["ipa", "trust-add", trusted.domain, "--admin", "Administrator", "--password"],
+            input=trusted.adminpw,
+        )
+
+        # Do not enroll client into IPA domain if it is already joined
```
justin-stephenson, line 218: Why is this here and not for AD join?

sssd_test_framework/topology_controllers.py
```diff
@@ -14,59 +23,173 @@
...
+        if self.provisioned:
+            self.logger.info(f"Topology '{self.name}' is already provisioned")
+            return
 
+        self.logger.info(f"Enrolling {client.hostname} into {provider.domain}")
 
-class SambaTopologyController(BackupTopologyController):
+        # Remove any existing Kerberos configuration and keytab
+        client.fs.rm("/etc/krb5.conf")
+        client.fs.rm("/etc/krb5.keytab")
```
justin-stephenson, line 183: I'm a bit unclear why this is needed if realm join updates these files.

sssd_test_framework/hosts/ldap.py
```diff
@@ -117,6 +119,8 @@ def restore(self, backup_data: Any | None) -> None:
         if backup_data is None:
             return
 
+        self.logger.info("Restoring LDAP server from memory")
```
justin-stephenson, line 122: from memory? I don't quite understand this part.

sssd_test_framework/utils/pam.py
```diff
@@ -8,48 +8,12 @@
 from pytest_mh.utils.fs import LinuxFileSystem
 
 __all__ = [
-    "PAMUtils",
-    "PAMAccess",
-    "PAMFaillock",
+    "PAMAccessUtils",
```
danlavu, line 11: Why rename the classes?

sssd_test_framework/utils/pam.py
```diff
@@ -169,63 +142,57 @@ def test_example(client: Client, provider: GenericProvider):
...
-            assert client.auth.ssh.password("user-1", "Secret123")
+                # Check the results
+                assert client.auth.ssh.password("user-1", "Secret123")
 
-            # Three failed login attempts
-            for i in range(3):
-                assert not client.auth.ssh.password("user-1", "bad_password")
+                # Three failed login attempts
+                for i in range(3):
+                    assert not client.auth.ssh.password("user-1", "bad_password")
```
danlavu, line 158: You have tabs from line 150 to 158

danlavu:
I've gotten all the topologies working except for AD

```
           systemctl reset-failed "sssd"; systemctl start "sssd" || systemctl status "sssd"
E         CWD:
E         Env:
E         Output:
E           × sssd.service - System Security Services Daemon
E                Loaded: loaded (/usr/lib/systemd/system/sssd.service; enabled; preset: enabled)
E               Drop-In: /usr/lib/systemd/system/service.d
E                        └─10-timeout-abort.conf
E                Active: failed (Result: exit-code) since Fri 2024-05-17 03:02:34 UTC; 5ms ago
E              Duration: 13.257s
E               Process: 49058 ExecStart=/usr/sbin/sssd -i ${DEBUG_LOGGER} (code=exited, status=1/FAILURE)
E              Main PID: 49058 (code=exited, status=1/FAILURE)
E                   CPU: 198ms
E       
E           May 17 03:02:34 client.test sssd_be[49070]: krb5_kt_start_seq_get failed: Key table file '/etc/krb5.keytab' not found
E           May 17 03:02:34 client.test sssd_be[49070]: krb5_kt_start_seq_get failed: Key table file '/etc/krb5.keytab' not found
E           May 17 03:02:34 client.test sssd_be[49070]: krb5_kt_start_seq_get failed: Key table file '/etc/krb5.keytab' not found
E           May 17 03:02:34 client.test sssd_be[49070]: krb5_kt_start_seq_get failed: Key table file '/etc/krb5.keytab' not found
E           May 17 03:02:34 client.test sssd_be[49070]: krb5_kt_start_seq_get failed: Key table file '/etc/krb5.keytab' not found
E           May 17 03:02:34 client.test sssd_be[49070]: Failed to read keytab [FILE:/etc/krb5.keytab]: No suitable principal found in keytab
E           May 17 03:02:34 client.test sssd[49058]: Exiting the SSSD. Could not restart critical service [test].
E           May 17 03:02:34 client.test systemd[1]: sssd.service: Main process exited, code=exited, status=1/FAILURE
E           May 17 03:02:34 client.test systemd[1]: sssd.service: Failed with result 'exit-code'.
E           May 17 03:02:34 client.test systemd[1]: Failed to start sssd.service - System Security Services Daemon.
E         Error output:
E           Job for sssd.service failed because the control process exited with error code.
E           See "systemctl status sssd.service" and "journalctl -xeu sssd.service" for details.
```

sssd_test_framework/topology_controllers.py
pbrezina: Can you clarify this question?

sssd_test_framework/topology_controllers.py
pbrezina: There may be already some principal stored in the keytab or some existing kerberos configuration. In theory, we could just backup these files so we can revert any changes, but imho removing them is better so we always start with a clean state.

sssd_test_framework/hosts/ldap.py
pbrezina: LDAP backup is stored into memory - into python variable. Other hosts are storing it into a file.

sssd_test_framework/utils/pam.py
pbrezina: These classes were written incorrectly and setup and teardown of PAMAccess and PAMFaillock would not be called. So I fix that, then renamed the classes to follow our naming convention.

danlavu (approved):

sssd_test_framework/topology_controllers.py
justin-stephenson: In `IPATrustADTopologyController` topology_setup method you have:

```
        # Do not enroll client into IPA domain if it is already joined
        if "ipa" not in self.multihost.provisioned_topologies:

```

But this same check does not exist in `ADTopologyController`, where both execute 'realm join'

pbrezina: This is different. This handles the case where the client is already joined into an IPA domain (from ansible), but trust between IPA and AD is not yet created. Which is the default setup on containers. So in this case, you only want to create the trust between IPA and AD but client is already enrolled into IPA domain.

In ADTopologyController, you just want to joint the client into AD.

justin-stephenson: I understand now, thanks for clarification.

justin-stephenson (approved):
Ack, thank you for your patience.

PR #113, opened 2024-07-11
use generic connection interface from latest pytest-mh

pbrezina:
Latest version added an option to replace SSH connections with podman
or docker, therefore a generic interface was created. Most notably,
`host.ssh` was replaced with `host.conn`.

Depends on:
* https://github.com/next-actions/pytest-mh/pull/67


justin-stephenson (dismissed):
Ack, thank you.

pbrezina:
Just rebased.

PR #114, opened 2024-07-15
hosts: Add distro information to BaseLinuxHost

jakub-vavra-cz:
We need distro information in order to match expected features to distros and versions.

### DIFF
--- a/sssd_test_framework/hosts/base.py
+++ b/sssd_test_framework/hosts/base.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import csv
 from typing import Any
 
 import ldap
@@ -201,3 +202,65 @@ class BaseLinuxHost(MultihostHost[SSSDMultihostDomain]):
 
         self.fs: LinuxFileSystem = LinuxFileSystem(self)
         self.svc: SystemdServices = SystemdServices(self)
+        self._os_release: dict = {}
+        self._distro_name: str = "unknown"
+        self._distro_major: int = 0
+        self._distro_minor: int = 0
+
+    def _distro_information(self):
+        """
+        Pulls distro information from a host from /ets/os-release
+        """
+        self.logger.info(f"Detecting distro information on {self.hostname}")
+        os_release = self.fs.read("/etc/os-release")
+        self._os_release = dict(csv.reader([x for x in os_release.splitlines() if x], delimiter="="))
+        if "NAME" in self._os_release:
+            self._distro_name = self._os_release["NAME"]
+        if "VERSION_ID" not in self._os_release:
+            return
+        if "." in self._os_release["VERSION_ID"]:
+            self._distro_major = int(self._os_release["VERSION_ID"].split(".", maxsplit=1)[0])
+            self._distro_minor = int(self._os_release["VERSION_ID"].split(".", maxsplit=1)[1])
+        else:
+            self._distro_major = int(self._os_release["VERSION_ID"])
+
+    @property
+    def distro_name(self) -> str:
+        """
+        Host distribution
+
+        :return: Distribution name or "unknown"
+        :rtype: str
+        """
+        # NAME item from os-release
+        if not self._os_release:
+            self._distro_information()
+        return self._distro_name
+
+    @property
+    def distro_major(self) -> int:
+        """
+        Host distribution major version
+
+        :return: Major version
+        :rtype: int
+        """
+        # First part of VERSION_ID from os-release
+        # Returns zero when could not detect
+        if not self._os_release:
+            self._distro_information()
+        return self._distro_major
+
+    @property
+    def distro_minor(self) -> int:
+        """
+        Host distribution minor version
+
+        :return: Minor version
+        :rtype: int
+        """
+        # Second part of VERSION_ID from os-release
+        # Returns zero when no minor version is present
+        if not self._os_release:
+            self._distro_information()
+        return self._distro_minor


justin-stephenson:
Would it be better to use https://github.com/python-distro/distro instead?

sssd_test_framework/hosts/base.py
```diff
@@ -296,3 +296,64 @@ def __init__(self, *args, **kwargs) -> None:
...
+        self._distro_name: str | None = None
+        self._distro_major: int | None = None
+        self._distro_minor: int | None = None
+
+    def _distro_information(self):
+        """
+        Pulls distro information from a host from /ets/os-release
+        """
+        self.logger.info(f"Detecting distro information on {self.hostname}")
+        os_release = self.ssh.run("cat /etc/os_release", log_level=SSHLog.Error).stdout_lines
```
justin-stephenson, line 309: `/etc/os-release` not `/etc/os_release` right?

jakub-vavra-cz: fixed.

alexey-tikhonov:
Is it expected that 'CI / tox' fail all targets with "docs-upstream: FAIL"?

jakub-vavra-cz:
There is supposedly issue with docs/sphinx.
https://github.com/sphinx-doc/sphinx/issues/12589

sssd_test_framework/hosts/base.py
jakub-vavra-cz: I did not propagate local changes from debugging. :-)

alexey-tikhonov:
> Would it be better to use https://github.com/python-distro/distro instead?

@jakub-vavra-cz, what about this ^^ ?

jakub-vavra-cz:
> > Would it be better to use https://github.com/python-distro/distro instead?
> 
> @jakub-vavra-cz, what about this ^^ ?

@alexey-tikhonov 
The library/cli reports about local machine. 
I do not want to install this as additional dependency on every machine/container. 
I could probably download the release files from the machine, store them somewhere and make the library to parse them but it seems more involved and the code would be probably bigger.

sssd_test_framework/hosts/base.py
```diff
@@ -296,3 +296,66 @@ def __init__(self, *args, **kwargs) -> None:
...
+            self._distro_information()
+        return self._distro_name
+
+    @property
+    def distro_major(self) -> int | None:
+        """
+        Host distribution major version
+
+        :return: Major version
+        :rtype: int | None
```
alexey-tikhonov, line 342: I don't think it can return 'None'

sssd_test_framework/hosts/base.py
```diff
@@ -296,3 +296,66 @@ def __init__(self, *args, **kwargs) -> None:
...
+            self._distro_information()
+        return self._distro_major
+
+    @property
+    def distro_minor(self) -> int | None:
+        """
+        Host distribution minor version
+
+        :return: Minor version
+        :rtype: int | None
```
alexey-tikhonov, line 355: I don't think it can return 'None'

justin-stephenson:
Pavel can provide his input on this when he is back but as this OS information is more generic and not SSSD specific, it should more likely be part of pytest-mh. I think it could be added as a new class  in https://github.com/next-actions/pytest-mh/tree/master/pytest_mh/utils

sssd_test_framework/hosts/base.py
jakub-vavra-cz: You are right I wanted to keep it None when the version could not be retrieved.

sssd_test_framework/hosts/base.py
jakub-vavra-cz: You are right I wanted to keep it None when the version could not be retrieved.

pbrezina:
> Pavel can provide his input on this when he is back but as this OS information is more generic and not SSSD specific, it should more likely be part of pytest-mh. I think it could be added as a new class in https://github.com/next-actions/pytest-mh/tree/master/pytest_mh/utils

Hi, this sounds like a good idea.

sssd_test_framework/hosts/base.py
```diff
@@ -296,3 +296,69 @@ def __init__(self, *args, **kwargs) -> None:
...
+            return
+        if "." in self._os_release["VERSION_ID"]:
+            self._distro_major = int(self._os_release["VERSION_ID"].split(".", maxsplit=1)[0])
+            self._distro_minor = int(self._os_release["VERSION_ID"].split(".", maxsplit=1)[1])
+        else:
+            self._distro_major = int(self._os_release["VERSION_ID"])
+            self._distro_minor = 0
+
+    @property
+    def distro_name(self) -> str | None:
```
pbrezina, line 327: Do you really want to return None as a possible value? I think it would be better to keep them always non-None with default value (name=unknown, version=0.0), otherwise you will have to assert that it is not None before every use.

sssd_test_framework/hosts/base.py
```diff
@@ -296,3 +296,69 @@ def __init__(self, *args, **kwargs) -> None:
 
         self.fs: LinuxFileSystem = LinuxFileSystem(self)
         self.svc: SystemdServices = SystemdServices(self)
+        self._os_release: dict = {}
+        self._distro_name: str | None = None
+        self._distro_major: int | None = None
+        self._distro_minor: int | None = None
```
pbrezina, line 302: It might be nicer to initialize here to some default values and move the call to `_distro_information` into `pytest_setup` and remove the `@property`ies. It would simplify the code. But I don't insist, whatever you prefer.

sssd_test_framework/hosts/base.py
```diff
@@ -296,3 +296,69 @@ def __init__(self, *args, **kwargs) -> None:
...
+
+    def _distro_information(self):
+        """
+        Pulls distro information from a host from /ets/os-release
+        """
+        self.logger.info(f"Detecting distro information on {self.hostname}")
+        os_release = self.ssh.run("cat /etc/os-release", log_level=SSHLog.Error).stdout_lines
```
pbrezina, line 309: `self.fs.read("/etc/os-release")`

sssd_test_framework/hosts/base.py
```diff
@@ -296,3 +296,69 @@ def __init__(self, *args, **kwargs) -> None:
...
+        for line in os_release:
+            if "=" not in line:
+                continue
+            k, v = line.rstrip().split("=", maxsplit=1)
+            self._os_release[k] = v.strip('"')
```
pbrezina, line 314: `dict(csv.reader(os_release.splitlines(), delimiter="="))`

pbrezina (changes requested):

sssd_test_framework/hosts/base.py
pbrezina: This would be a more robust one liner. Just import the csv library that is built-in python.

jakub-vavra-cz: Ack, will do that.

sssd_test_framework/hosts/base.py
jakub-vavra-cz: done

sssd_test_framework/hosts/base.py
jakub-vavra-cz: Ack. Initially I wanted to use None when I did not try to detect distro yet but I can use self._os_release for that.

pbrezina (dismissed):

pbrezina:
Code-wise ack, I did not test it.

sssd_test_framework/hosts/base.py
pbrezina: It would be run only once, not for each test.

jakub-vavra-cz:
@pbrezina 
Just tested, the suggested csv oneliner fails me in comparison to my code that actually worked before:

```
>       self._os_release = dict(csv.reader(os_release.splitlines(), delimiter="="))
E       ValueError: dictionary update sequence element #15 has length 0; 2 is required
```
The file has an empty line:
```
[root@client ~]# cat /etc/os-release 
NAME="Red Hat Enterprise Linux"
VERSION="10.0 (Coughlan)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="10.0"
PLATFORM_ID="platform:el10"
PRETTY_NAME="Red Hat Enterprise Linux 10.0 Beta (Coughlan)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:10::baseos"
HOME_URL="https://www.redhat.com/"
VENDOR_NAME="Red Hat"
VENDOR_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/10"
BUG_REPORT_URL="https://issues.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 10"
REDHAT_BUGZILLA_PRODUCT_VERSION=10.0
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="10.0 Beta"
```

pbrezina (approved):
Code-wise ack. I did not test it, I'll leave it to other reviewers.

If you'd like, you can simplify it more by using `@cached_property` from `functools`, you could remove the private properties form the constructor and make distro_information a cached property.

pbrezina:
Why did you unassigned @justin-stephenson and @alexey-tikhonov?

pbrezina (approved):

PR #118, opened 2024-07-26
SSSD: use systemd service description to configure service user

alexey-tikhonov:
Usage example: https://github.com/SSSD/sssd/pull/7513

### DIFF
--- a/sssd_test_framework/hosts/client.py
+++ b/sssd_test_framework/hosts/client.py
@@ -29,6 +29,12 @@ class ClientHost(BaseBackupHost, BaseLinuxHost):
         super().__init__(*args, **kwargs)
 
         self._features: dict[str, bool] | None = None
+        self.sssd_service_user: str = ""
+        """ SSSD service user configured by default install """
+
+    def pytest_setup(self) -> None:
+        super().pytest_setup()
+        self.sssd_service_user = self.svc.get_property("sssd", "User")
 
     @property
     def features(self) -> dict[str, bool]:
--- a/sssd_test_framework/utils/sssd.py
+++ b/sssd_test_framework/utils/sssd.py
@@ -9,7 +9,8 @@ from typing import TYPE_CHECKING, Literal
 from pytest_mh import MultihostHost, MultihostRole, MultihostUtility
 from pytest_mh.conn import Process, ProcessLogLevel, ProcessResult
 
-from ..hosts.base import BaseDomainHost, BaseHost
+from ..hosts.base import BaseDomainHost
+from ..hosts.client import ClientHost
 from ..misc import to_list
 
 if TYPE_CHECKING:
@@ -113,6 +114,7 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
     def async_start(
         self,
         service="sssd",
+        service_user="sssd",
         *,
         apply_config: bool = True,
         check_config: bool = True,
@@ -132,6 +134,8 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         :return: Running SSH process.
         :rtype: Process
         """
+        self.set_service_user(service_user)
+
         if apply_config:
             self.config_apply(check_config=check_config, debug_level=debug_level)
 
@@ -144,6 +148,7 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
     def start(
         self,
         service="sssd",
+        service_user="sssd",
         *,
         raise_on_error: bool = True,
         apply_config: bool = True,
@@ -166,6 +171,8 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         :return: SSH process result.
         :rtype: ProcessResult
         """
+        self.set_service_user(service_user)
+
         if apply_config:
             self.config_apply(check_config=check_config, debug_level=debug_level)
 
@@ -303,19 +310,40 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
 
     def set_service_user(self, user: str) -> None:
         """
-        Set [sssd]/user option.
+        Reconfigures 'sssd.service' systemd service description
+        to run SSSD service under 'user' (only 'root' or 'sssd'
+        are supported by SSSD).
+        Take a note, this currently doesn't handle reconfiguration
+        of socket activated services.
 
         :param user: Option value to set.
         :type user: str
-        :raises ValueError: If required feature wasn't built.
+        :raises ValueError: in case error happens.
         """
-        if isinstance(self.host, BaseHost):
-            if (user != "root") and (not self.host.features["non-privileged"]):
-                raise ValueError("SSSD was built without support of running under non-root")
+        if isinstance(self.host, ClientHost):
+            if not self.host.features["non-privileged"]:
+                return  # service user configuration isn't supported at all
         else:
             raise ValueError("Unexpected host type")
 
-        self.sssd["user"] = user
+        if user == self.host.sssd_service_user:
+            return  # requested service user matches default, nothing to do
+
+        service_file = "/usr/lib/systemd/system/sssd.service"
+        self.fs.backup(service_file)
+        cmd = f'sed -i "s/^User=.*/User={user}/g" {service_file}\n'
+        cmd += f'sed -i "s/^Group=.*/Group={user}/g" {service_file}\n'
+        if user == "root":
+            cmd += f'sed -i "s/^#SupplementaryGroups=sssd$/SupplementaryGroups=sssd/g" {service_file}\n'
+            cmd += f'sed -i "s/sssd:sssd/root:root/g" {service_file}\n'
+        elif user == "sssd":
+            cmd += f'sed -i "s/^SupplementaryGroups=sssd$/#SupplementaryGroups=sssd/g" {service_file}\n'
+            cmd += f'sed -i "s/root:root/sssd:sssd/g" {service_file}\n'
+        else:
+            raise ValueError("Unexpected value of 'user'")
+        cmd += f"chown -f {user}:{user} /var/lib/sss/db/*.ldb || true"
+        self.host.conn.run(cmd)
+        self.svc.reload_daemon()
 
     def enable_responder(self, responder: str) -> None:
         """


sssd_test_framework/utils/sssd.py
```diff
@@ -113,6 +113,7 @@ def setup(self) -> None:
     def async_start(
         self,
         service="sssd",
+        service_user="sssd",
```
alexey-tikhonov, line 116: It feels weird that `service` can be configured to anything else but 'sssd' in 'SSSD' class.

sssd_test_framework/utils/sssd.py
```diff
@@ -132,6 +133,8 @@ def async_start(
         :return: Running SSH process.
         :rtype: SSHProcess
         """
+        self.set_service_user(service_user)
```
alexey-tikhonov, line 136: Maybe it should be moved inside `if service == "sssd":`, not sure...

sssd_test_framework/utils/sssd.py
pbrezina: With systemd, we have many services. Most of them are controlled by sssd.service, but e.g. sssd-kcm is not. The original intention was to allow to also control sssd-kcm. However, Alejandor added commit to start/stop sssd-kcm with sssd as well so maybe it is not needed anymore.

sssd_test_framework/utils/sssd.py
```diff
@@ -307,15 +313,26 @@ def set_service_user(self, user: str) -> None:
...
         if isinstance(self.host, BaseHost):
-            if (user != "root") and (not self.host.features["non-privileged"]):
-                raise ValueError("SSSD was built without support of running under non-root")
+            if not self.host.features["non-privileged"]:
+                return
         else:
             raise ValueError("Unexpected host type")
 
-        self.sssd["user"] = user
+        service_file = "/usr/lib/systemd/system/sssd.service"
```
pbrezina, line 324: Please, add also `self.fs.backup(service_file)` to make sure the file is restored after the test is finished.

pbrezina (changes requested):
What about socket activated services? If they are started via systemd and not the monitor, we need to also update other units as well, rigth? But I suppose we don't have tests for it yet, right?

alexey-tikhonov:
> What about socket activated services? If they are started via systemd and not the monitor, we need to also update other units as well, rigth? But I suppose we don't have tests for it yet, right?

I didn't touch it yet because so far we don't have any test for "socket activated responders" (but it's planned, CC @danlavu)

sssd_test_framework/utils/sssd.py
alexey-tikhonov: Done.

alexey-tikhonov:
Rebased.

sssd_test_framework/utils/sssd.py
```diff
@@ -307,15 +313,27 @@ def set_service_user(self, user: str) -> None:
 
         :param user: Option value to set.
```
justin-stephenson, line 314: Comment  above this needs to be changed.

`        Set [sssd]/user option.
`

sssd_test_framework/utils/sssd.py
```diff
@@ -132,6 +133,8 @@ def async_start(
         :return: Running SSH process.
         :rtype: Process
         """
+        self.set_service_user(service_user)
```
justin-stephenson, line 136: Can we skip calling `set_service_user` method when `service_user` argument (taken from default value, or passed as argument) == default service_user value 'sssd' ?

alexey-tikhonov: We could query current value (I wouldn't rely of "default").
Do you think it makes sense?

alexey-tikhonov: I mean, this is an ssh (or podman) command anyway.

justin-stephenson: > Do you think it makes sense?

Makes sense to me, I was trying to avoid calling `set_service_user` every single time these `*start` or `start` methods are run.

alexey-tikhonov: This check (if current user matches requested) would be within `set_service_user()` anyway.
And to perform this check, we need to execute ssh/podmand command on a host anyway.
So in best case it will be the same 1 command, in worst case it will be 2 commands (check + change) instead of 1.
It can also save backing up a file.
Does it really worth the trouble?

justin-stephenson: `start` method is used quite often so it seems worthwhile IMO especially the `self.host.conn.run("systemctl daemon-reload")` seems intensive.

```
justin@fedora:~/github/sssd/src/tests/system$ grep -irn 'sssd.start' tests/ | wc -l
190
```

@pbrezina Can we read the SSSD service user at host startup and check this value to conditionally run set_service_user() or do you prefer to keep the current PR state?

alexey-tikhonov: Ok, I'll add a check at the beginning of `set_service_user()` but I wouldn't rely on a state at "host startup" (what moment is it?)
What if host starts with sssd-2.9 and sssd-2.10 is installed later?
What is cleanup from previous test didn't work as expected and didn't restore "default" sssd.service (or is this out of scope)?

justin-stephenson: Maybe it makes sense to hold off on this check until all supported SSSD versions where test framework are running with the same default service user 'sssd', and we won't be switching back and forth often with

`@pytest.mark.parametrize("sssd_service_user", ("root", "sssd"))`

alexey-tikhonov: For sssd < 2.10 `set_service_user()` is a no-op because of:
```
            if not self.host.features["non-privileged"]:
                return
```

I didn't get "we won't be switching back and forth often". Could you please clarify?

The idea of `@pytest.mark.parametrize("sssd_service_user", ("root", "sssd"))` is to run the same test twice with SSSD running both under 'root' and 'sssd'. This should be small set of selected core tests.

justin-stephenson: My proposal was to:
- read the `User=$user` value from `/usr/lib/systemd/system/sssd.service` at the start of the pytest session, and will last the entire pytest session. I think this can be done adding functionality to https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/hosts/client.py but I need to test it.
- add conditional check, checking against the above value :
```
    if service_user == self.host.default_sssd_service_user:
                return
```
After each test run, the test restores the original `/usr/lib/systemd/system/sssd.service`.
>  What if host starts with sssd-2.9 and sssd-2.10 is installed later?

Does this happen now for some system tests? I didn't think so but i'm not sure.

> What if cleanup from previous test didn't work as expected and didn't restore "default" sssd.service (or is this out of scope)?

If cleanup doesn't work properly then test environment is tainted and all bets are off, we have seen this already in certain test failures in PRCI.

alexey-tikhonov: > > What if host starts with sssd-2.9 and sssd-2.10 is installed later?
> 
> Does this happen now for some system tests? I didn't think so but i'm not sure.

I was thinking of a case where Fedora comes with stock sssd-2.9 and CI installs sssd-2.10 on top.
But by the time of pytest session it should be already upgraded, so not an issue.
I wonder if we ever want to introduce a real "upgrade test" in the CI...

Ok, let's see if we can cache it.

pbrezina: > `start` method is used quite often so it seems worthwhile IMO especially the `self.host.conn.run("systemctl daemon-reload")` seems intensive.
> 
> ```
> justin@fedora:~/github/sssd/src/tests/system$ grep -irn 'sssd.start' tests/ | wc -l
> 190
> ```
> 
> @pbrezina Can we read the SSSD service user at host startup and check this value to conditionally run set_service_user() or do you prefer to keep the current PR state?

Yes, you can add it to `pytest_setup` method. This is executed only once and SSSD version under test is already installed.

alexey-tikhonov:
CI run after recent update:
 - F39:
```
tests/test_authentication.py::test_authentication__with_default_settings[root-su] (ipa) PASSED [ 14%]
tests/test_authentication.py::test_authentication__with_default_settings[root-ssh] (ipa) PASSED [ 15%]
tests/test_authentication.py::test_authentication__with_default_settings[sssd-su] (ipa) SKIPPED [ 15%]
tests/test_authentication.py::test_authentication__with_default_settings[sssd-ssh] (ipa) SKIPPED [ 15%]
```
 - F41:
```
tests/test_authentication.py::test_authentication__with_default_settings[root-su] (ipa) PASSED [ 14%]
tests/test_authentication.py::test_authentication__with_default_settings[root-ssh] (ipa) PASSED [ 15%]
tests/test_authentication.py::test_authentication__with_default_settings[sssd-su] (ipa) PASSED [ 15%]
tests/test_authentication.py::test_authentication__with_default_settings[sssd-ssh] (ipa) PASSED [ 15%]
...
tests/test_passkey.py::test_passkey__su_user (ipa) PASSED                [ 32%]
tests/test_passkey.py::test_passkey__su_user_with_failed_pin (ipa) PASSED [ 32%]
tests/test_passkey.py::test_passkey__su_user_with_incorrect_mapping (ipa) PASSED [ 32%]
...
```

sssd_test_framework/utils/sssd.py
alexey-tikhonov: @justin-stephenson, please take a look at the latest version.

alexey-tikhonov:
Hm, looks like the latest update have broken a bunch of system tests...

alexey-tikhonov:
> Hm, looks like the latest update have broken a bunch of system tests...

In all the broken tests it's either `ssh.password()` or `su.password()` what is failing.
In both cases it gets EOF instead of expected prompt.
SSSD seems to start correctly but there are no traces of related requests in SSSD logs.

@pbrezina, when a test is run multiple times due to parametrization like in:
```
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[root-su] (samba) PASSED [ 75%]
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[root-ssh] (samba) FAILED [ 76%]
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[sssd-su] (samba) FAILED [ 76%]
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[sssd-ssh] (samba) FAILED [ 76%]
```
  --  does teardown / restoration of backups work for every run?

alexey-tikhonov:
```
tests/test_authentication.py::test_authentication__with_default_settings[root-su] (ldap) PASSED [ 39%]
tests/test_authentication.py::test_authentication__with_default_settings[root-ssh] (ldap) FAILED [ 40%]
tests/test_authentication.py::test_authentication__with_default_settings[sssd-su] (ldap) PASSED [ 40%]
tests/test_authentication.py::test_authentication__with_default_settings[sssd-ssh] (ldap) FAILED [ 40%]
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[root-su] (ldap) PASSED [ 40%]
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[root-ssh] (ldap) FAILED [ 41%]
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[sssd-su] (ldap) PASSED [ 41%]
tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[sssd-ssh] (ldap) FAILED [ 41%]
```

pbrezina:
> > Hm, looks like the latest update have broken a bunch of system tests...
> 
> In all the broken tests it's either `ssh.password()` or `su.password()` what is failing. In both cases it gets EOF instead of expected prompt. SSSD seems to start correctly but there are no traces of related requests in SSSD logs.
> 
> @pbrezina, when a test is run multiple times due to parametrization like in:
> 
> ```
> tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[root-su] (samba) PASSED [ 75%]
> tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[root-ssh] (samba) FAILED [ 76%]
> tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[sssd-su] (samba) FAILED [ 76%]
> tests/test_authentication.py::test_authentication__default_settings_when_the_provider_is_offline[sssd-ssh] (samba) FAILED [ 76%]
> ```
> 
> -- does teardown / restoration of backups work for every run?

Yes, of course.

pbrezina:
> @pbrezina, any chance [containers/podman#9096](https://github.com/containers/podman/issues/9096) might affect this?



> @pbrezina, any chance [containers/podman#9096](https://github.com/containers/podman/issues/9096) might affect this?

It's possible.

sssd_test_framework/hosts/client.py
```diff
@@ -30,6 +30,10 @@ def __init__(self, *args, **kwargs) -> None:
 
         self._features: dict[str, bool] | None = None
 
+    def pytest_setup(self) -> None:
```
pbrezina, line 33: Don't forget to run `super().pytest_setup()`

sssd_test_framework/hosts/client.py
```diff
@@ -30,6 +30,10 @@ def __init__(self, *args, **kwargs) -> None:
...
+        result = self.conn.run("systemctl show sssd --value --property User")
```
pbrezina, line 34: You can use `self.svc.get_property()`

sssd_test_framework/utils/sssd.py
```diff
@@ -303,19 +310,40 @@ def clear(self, *, db: bool = True, memcache: bool = True, config: bool = False,
...
+            cmd += f'sed -i "s/^#SupplementaryGroups=sssd$/SupplementaryGroups=sssd/g" {service_file}\n'
+            cmd += f'sed -i "s/sssd:sssd/root:root/g" {service_file}\n'
+        elif user == "sssd":
+            cmd += f'sed -i "s/^SupplementaryGroups=sssd$/#SupplementaryGroups=sssd/g" {service_file}\n'
+            cmd += f'sed -i "s/root:root/sssd:sssd/g" {service_file}\n'
+        else:
+            raise ValueError("Unexpected value of 'user'")
+        cmd += f"chown -f {user}:{user} /var/lib/sss/db/*.ldb || true"
+        self.host.conn.run(cmd)
+        self.host.conn.run("systemctl daemon-reload")
```
pbrezina, line 346: `self.svc.reload_daemon`

sssd_test_framework/hosts/client.py
```diff
@@ -30,6 +30,10 @@ def __init__(self, *args, **kwargs) -> None:
...
+        self._sssd_service_user = result.stdout.strip()
```
pbrezina, line 35: Set this property in constructor to default value and set real value here. Also remove the underscore. This is a public property since you are using it outside of this class.

I.e.

```python
def __init__(...):
    self.sssd_service_user: str = "root"
    """Default SSSD service user."""

def pytest_setup(self) -> None:
    super().pytest_setup()
    self.sssd_service_user = self.svc.get_property(...)
```

This way, it will be correctly documented.

pbrezina (changes requested):

sssd_test_framework/utils/sssd.py
alexey-tikhonov: I can try again, but I tried this initially and it didn't work :-/

sssd_test_framework/hosts/client.py
alexey-tikhonov: Fixed

sssd_test_framework/hosts/client.py
alexey-tikhonov: Fixed

sssd_test_framework/hosts/client.py
alexey-tikhonov: Fixed (but I don't it to defaults to 'root')

alexey-tikhonov:
> Hm, looks like the latest update have broken a bunch of system tests...

Fixed now.

sssd_test_framework/utils/sssd.py
alexey-tikhonov: Works.

sssd_test_framework/hosts/client.py
pbrezina: Ok, then it can be done like this:

```
def __init__(self):
    self.__sssd_service_user: str | None = None

@property
def sssd_service_user(self):
    if self.__sssd_service_user is None:
        raise ValueError(...)

    return self.__sssd_service_user
```

But it's probably not necessary in this case.

justin-stephenson (approved):
Ack, thank you.

PR #121, opened 2024-09-13
Make code compatible with latest pytest-mh changes

pbrezina:
While rewriting pytest-mh documentation I realized that some methods
are marked with underscore to make them private, however, these methods
are actually expected to be overwritten. Therefore they should be public.

* https://github.com/next-actions/pytest-mh/pull/81
* https://github.com/next-actions/pytest-mh/pull/82

Additionally, I also make our backup code generic and pushed it into pytest-mh. This is not a breaking change, but it starts using this generic code instead of our own implementation.

* https://github.com/next-actions/pytest-mh/pull/79

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 jc
 pytest
 python-ldap
-pytest-mh >= 1.0.18
+pytest-mh >= 1.0.19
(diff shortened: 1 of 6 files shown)


justin-stephenson (dismissed):
Ack.

PR #122, opened 2024-09-16
hosts: Add restorecon to restore functions for ipa and client.

jakub-vavra-cz:
Make the restored config file have the right selinux context on client and ipa.

### DIFF
--- a/sssd_test_framework/hosts/client.py
+++ b/sssd_test_framework/hosts/client.py
@@ -145,6 +145,7 @@ class ClientHost(BaseBackupHost, BaseLinuxHost):
                 rm --force --recursive "$2"
                 if [ -d "$1" ] || [ -f "$1" ]; then
                     cp --force --archive "$1" "$2"
+                    restorecon -rf "$2"
                 fi
             }}
 


pbrezina:
What is the issue that you saw? `cp --archive` should restore the original selinux context associated with the backup file, so restorecon should not be needed.

And if you create a new file/dir (pytest-mh pull request), then you already should get the same context that restorecon would set, since restorecon sets the default context which is set automatically.

So in both cases restorecon should not be needed.

---

We can add restorecon here if we need to, but it will become a problem if we ever start testing against non-selinux distros like Debian. But ok, that's not likely to happens soon if ever.

We probably can't do that in pytest-mh since the code has to work on SELinux unaware systems as well.

sssd_test_framework/hosts/client.py
```diff
@@ -145,6 +145,7 @@ def restore(self, backup_data: Any | None) -> None:
                 rm --force --recursive "$2"
                 if [ -d "$1" ] || [ -f "$1" ]; then
                     cp --force --archive "$1" "$2"
+                    restorecon -rf "$2"
```
pbrezina, line 148:
```
       -f infilename
              infilename contains a list of files to be processed. Use “-” for stdin.
```
You probably want just `restorecon -r "$2"`

jakub-vavra-cz: I was not aware --archive  knows also selinux contexts.

jakub-vavra-cz: I will close this one and focus on the mh one as it is probably the root cause of the failures. (this was more like a bandaid)

PR #125, opened 2024-09-23
sssdutils: adding with-mkhomedir common configuration

danlavu:


danlavu:
Test >> https://github.com/SSSD/sssd/pull/7617

pbrezina (changes requested):
Hi Dan, could you please split the doc changes to another commit?

It is a good practice to keep commits small so it is easy to revert the patch if needed. If you revert the patch now, you also revert the doc changes.

Otherwise ack.

danlavu:
@pbrezina Done, added one thing to the function, parameterized the homedir path and it defaults to '/home'

pbrezina (approved):

PR #126, opened 2024-09-25
roles: added naming_context to generic ad role

danlavu:


danlavu:
I forgot to run mypy against the test code, and this is causing the sssd tests to fail.

https://github.com/SSSD/sssd/actions/runs/11038142612/job/30660629300?pr=7595

justin-stephenson (approved):

PR #127, opened 2024-09-26
roles: make naming_context a @property

pbrezina:
Otherwise we use the ssh/ldap connection inside the constructor
which can cause troubles.


justin-stephenson (approved):
Thanks, I missed that. Ack.

PR #129, opened 2024-10-01
authentication: detect a password change request

sumit-bose:
If a password is expired the user might be prompted to renew the
password immediately. This patch adds a detection for this case and
returns a dedicated error code.


pbrezina (approved):
Thank you, ack.

PR #131, opened 2024-10-04
ipa role: adding sshpubkey parameter to users

danlavu:


danlavu:
I'm still getting comfortable working with multiple commits and amending changes to previous commits. I have an extra commit I was trying to squash.

danlavu:
tests that use this enhancement https://github.com/SSSD/sssd/pull/7608

sssd_test_framework/roles/client.py
```diff
@@ -115,3 +115,16 @@ def sss_ssh_knownhosts(self, *args: str) -> ProcessResult:
...
+        Execute sss_ssh_authorizedkeys.
+
+        :param `*args`: Command arguments.
+        :type `*args`: str
+        :return: Command result.
+        :rtype: ProcessResult
+        """
+        result = self.host.conn.exec(["sss_ssh_authorizedkeys", *args], raise_on_error=False)
+
+        return result
```
spoore1, line 130: Couldn't this be combined and drop the result variable?

`return self.host.exec([...])`

sssd_test_framework/utils/tools.py
```diff
@@ -476,6 +478,11 @@ def __init__(self, host: MultihostHost, fs: LinuxFileSystem) -> None:
         Run ``getent`` command.
         """
 
+        self.ssh: SSHKeysUtils = SSHKeysUtils(host, fs)
```
spoore1, line 481: Could naming it `ssh` be confusing with the authentication ssh() method?   Since it's for key utils, maybe naming it something like `sshKey` would be more descriptive and obvious?

sssd_test_framework/utils/tools.py
```diff
@@ -733,3 +740,81 @@ def __exec(self, cls, cmd: str, name: str | int, service: str | None = None) ->
...
+        self.user: str | None = None
+        self.path: str | None = None
+        self.file: str | None = None
+
+    def keygen(
+        self,
+        user: str,
+        path: str,
+        file: str = "id_rsa",
+        cipher: str = "rsa",
```
spoore1, line 763: Is there any use for also supporting a `bits` option as well?  Or it can just be handled if needed with `args`?

sssd_test_framework/utils/tools.py
```diff
@@ -733,3 +740,81 @@ def __exec(self, cls, cmd: str, name: str | int, service: str | None = None) ->
...
+        group: str | None = None,
+        args: list[Any] | None = None,
+    ) -> SSHKeysUtils:
+        """
+        Creates home directory and SSH key pair for given user.
+
+        :param user: Username.
+        :type user: str
+        :param path: Homedir directory.
```
spoore1, line 772: Change `Homedir` to `Home`.

spoore1 (commented):
Overall looks good.   Just a few small things to check on.

sssd_test_framework/utils/tools.py
danlavu: That can be handled by args.

spoore1 (dismissed):
LGTM

sssd_test_framework/utils/tools.py
```diff
@@ -733,3 +740,81 @@ def __exec(self, cls, cmd: str, name: str | int, service: str | None = None) ->
...
+        :type group: str, optional
+        :param args: Additional arguments to pass to ssh-keygen, defaults to None
+        :type args: list[Any] | None
+        :return:
+        """
+
+        self.fs.backup("/home")
+        self.user = user
+        self.path = path
+        self.file = file
```
pbrezina, line 788: This won't work if you call keygen multiple times.

You can:
* return (pubkey, privatekey) pair instead of self
* or return a new object e.g. `SSHKey` that will hold the information

pbrezina (changes requested):

sssd_test_framework/utils/tools.py
pbrezina: But SSHKeysUtils is not create multiple times, it is created only once for each test.

If you want it to be created multiple times, then you should do it via `@property`. It also should not inherit from MultihostUtility, but from object in this case.

```python
@property
def keygen(...) -> SSHKeyUtils:
    return SSHKeyUtils(...)

client.tools.keygen.generate()
```

danlavu: Removed the class inheritance and made keygen a property that calls SSHKeyUtils

sssd_test_framework/utils/tools.py
```diff
@@ -733,3 +743,66 @@ def __exec(self, cls, cmd: str, name: str | int, service: str | None = None) ->
             return None
 
         return cls.FromOutput(command.stdout)
+
+
+class SSHKeyUtils:
+    """
+    Interface to ssh-keygen command.
```
spoore1, line 750: Will this just serve as an interface to ssh-keygen or possibly other ssh key related functions as well later?  If it may be used for a wider purpose, I think the name fits.  If it is only for ssh-keygen, maybe the class name should be SSHKeyGen or SSHKeyGenUtils?

Same for the property above.   I'm not going to push so if this makes more sense as is, I'm ok with it.

spoore1 (commented):
Just a question about the class name.

sssd_test_framework/utils/tools.py
danlavu: I'm not sure if we'll have more utilities for it, let's figure that out when/if we come to it?

sssd_test_framework/utils/tools.py
```diff
@@ -733,3 +743,66 @@ def __exec(self, cls, cmd: str, name: str | int, service: str | None = None) ->
...
+        self,
+        user: str,
+        path: str,
+        group: str | None = None,
+        file: str = "id_rsa",
+        cipher: str = "rsa",
+        args: list[Any] | None = None,
+    ) -> tuple[str, str]:
+        """
+        Creates user's home directory and SSH key pair.
```
pbrezina, line 773: Add blank line.

sssd_test_framework/utils/tools.py
```diff
@@ -733,3 +743,66 @@ def __exec(self, cls, cmd: str, name: str | int, service: str | None = None) ->
...
+        :param file: SSH key file, defaults to "id_rsa"
+        :type file: str, optional
+        :param cipher: Encryption algorithm, defaults to "rsa"
+        :type cipher: str, optional
+        :param args: Additional arguments to pass to ssh-keygen, defaults to None
+        :type args: list[Any] | None
+        :return: Public key, private key
+        :rtype: tuple[str, str]
+        """
+
```
pbrezina, line 789: Remove blank line :-)

sssd_test_framework/utils/tools.py
```diff
@@ -587,6 +587,16 @@ def faillock(self, args: list[Any]) -> ProcessResult:
...
 
+    @property
+    def sshkey(self) -> SSHKeyUtils:
+        """
+        Execute ssh-keygen command.
+        :return: SSHKeyUtils object.
+        :rtype: SSHKeyUtils
+        """
+
+        return SSHKeyUtils(self.host, self.__fs)
```
pbrezina, line 598: At this point, there is only single method "generate" and you do not set any internal state. So it is not necessary to create new object every time, but it's up to you.

pbrezina (changes requested):
Hi Dan, it looks good, thank you. See nitpicks inside.

sssd_test_framework/utils/tools.py
danlavu: Let's keep it as is. I can see us extending this eventually.

pbrezina (approved):

sssd_test_framework/utils/tools.py
spoore1: Sure.  Works for me.

spoore1 (approved):
LGTM

PR #134, opened 2024-10-22
utils: use host.cli instead of new instance

pbrezina:
`host.cli` was added later, we can just reuse it instead
of creating a new instance.


pbrezina:
Hmm, I just clicked on "rebase" and it got merged, not sure how this happened. Anyway, CI is green and this is just a small change.

PR #135, opened 2024-10-29
ipa: do not attempt to kinit if features are already detected

pbrezina:
kinit was run every time this property was accessed. This caused
"Calling exit but enter was not called" exception in pytest-mh
under special conditions:

- a test fails in a way that leaves IPA inoperable
- IPA restore does not work
- new test has `builtwith` marker that access IPA.features

In this case kinit was called, failed and exepction was thrown.
pytest-mh then skipped setup but attempted to teardown

### DIFF
--- a/sssd_test_framework/hosts/ipa.py
+++ b/sssd_test_framework/hosts/ipa.py
@@ -72,12 +72,11 @@ class IPAHost(BaseDomainHost, BaseLinuxHost):
         """
         Features supported by the host.
         """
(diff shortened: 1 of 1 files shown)


pbrezina:
The condition is also handled directly in https://github.com/next-actions/pytest-mh/pull/92 but it is still good to avoid unnecessary kinit call.

justin-stephenson (approved):
Ack.

PR #136, opened 2024-11-05
utils sssd, adding clean flag, clearing the cache on starts and restarts

danlavu:
Two commits:

1. Adding a clean parameter to mkhomedir common configuration, the override_homedirs are failing because some directories already exist from other tests.  After backing up homedir path, it will not clear it as well. 
2. Adding a clean parameter to sssd start, restart to condense 
```
client.sssd.stop()
client.sssd.clear()
client.sssd.start()
```
to 
```
client.sssd.restart(clean=True)
```

### DIFF
--- a/sssd_test_framework/hosts/client.py
+++ b/sssd_test_framework/hosts/client.py
@@ -113,6 +113,7 @@ class ClientHost(BaseHost, BaseLinuxHost):
             backup /etc/sssd "$path/config"
             backup /var/log/sssd "$path/logs"
             backup /var/lib/sss "$path/lib"
+            backup /home "$path/home"
 
             echo $path
             """,
@@ -148,12 +149,13 @@ class ClientHost(BaseHost, BaseLinuxHost):
                 fi
             }}
 
-            rm --force --recursive /etc/sssd /var/lib/sss /var/log/sssd
+            rm --force --recursive /etc/sssd /var/lib/sss /var/log/sssd /home
             restore "{backup_path}/krb5.conf" /etc/krb5.conf
             restore "{backup_path}/krb5.keytab" /etc/krb5.keytab
             restore "{backup_path}/config" /etc/sssd
             restore "{backup_path}/logs" /var/log/sssd
             restore "{backup_path}/lib" /var/lib/sss
+            restore "{backup_path}/home" /home
             """,
             log_level=ProcessLogLevel.Error,
         )
--- a/sssd_test_framework/utils/sssd.py
+++ b/sssd_test_framework/utils/sssd.py
@@ -900,20 +900,15 @@ class SSSDCommonConfiguration(object):
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
 
-    def mkhomedir(self, homedir: str = "/home") -> None:
+    def mkhomedir(self) -> None:
         """
         Configure SSSD with mkhomedir and oddjobd.
 
-        :param homedir: Home directory path.
-        :type homedir: str | None, optional
-
         #. Select authselect sssd profile with 'with-mkhomedir'
         #. Start oddjobd.service
-        #. Backup home directory
         """
         self.sssd.authselect.select("sssd", ["with-mkhomedir"])
         self.sssd.svc.start("oddjobd.service")
-        self.sssd.fs.backup(homedir)
 
     def proxy(
         self,
--- a/sssd_test_framework/utils/sssd.py
+++ b/sssd_test_framework/utils/sssd.py
@@ -119,6 +119,7 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         apply_config: bool = True,
         check_config: bool = True,
         debug_level: str | None = "0xfff0",
+        clean: bool = False,
     ) -> Process:
         """
         Start the SSSD and KCM services. Non-blocking call.
@@ -133,6 +134,8 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         :type check_config: bool, optional
         :param debug_level: Automatically set debug level to the given value, defaults to 0xfff0
         :type debug_level:  str | None, optional
+        :param clean: Does a clean restart, clearing the cache, defaults to False
+        :type clean: bool, defaults to False
         :return: Running SSH process.
         :rtype: Process
         """
@@ -145,6 +148,9 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         if service == "sssd":
             self.svc.async_stop("sssd-kcm.service")
 
+        if clean and service == "sssd":
+            self.clear()
+
         return self.svc.async_start(service)
 
     def start(
@@ -156,6 +162,7 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         apply_config: bool = True,
         check_config: bool = True,
         debug_level: str | None = "0xfff0",
+        clean: bool = False,
     ) -> ProcessResult:
         """
         Start the SSSD and KCM services. The call will wait until the operation is finished.
@@ -172,6 +179,8 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         :type check_config: bool, optional
         :param debug_level: Automatically set debug level to the given value, defaults to 0xfff0
         :type debug_level:  str | None, optional
+        :param clean: Does a clean restart, clearing the cache, defaults to False
+        :type clean: bool, defaults to False
         :return: SSH process result.
         :rtype: ProcessResult
         """
@@ -183,6 +192,9 @@ class SSSDUtils(MultihostUtility[MultihostHost]):
         # Also stop kcm so that it is started when first used.
         if service == "sssd":
             self.svc.stop("sssd-kcm.service")
+
+        if clean and service == "sssd":
+            self.clear()
 
         return self.svc.start(service, raise_on_error=raise_on_error)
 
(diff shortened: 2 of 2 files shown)


sssd_test_framework/utils/sssd.py
```diff
@@ -875,20 +903,25 @@ def autofs(self) -> None:
...
 
-    def mkhomedir(self, homedir: str = "/home") -> None:
+    def mkhomedir(self, homedir: str = "/home", clean: bool | None = False) -> None:
         """
         Configure SSSD with mkhomedir and oddjobd.
 
         :param homedir: Home directory path.
         :type homedir: str | None, optional
+        :param clean: Clears contents for a clean directory.
+        :type clean: bool | None, optional
```
jakub-vavra-cz, line 913:
```suggestion
        :type clean: bool
```

jakub-vavra-cz (changes requested):
Looks good, I would not use None for the clean as it dowes not make a sense to me.
Maybe the clean should be default True? (no strong opinion there)

sssd_test_framework/utils/sssd.py
```diff
@@ -875,20 +903,25 @@ def autofs(self) -> None:
         self.sssd.authselect.select("sssd")
         self.sssd.enable_responder("autofs")
 
-    def mkhomedir(self, homedir: str = "/home") -> None:
+    def mkhomedir(self, homedir: str = "/home", clean: bool | None = False) -> None:
```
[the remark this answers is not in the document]
danlavu: I wasn't thinking.

sssd_test_framework/utils/sssd.py
```diff
@@ -119,6 +119,7 @@ def async_start(
         apply_config: bool = True,
         check_config: bool = True,
         debug_level: str | None = "0xfff0",
+        clean: bool | None = False,
```
jakub-vavra-cz, line 122:
```suggestion
        clean: bool = False,
```

sssd_test_framework/utils/sssd.py
```diff
@@ -156,6 +162,7 @@ def start(
...
```
jakub-vavra-cz, line 165:
```suggestion
        clean: bool = False,
```

danlavu:
> Looks good, I would not use None for the clean as it dowes not make a sense to me. Maybe the clean should be default True? (no strong opinion there)

We would have to refactor the cache tests to set it to false. Let's minimize changing the behavior of existing methods.

jakub-vavra-cz:
> > Looks good, I would not use None for the clean as it dowes not make a sense to me. Maybe the clean should be default True? (no strong opinion there)
> 
> We would have to refactor the cache tests to set it to false. Let's minimize changing the behavior of existing methods.

Ok, then it makes a sense to keep it False . Disregard my previous comment.

sssd_test_framework/utils/sssd.py
```diff
@@ -875,20 +875,25 @@ def autofs(self) -> None:
...
         #. Select authselect sssd profile with 'with-mkhomedir'
         #. Start oddjobd.service
         #. Backup home directory
+        #. Clean home directory if set to ``True``
         """
         self.sssd.authselect.select("sssd", ["with-mkhomedir"])
         self.sssd.svc.start("oddjobd.service")
         self.sssd.fs.backup(homedir)
+        if clean:
+            self.sssd.fs.rm(f"{homedir}/*")
```
pbrezina, line 924: Can you check that the wildcard actually works? The command expands to `rm -fr 'homedir/*'` in single quotes, this should block the wildcard.

pbrezina: Can you point me to a test that is fixed by this?

danlavu: https://github.com/SSSD/sssd/pull/7682

It's hard to work on because this is failing in PRCI only. I've temporarily added print statements to a commit to see the directory's contents, and it does contain artifacts from existing tests. It can't create the directory because a homedir already exists and isn't owned by the user.

jakub-vavra-cz (dismissed):
LGTM

sssd_test_framework/utils/sssd.py
pbrezina: I will take some time and look into the issue. First, this line can not work, the wildcard will not be expanded. Second, the homedir is already backed up on the previous line so it should be restored to the original content after each test. If this is not happening, then we have a bug.

danlavu: @pbrezina , right, because the home dir is not empty, backup, clear, and restore to the state before the test is run, which includes some directories. FWIW, I do not see /home on the list to be backed up, so if a test does create a directory, it will persist.  

https://github.com/SSSD/sssd-test-framework/blob/7bcb65e2c47610aafc34d01494820bfe90c7ca22/sssd_test_framework/hosts/client.py#L91

So we add /home to that list, maybe enough to fix it.

pbrezina: there is `self.sssd.fs.backup(homedir)` right before your code, which defaults to /home. So /home is backed up once mkhomedir is setup. Noting that is created by it sould persist.

danlavu: https://github.com/SSSD/sssd/actions/runs/11959670217/job/33343281251?pr=7682#step:12:203

I added a 'ls -ltrh /home' before the setup, I did this before and I saw directories created from what I assume are previous tests. The host.backup doesn't restore this folder, so any content that is created from other tests will persist. I also added a clear statement, so they should also pass now. We should be able to look at the runner.log and see the contents before the method is called, to confirm.

pbrezina: host.backup does not restore it, fs module takes care of that

pbrezina: so the test is now failing for all providers, not just samba and this is because mkhomedir is in system-auth and oddjob is started by default, so other tests where user actually authenticates creates the home dir as a side effect. Adding /home to ClientHost.backup/restore will work.

I would:
* add backup and restore of /home to ClientHost
* ditch the commit that adds clean flag to mkhomedir as the wildcard won't work
* keep the second commit that adds clean to sssd, it probably is not needed in this case (I'd remove it from the test), but it may be helpful in other tests

danlavu: Exactly, so I think the best fix is to update host.backup to backup /home/ so the state is clean over the clear in this PR.

danlavu: I added another commit, add /home to backup and restore in hosts/client.py  , this should keep the directory clean.

pbrezina: Ditch the first commit and its an ack.

jakub-vavra-cz: @pbrezina How about actually fixing the wildcard for rm?  This is an unwanted/unexpected limitation that will just push people to use command rm directly instead.

jakub-vavra-cz (dismissed):

sssd_test_framework/hosts/client.py
```diff
@@ -107,7 +107,7 @@ def backup(self) -> Any:
                 fi
             }
 
-            path=`mktemp -d`
+            path=`mktemp -d`q
```
pbrezina, line 110: Is this intentional?

sssd_test_framework/utils/sssd.py
```diff
@@ -247,6 +262,11 @@ def async_restart(
         if service == "sssd":
             self.svc.async_stop("sssd-kcm.service")
 
+        if clean and service == "sssd":
+            self.svc.stop(service)
+            self.clear()
+            return self.svc.async_restart(service)
```
pbrezina, line 268: Just clear is sufficient as in other places, there is no need to stop it explicitly. It will also break the async purpose of this method.

sssd_test_framework/utils/sssd.py
```diff
@@ -281,6 +304,11 @@ def restart(
         if service == "sssd":
             self.svc.stop("sssd-kcm.service")
 
+        if clean and service == "sssd":
+            self.svc.stop(service)
+            self.clear()
+            return self.svc.start(service, raise_on_error=raise_on_error)
```
pbrezina, line 310: Just clear is sufficient as in other places, there is no need to stop it explicitly.

pbrezina (changes requested):

sssd_test_framework/hosts/client.py
danlavu: Not at all.

sssd_test_framework/utils/sssd.py
```diff
@@ -247,6 +262,10 @@ def async_restart(
         if service == "sssd":
             self.svc.async_stop("sssd-kcm.service")
 
+        if clean and service == "sssd":
+            self.clear()
+            return self.svc.async_restart(service)
```
pbrezina, line 267: Not needed, just fall through.

sssd_test_framework/utils/sssd.py
```diff
@@ -281,6 +303,10 @@ def restart(
         if service == "sssd":
             self.svc.stop("sssd-kcm.service")
 
+        if clean and service == "sssd":
+            self.clear()
+            return self.svc.start(service, raise_on_error=raise_on_error)
```
pbrezina, line 308: Not needed (and should be restart), just fall through.

sssd_test_framework/utils/sssd.py
```diff
@@ -888,7 +914,6 @@ def mkhomedir(self, homedir: str = "/home") -> None:
         """
         self.sssd.authselect.select("sssd", ["with-mkhomedir"])
         self.sssd.svc.start("oddjobd.service")
-        self.sssd.fs.backup(homedir)
```
pbrezina, line 891: Could you move this one to the second  commit please?

pbrezina (changes requested):

pbrezina (approved):

jakub-vavra-cz (approved):
LGTM

PR #144, opened 2024-12-20
authentication: adding hostname parameter to ssh method

danlavu:
Some tests will require to connect using the client hostname instead of localhost, adding this flexibility.

### DIFF
--- a/sssd_test_framework/utils/authentication.py
+++ b/sssd_test_framework/utils/authentication.py
@@ -599,7 +599,9 @@ class SSHAuthenticationUtils(MultihostUtility[MultihostHost]):
         self.opts = "-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
         """SSH CLI options."""
 
-    def password_with_output(self, username: str, password: str) -> tuple[int, int, str, str]:
+    def password_with_output(
+        self, username: str, password: str, hostname: str = "localhost"
+    ) -> tuple[int, int, str, str]:
         """
         SSH to the remote host and authenticate the user with password and captures standard output and error.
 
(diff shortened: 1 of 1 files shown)


danlavu:
```
INFO     2024-12-20 07:03:42,965          client.test Previous command completed with exit code 202
                                                      Output: |
                                                        spawn ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PreferredAuthentications=password -o NumberOfPasswordPrompts=1 -l user1 localhost
                                                        KnownHostsCommand-ORDER /usr/bin/sss_ssh_knownhosts localhost failed, status 1

                                                        KnownHostsCommand failed
```

Ran into the following issue, https://github.com/SSSD/sssd/issues/7664 , I'm sure the host is not up to date, but it bring this to my attention.

ikerexxe:
mmm let me see if I understand it correctly. If we assume a topology with 2 clients and an LDAP-like server. Currently we are only able to ssh to localhost, thus client1 can only ssh to client1 and client2 to client2. You want to use this feature to be able to ssh from client1 to client2 (or some other host).  Am I right?

danlavu:
> mmm let me see if I understand it correctly. If we assume a topology with 2 clients and an LDAP-like server. Currently we are only able to ssh to localhost, thus client1 can only ssh to client1 and client2 to client2. You want to use this feature to be able to ssh from client1 to client2 (or some other host). Am I right?

It can be used in that case, but specifically, this is for a bug with sss_known_hosts, https://github.com/SSSD/sssd/issues/7664 . This was causing the tests to fail. 

The other part of it is that some tests will want to connect to the hostname instead of localhost, i.e., certificate authentication, so it will be helpful down the road.

ikerexxe (approved):
Ok, it makes sense. Thank you for the explanation and the patch. LGTM!

PR #145, opened 2025-01-09
Adcli: adding class and methods for adcli

shridhargadekar:
Adding adcli class, and methods including info, discovery, join

### DIFF
--- a/sssd_test_framework/hosts/ad.py
+++ b/sssd_test_framework/hosts/ad.py
@@ -43,6 +43,9 @@ class ADHost(BaseDomainHost):
         self.adminpw: str = self.config.get("adminpw", "Secret123")
         """Password of the Administrator user, defaults to ``Secret123``."""
 
+        self.adminuser: str = self.config.get("adminuser", "administrator")
+        """Administrator user, defaults to ``administrator``."""
+
         # Additional client configuration
         self.client.setdefault("id_provider", "ad")
         self.client.setdefault("access_provider", "ad")
--- a/sssd_test_framework/hosts/ipa.py
+++ b/sssd_test_framework/hosts/ipa.py
@@ -51,6 +51,9 @@ class IPAHost(BaseDomainHost, BaseLinuxHost):
         self.adminpw: str = self.config.get("adminpw", "Secret123")
         """Password of the admin user, defaults to ``Secret123``."""
 
+        self.adminuser: str = self.config.get("adminuser", "admin")
+        """Administrator user, defaults to ``admin``."""
+
         self._features: dict[str, bool] | None = None
 
         # Additional client configuration
--- a/sssd_test_framework/hosts/samba.py
+++ b/sssd_test_framework/hosts/samba.py
@@ -30,7 +30,7 @@ class SambaHost(BaseLDAPDomainHost, BaseLinuxHost):
 
         self._features: dict[str, bool] | None = None
 
-        self.admin: str = self.config.get("username", "Administrator")
+        self.adminuser: str = self.config.get("username", "Administrator")
         """Username of the admin user, defaults to value of ``Administrator``."""
 
         self.adminpw: str = self.config.get("adminpw", self.bindpw)
--- a/sssd_test_framework/roles/samba.py
+++ b/sssd_test_framework/roles/samba.py
@@ -937,7 +937,7 @@ class SambaGPO(SambaObject):
         """Group policy search base."""
 
         # samba-tool gpo commands edit the database files directly and need to be authenticated.
-        self.credentials: str = f" --username={self.role.host.admin} --password={self.role.host.adminpw}"
+        self.credentials: str = f" --username={self.role.host.adminuser} --password={self.role.host.adminpw}"
         """Credentials to manage GPOs."""
 
     def add(self) -> SambaGPO:
@@ -986,7 +986,7 @@ class SambaGPO(SambaObject):
             "Guid": (self.cli.option.POSITIONAL, self.cn),
             "enforce": (self.cli.option.SWITCH, enforced),
             "disable": (self.cli.option.SWITCH, disabled),
-            "username": (self.cli.option.VALUE, self.role.host.admin),
+            "username": (self.cli.option.VALUE, self.role.host.adminuser),
             "password": (self.cli.option.VALUE, self.role.host.adminpw),
         }
 


sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,198 @@
...
+            "domain": (self.cli.option.VALUE, domain),
+            "domain_controller": (self.cli.option.VALUE, domain_controller),
+            "host_keytab": (self.cli.option.VALUE, host_keytab),
+        }
+        self.host.conn.exec(["adcli", "testjoin"] + self.cli.args(args))
+
+    def _join(
+        self,
+        *,
+        domain: str| None = None,
```
jakub-vavra-cz, line 75: There should be probably either domain or domain_controller.  
The manual page states that it expects domain so I would go with that.
Getting both None or both populated is a recipe for a disaster.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,198 @@
...
+    def __init__(self, host: MultihostHost, fs: LinuxFileSystem) -> None:
+        super().__init__(host)
+
+        self.cli: CLIBuilder = self.host.cli
+        """Command line builder."""
+
+        self.fs: LinuxFileSystem = fs
+        """Filesystem utils."""
+
+    def _info(
```
pbrezina, line 29: Why do you use underscores on the methods name? Those should be all public methods, right?

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,198 @@
...
+        self,
+        *,
+        domain: str | None = None,
+        domain_controller: str | None,
```
pbrezina, line 33: default to None as well?

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,198 @@
...
+        setattrs: str| None = None,
+        no_password: str| None = None,
+        promp_password: str| None = None,
+        stdin_password: str| None = None,
+        one_time_password: str| None = None,
+        show_password: str| None = None,
+        show_details: str| None = None,
+        add_samba_data: str| None = None,
+        samba_data_tool: str| None = None,
+        ldap_passwd: str| None = None,
```
pbrezina, line 103: Use **kwargs or *args to handle additional argument and keep only required arguments here. Otherwise it would be difficult to maintain.

pbrezina (changes requested):
I'd like to bring in Sumit as well.

I don't really know if we have use for this in SSSD, I don't really think so. I'm fine with including the patch though. However, in this case, operation that changes something must be also able to revert that change. In this case `join`, so this should be MultihostReentrantUtility.

sumit-bose:
> I'd like to bring in Sumit as well.
> 
> I don't really know if we have use for this in SSSD, I don't really think so. I'm fine with including the patch though. However, in this case, operation that changes something must be also able to revert that change. In this case `join`, so this should be MultihostReentrantUtility.

Hi,

on which level the revert should work. E.g. before calling `adcli join` we can save the current keytab and restore it afterwards. But since the state on AD/Samba DC will be changed as well they keys from the old keytab won't work anymore. So it might be better to keep the new keytab or do a leave/join cycle with fixed options to get back into a defined state?

bye,
Sumit

sssd_test_framework/utils/adcli.py
shridhargadekar: ack

pbrezina:
> > I'd like to bring in Sumit as well.
> > I don't really know if we have use for this in SSSD, I don't really think so. I'm fine with including the patch though. However, in this case, operation that changes something must be also able to revert that change. In this case `join`, so this should be MultihostReentrantUtility.
> 
> Hi,
> 
> on which level the revert should work. E.g. before calling `adcli join` we can save the current keytab and restore it afterwards. But since the state on AD/Samba DC will be changed as well they keys from the old keytab won't work anymore. So it might be better to keep the new keytab or do a leave/join cycle with fixed options to get back into a defined state?
> 
> bye, Sumit

It's a question of how is it going to be used.

Right now, in sssd-test-framework, we have this: https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/topology_controllers.py#L74

* pytest_setup -> each host performs a backup (only once at start)
* topology_setup -> run realmd join (only once when entering topology)
* topology teardown -> restore from the host backup (only once when leaving topology)

In adcli, I assume that Shridar is going to call adcli.join() insinde a test. So the logic should be:

MultihostUtility (cannot be used in MultihostHost objects, only in MultihostRole objects inside tests):
* utility.setup: do nothing or backup what is needed
* inside test: adcli.join
* utility.teardown: adcli.leave and clean up possible leftovers

MultihostReentrantUtility (cannot be used in MultihostHost objects and in MultihostRole objects):
* utility.\_\_enter__: enter new setup level, do nothing or backup what is needed
* inside test or topology setup: adcli.join
* utility.\_\_exit__: adcli.leave and clean up possible leftovers

Docs: https://pytest-mh.readthedocs.io/en/latest/articles/extending/multihost-utilities.html

Other way would be to rely on the backup/restore mechanisms to make sure everything is reverted after each test. But in this case, it needs to be thoroughly documented so the expectation is clear. That might be enough.

So it really depends on how is it going to be used.

jakub-vavra-cz:
@pbrezina, @shridhargadekar : We will probably need a way to either skip topology setup/teardown (no enrolling) so we can join ourselves or create topologies like KnownTopology.UnjoinedAD that will have the machines but not enrolled client. We can use these for example for tests like ldap sasl authid that needs a specific setup (--user-principal=host/name@REALM)  to work correctly.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,80 @@
...
+            "domain": (self.cli.option.VALUE, domain),
+            "domain_controller": (self.cli.option.VALUE, domain_controller),
+            "host_keytab": (self.cli.option.VALUE, host_keytab),
+        }
+        return self.host.conn.exec(["adcli", "testjoin"] + self.cli.args(args))
+
+    def join(
+        self,
+        domain: str,
+    ) -> str:
```
spoore1, line 70: Could we need to run join with additional options/arguments like computer name, host keytab, user principal or others?   If we do, I think we might be able to use something like the arg list used in utils/tools.py in SSHKeyUtils.generate here:

https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/utils/tools.py#L770

spoore1 (commented):
I think there are still some questions to be addressed but, it looks like a good start for an interface to adcli.   My only question currently is about additional join options.

sssd_test_framework/utils/adcli.py
spoore1: Does the change work?   I'd think you'd need to keep the domain inside the CLIBuilderArgs to do something like:

```
        if args is None:
            args = []

        bargs: CLIBuilderArgs = {
            "domain": (self.cli.option.POSITIONAL, domain),
        }
        self.host.conn.exec(["adcli", "join"] + self.cli.args(bargs) + [*args])
```

I haven't tested the above but, I'm just trying to mirror what was done in the SSHKeyUtils.generate().  Also, I'm not sure of the adcli command order.   Maybe args should be right after the join?

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,92 @@
...
+        self,
+        domain: str,
+        args: list[Any] | None = None
+    ) -> str:
+        """
+        call ``adcli join`` with given arguments.
+
+        """
+        args: CLIBuilderArgs = {
+        }
```
pbrezina, line 91: Can be removed.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,92 @@
...
+        self.host.conn.exec(["adcli", "join"] + self.cli.args(args))
```
pbrezina, line 92: `*args`

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,92 @@
...
+from pytest_mh.utils.fs import LinuxFileSystem
+
+__all__ = [
+    "ADCLIUtils",
+]
+
+
+class ADCLI(MultihostUtility[MultihostHost]):
+    """
+    Call commands from adcli
```
pbrezina, line 17: Make a note that this utility does not revert changes.

Given the use case, I think it is fine to rely on reverting the AD host in the topology controller, but it needs to be documented.

sssd_test_framework/utils/adcli.py
pbrezina, line 92: Does this call write any files? Like /etc/krb5.keytab maybe? If yes, you need to call fs.backup(file) to make sure it is restored. For AD state, we can rely on the topology controller, but we should handle any changes done to the host.

pbrezina (changes requested):

sssd_test_framework/utils/adcli.py
spoore1: I think the change with Pavel's suggested update resolves my questions for this one.   marking resolved.

sssd_test_framework/utils/adcli.py
shridhargadekar: implemented.

sssd_test_framework/utils/adcli.py
shridhargadekar: rearranged.

sssd_test_framework/utils/adcli.py
shridhargadekar: removed

sssd_test_framework/utils/adcli.py
shridhargadekar: dont need to back up any files.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,115 @@
...
+        call ``adcli delete computer `` with given arguments.
+
+        This method deletes a computer account in the domain
+
+        :param host: client host account to be deleted from AD
+        :type host: str | None,
+        :param args: A list of additional arguments, flags to be used with adcli
+        :type args: list | Any, optional
+        :type domain: list | None, optional
+        :param passwd: credential to be carry out the operations
```
spoore1, line 79: I'd suggest dropping "be" above.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,115 @@
...
+        user: str | None = "Administrator",
+        krb: bool | None = False,
+    ) -> ProcessResult:
+        """
+        call ``adcli join`` with given arguments.
+
+        :type domain: list | None, optional
+        :param args: A list of additional arguments, flags to be used with adcli
+        :type args: list | Any, optional
+        :param passwd: credential to be carry out the operations
```
spoore1, line 101: Drop be here too.

spoore1 (commented):
Just a couple of small comment nit-picks.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,109 @@
...
+    This utility will not revert any changes. It relies on AD host topology
+    for clean up.
+    """
+
+    def __init__(self, host: MultihostHost) -> None:
+        """
+        :param host: Remote host instance.
+        :type host: MultihostHost
+        """
+        super().__init__(host)
```
pbrezina, line 27: This can be omitted.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def info(self, domain: str, args: list[Any] | None = None) -> ProcessResult:
+        """
+        Call ``adcli info`` with given arguments.
+
+        :param domain: Active Directory domain to discover
+        :type domain: str,
+        :param args: A list of additional arguments, flags used with adcli
+        :type args: list | Any, optional
```
pbrezina, line 36: Fix type. Same on other places.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        super().__init__(host)
+
+    def info(self, domain: str, args: list[Any] | None = None) -> ProcessResult:
+        """
+        Call ``adcli info`` with given arguments.
+
+        :param domain: Active Directory domain to discover
+        :type domain: str,
+        :param args: A list of additional arguments, flags used with adcli
```
pbrezina, line 35: Mention default value. Same on other places.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,109 @@
...
+        This method will return if the machine account password and the join are still valid in AD
+
+        :param args: A list of additional arguments, flags used with adcli
+        :type args: list | Any, optional
+        """
+        if args is None:
+            args = []
+        return self.host.conn.exec(["adcli", *args, "--verbose", "testjoin"])
+
+    def del_com(
```
pbrezina, line 56: Do not shortcut the name.

pbrezina: `delete_computer`

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,109 @@
...
+        return self.host.conn.exec(["adcli", "delete-computer", *args, host], input=passwd)
+
+    def join(
+        self,
+        domain: str,
+        args: list[Any] | None = None,
+        *,
+        passwd: str | None = None,
+        user: str | None = "Administrator",
+        krb: bool | None = False,
```
pbrezina, line 86: All of this should be keyword only argument.

pbrezina, line 86: if you do not use None at all, you can remove it from the type hint

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,109 @@
...
+        :type krb: bool | False, optional
+        """
+        if args is None:
+            args = []
+        if krb is True:
+            self.host.conn.exec(["kinit", user], input=passwd)
+            return self.host.conn.exec(["adcli", "join", "--login-ccache", *args, domain])
+        else:
+            return self.host.conn.exec(
+                ["echo", passwd, "|", "adcli", "join", "--stdin-password", "--verbose", *args, domain]
```
pbrezina, line 108: remove echo and use `input=password`

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,109 @@
...
+        if args is None:
+            args = []
+        return self.host.conn.exec(["adcli", "delete-computer", *args, host], input=passwd)
+
+    def join(
+        self,
+        domain: str,
+        args: list[Any] | None = None,
+        *,
+        passwd: str | None = None,
```
pbrezina, line 84: Rename to password to make it consistent with other code.

pbrezina (changes requested):
Hi, see some minor comments inline.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,101 @@
...
+        :type args: list[Any] | None, optional
+        """
+        if args is None:
+            args = []
+
+        return self.host.conn.exec(["adcli", "info", *args, "--verbose", f"{domain}"])
+
+    def testjoin(self, args: list[Any] | None = None) -> ProcessResult:
+        """
+        call ``adcli testjoin`` with given arguments.
```
pbrezina, line 38: Capital C

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,101 @@
...
+
+        This method will return if the machine account password and the join are still valid in AD
+
+        :param args: A list of additional arguments, flags used with adcli, defaults to None
+        :type args: list[Any] | None, optional
+        """
+        if args is None:
+            args = []
```
pbrezina, line 46: Add new line.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,101 @@
...
+        return self.host.conn.exec(["adcli", *args, "--verbose", "testjoin"])
+
+    def delete_computer(
+        self,
+        host: str,
+        args: list[Any] | None = None,
+        password: str | None = None,
```
pbrezina, line 53: Use keyword only arguments here.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,101 @@
...
+    ) -> ProcessResult:
+        """
+        call ``adcli delete computer`` with given arguments.
```
pbrezina, line 56: Capital C

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,101 @@
...
+
+        :param host: client host account
+        :type host: str | None,
+        :param args: A list of additional arguments, flags to use with adcli, defaults to None
+        :type args: list[Any] | None, optional
+        :param password: credentials to carry out the operations, defaults to None
+        :type password: str | None, optional
+        """
+        if args is None:
+            args = []
```
pbrezina, line 68: Add blank line.

sssd_test_framework/utils/adcli.py
pbrezina: Not fixed yet.

sssd_test_framework/utils/adcli.py
shridhargadekar: done

sssd_test_framework/utils/adcli.py
shridhargadekar: Done

sssd_test_framework/utils/adcli.py
shridhargadekar: done

sssd_test_framework/utils/adcli.py
shridhargadekar: reverting  to unresolved state

sssd_test_framework/utils/adcli.py
shridhargadekar: done

sssd_test_framework/utils/adcli.py
shridhargadekar: dropped.

sssd_test_framework/utils/adcli.py
shridhargadekar: done

sssd_test_framework/utils/adcli.py
shridhargadekar: Added a method with `input=password`

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,168 @@
...
+        :param domain: domain.
+        :type domain: str,
+        :param args: additional arguments, defaults to None
+        :type args: list[str] | None, optional
+        :param password: password, defaults to None
+        :type password: str | None, optional
+        :param user: Authenticating User, defaults to None
+        :type user: str | None, optional
+        :param krb: Use kerberos credentials.
+        :type krb: bool | False, optional
```
spoore1, line 101: Missing stdin param/type entry.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,168 @@
...
+
+        if krb:
+            self.host.conn.exec(["kinit", user], input=password)
+            command = self.host.conn.exec(["adcli", "join", "--verbose", "--login-ccache", *args, domain])
+        elif prompt_password:
+            command = self.host.conn.exec(["adcli", "join", "-W", "--verbose", *args, domain], input=password)
+        elif stdin:
+            command = self.host.conn.exec(
+                ["echo", password, "|", "adcli", "join", "--stdin-password", "--verbose", *args, domain]
+            )
```
spoore1, line 114: If krb|prompt_password|stdin are mutually exclusive, is there a better way to either define it or better explanation/description in the docstring?

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,168 @@
...
+        username: str,
+        domain: str | None = None,
+        *,
+        args: list[str] | None = None,
+    ) -> ProcessResult:
+        """
+        Create user.
+
+        :param user: Username.
+        :type user: str
```
spoore1, line 129: Shouldn't this be `username` instead of `user`?

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,168 @@
...
+        :param domain: Domain.
+        :type domain: str,
+        :param args: additional arguments, defaults to None
+        :type args: list[str] | None, optional
+        :param password: Password.
+        :type password: str | None, optional
+        :param krb: Use kerberos credentials.
+        :type krb: bool | False, optional
```
spoore1, line 137: I think you can drop `password` and `krb` entries from the docstring.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,168 @@
...
+
+        return self.host.conn.exec(["adcli", "delete-computer", *args, host], input=password)
+
+    def join(
+        self,
+        domain: str,
+        *,
+        args: list[str] | None = None,
+        password: str | None = None,
+        user: str | None = None,
```
spoore1, line 84: Should `join()` use `username` instead of `user` to match the other methods?

spoore1 (changes requested):
A few small things with docstrings and variable names.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,168 @@
...
+
+        :param args: additional arguments, defaults to None
+        :type args: list[str] | None, optional
+        """
+        if args is None:
+            args = []
+
+        return self.host.conn.exec(["adcli", *args, "--verbose", "testjoin"])
+
+    def delete_computer(
```
jakub-vavra-cz, line 56: We might want to have here also show-computer to be able to check that the delete worked.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,168 @@
...
+        elif prompt_password:
+            command = self.host.conn.exec(["adcli", "join", "-W", "--verbose", *args, domain], input=password)
+        elif stdin:
+            command = self.host.conn.exec(
+                ["echo", password, "|", "adcli", "join", "--stdin-password", "--verbose", *args, domain]
+            )
+
+        return command
+
+    def create_user(
```
jakub-vavra-cz, line 118: We might want to have also passwd-user so we can set password and try to login as the user.

sssd_test_framework/utils/adcli.py
shridhargadekar: rearranged all methods. After discussing it with Pavel, I've dropped stdin-param in favour of `input=password` method. They are essentially similar.

sssd_test_framework/utils/adcli.py
shridhargadekar: I would love to add that explanation about mutual exclusiveness in docstrings.

sssd_test_framework/utils/adcli.py
shridhargadekar: dropped

sssd_test_framework/utils/adcli.py
shridhargadekar: rearranged.

sssd_test_framework/utils/adcli.py
shridhargadekar: added

sssd_test_framework/utils/adcli.py
shridhargadekar: Please check now.

sssd_test_framework/utils/adcli.py
shridhargadekar: I've rearranged the logic. Check now.

spoore1: You dropped the --stdin-password option?   If that's ok, then this can be resolved but, if we need that, maybe just a note in the docstring would be enough here.

shridhargadekar: `--stdin-password` and `-W` are doing almost same thing. `--stdin-passwod` is useful with BASH automation, whereas `-W` working with `input=password` is doing same.

shridhargadekar: Do you want a note about `-W` and `--stdin-password` in each method docstring or at class-level note?  I think adding note at each method would be more accessible in long run.

shridhargadekar: I've rewritten the method to have `-W` that is `--prompt-password` as default. Added only condition to check if `kerberos` is `True` or not.

shridhargadekar: Added a note in docstring about `prompt-password` and `--stdin-password`.

spoore1 (commented):
LGTM.    Is there a test somewhere I could use to try it out?

spoore1 (changes requested):
A few small findings after testing.

jakub-vavra-cz (approved):

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,175 @@
...
+        if krb:
+            self.host.conn.exec(["kinit", login_user], input=password)
+            command = self.host.conn.exec(["adcli", "join", "--verbose", "-C", *args, domain])
+        else:
+            command = self.host.conn.exec(
+                ["adcli", "join", "--prompt-password", "--verbose", f"--login-user={login_user}", *args, domain],
+                input=password,
+            )
+
+        return command
```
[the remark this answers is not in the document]
shridhargadekar: I've update the test case PR with stderr

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        if args is None:
+            args = []
+
+        if krb:
+            self.host.conn.exec(["kinit", login_user], input=password)
+            command = self.host.conn.exec(["adcli", "join", "--verbose", "-C", *args, domain])
+        else:
+            command = self.host.conn.exec(
+                ["adcli", "join", "--prompt-password", "--verbose", f"--login-user={login_user}", *args, domain],
```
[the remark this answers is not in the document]
shridhargadekar: changed to `--stdin-password`

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,182 @@
...
+        return self.host.conn.exec(["adcli", "testjoin", "--verbose", domain, *args])
+
+    def join(
+        self,
+        domain: str,
+        *,
+        args: list[str] | None = None,
+        password: str,
+        login_user: str,
+        krb: bool | None = False,
```
pbrezina, line 75: Does not take None anymore.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,182 @@
...
+        :param domain: Domain.
+        :type domain: str,
+        :param args: Additional arguments, defaults to None
+        :type args: list[str] | None, optional
+        :param password: Password,
+        :type password: str,
+        :param login_user: Authenticating User,
+        :type login_user: str,
+        :param krb: Use Kerberos credentials.
+        :type krb: bool | False, optional
```
pbrezina, line 89: Just `bool, optional`

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,182 @@
...
+
+        :param domain: Domain.
+        :type domain: str,
+        :param args: Additional arguments, defaults to None
+        :type args: list[str] | None, optional
+        :param password: Password,
+        :type password: str,
+        :param login_user: Authenticating User,
+        :type login_user: str,
+        :param krb: Use Kerberos credentials.
```
pbrezina, line 88: `..., defaults to False`

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,182 @@
...
+
+        return command
+
+    def delete_computer(
+        self,
+        *,
+        args: list[str] | None = None,
+        password: str,
+        login_user: str,
+        krb: bool | None = False,
```
pbrezina, line 122: Same for the type and docstring as in previous comment.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,182 @@
...
+        return command
+
+    def show_computer(
+        self,
+        domain: str,
+        *,
+        password: str,
+        args: list[str] | None = None,
+        login_user: str,
+        krb: bool | None = False,
```
pbrezina, line 159: Same here.

sssd_test_framework/utils/adcli.py
```diff
@@ -0,0 +1,182 @@
...
+        :param login_user: Authenticating User,
+        :type login_user: str,
+        :param krb: Use Kerberos credentials.
+        :type krb: bool | False, optional
+        """
+        if args is None:
+            args = []
+
+        if krb:
+            self.host.conn.exec(["kinit", f"{login_user}@{domain.upper()}"], input=password)
```
pbrezina, line 142: domain is not defined

pbrezina (changes requested):

sssd_test_framework/utils/adcli.py
shridhargadekar: done

PR #149, opened 2025-01-28
tools.py - ssh-keygen

danlavu:
* renamed the path variable to homedir, matching other functions
* added comment to the keys making it easier to distinguish keys during debugging


spoore1 (approved):
LGTM

pbrezina:
> added comment to the keys making it easier to distinguish keys during debugging

I don't see this change, is it still relevant?

PR #152, opened 2025-02-07
utils: set_service_user skip on read only filesystem

jakub-vavra-cz:
In image mode the sevice file is not editable so we skip tests that are not applicable.

### DIFF
--- a/sssd_test_framework/utils/sssd.py
+++ b/sssd_test_framework/utils/sssd.py
@@ -6,6 +6,7 @@ import configparser
 from io import StringIO
 from typing import TYPE_CHECKING, Literal
 
+import pytest
 from pytest_mh import MultihostHost, MultihostRole, MultihostUtility
 from pytest_mh.conn import Process, ProcessLogLevel, ProcessResult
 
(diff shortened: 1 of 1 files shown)


spoore1 (approved):
This looks like it should work to me.   My only question: 

Is there a way to deselect the tests that this would skip instead of running and showing skip?

I'm ok with this for now, just thinking that if there is a better way to handle deselection we may want to implement it later.

pbrezina:
Why skip? Instead of modifying service file in /usr put override to /etc.

jakub-vavra-cz:
> Why skip? Instead of modifying service file in /usr put override to /etc.
We could probably do that as well. I think that switching user is covered enough outside of image mode.

jakub-vavra-cz:
> This looks like it should work to me. My only question:
> 
> Is there a way to deselect the tests that this would skip instead of running and showing skip?
> 
> I'm ok with this for now, just thinking that if there is a better way to handle deselection we may want to implement it later.

Pavel could probably figure out how to do that in similar way to topologies, but skipping tests in the run with a reason message makes it more transparent so I would rather keep it this way.

PR #157, opened 2025-03-14
tools.py - grep

thalman:
Grep should return False when pattern is not present, not raise an exception.


danlavu (approved):
I see why this is a problem, thanks.

pbrezina (approved):

PR #162, opened 2025-03-26
configuring sssd to use local domain when sssd.common.local() is called

danlavu:


sssd_test_framework/utils/sssd.py
```diff
@@ -832,6 +832,7 @@ def local(self) -> None:
             proxy_lib_name="files",
             proxy_pam_target="system-auth",
         )
+        self.sssd.default_domain = "local"
```
jakub-vavra-cz, line 835: I would put local as a default value in the SSSDUtils https://github.com/SSSD/sssd-test-framework/blob/master/sssd_test_framework/utils/sssd.py#L79.

We want default domain to be always a string instead of possible None.
That way I would not need to do a crap like this in order to please static analysis:
result = client.sssctl.domain_status(client.sssd.default_domain if client.sssd.default_domain else "local")

I would consider setting the SSSDUtils to use proxy local as the default (init) that will be done unless You explicitly set it to something else.

danlavu: The client is pre-configured with all topologies. Freshly built container, 

```
[root@client sssd]# cat sssd.conf 
[sssd]
services = nss, pam
domains = ipa.test, samba.test, ldap.test

[domain/ldap.test]
id_provider = ldap
ldap_uri = _srv_
ldap_tls_reqcert = demand
ldap_tls_cacert = /data/certs/ca.crt
dns_discovery_domain = ldap.test
use_fully_qualified_names = true

[domain/ipa.test]
id_provider = ipa
access_provider = ipa
ipa_server = _srv_
ipa_domain = ipa.test
ipa_hostname = client.test
krb5_keytab = /enrollment/ipa.test.keytab
ldap_krb5_keytab =  /enrollment/ipa.test.keytab
use_fully_qualified_names = true

[domain/samba.test]
id_provider = ad
access_provider = ad
ad_server = _srv_
ad_domain = samba.test
ad_hostname = client.test
krb5_keytab = /enrollment/samba.test.keytab
ldap_krb5_keytab = /enrollment/samba.test.keytab
use_fully_qualified_names = true
```
Giving it a default value of 'local' is strange considering.  It's evaluated in a few places to 'None'. Line 570, 563 and 419. The proxy local setup, we need to rewrite this method. 

https://github.com/SSSD/sssd-test-framework/blob/caed70f0d7a05562b5948cf3d5d1ea421fee82c8/sssd_test_framework/utils/sssd.py#L556

For now, it doesn't make sense to have that extra line in the test case, it should be wrapped into the method. 

```client.sssd.config.local()``` it's configured and ready to go.

Comment (jakub-vavra-cz) [reply sssd_test_framework/utils/sssd.py:835]: My idea was that client (by default) is a standalone machine running proxy files as the minimal configuration.
I would be pushing to have the proxy local config set when there are no other machines in the environment.
I avoid the multi-type varibles that  are str | None as static analysis quite often needlessly complains about them.
Having always a sane value or "" seem better to me as you can depend on the type.

File: sssd_test_framework/utils/sssd.py [hunk 2]
```diff
@@ -832,6 +832,7 @@ def local(self) -> None:
             proxy_lib_name="files",
```
justin-stephenson, line 832: I don't insist but maybe it makes sense to have this `local` call the self.proxy() method instead with the appropriate arguments, which already sets the default_domain

justin-stephenson (approved):

sssd_test_framework/utils/sssd.py
danlavu: I intend to rework the proxy provider stuff, so this is an interim fix.

sssd_test_framework/utils/sssd.py
danlavu: The multi-type variables are all over the framework, it's way too embedded to change. Let's figure out the proxy methods and figure out how we want to use and then change all of this stuff.

PR #165, opened 2025-03-29
gdm idp utils

spoore1:


sssd_test_framework/utils/gdm.py
```diff
@@ -0,0 +1,80 @@
...
+class GDM(MultihostUtility[MultihostHost]):
+    """
+    Call commands from scautolib.
+    """
+
+    def init(self) -> None:
+        """
+        Initialize GDM for testing
+        """
+        self.host.conn.exec(["/opt/test_venv/bin/scauto", "gui", "init"])
```
ikerexxe, line 24: IMO, this should be handled when any other method is called and keep in memory whether `scauto` has been initialized.

sssd_test_framework/utils/gdm.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def done(self) -> bool:
+        """
+        Run scauto gui done
+
+        :return: run cleanup after gui tests
+        :rtype: bool
+        """
+        result = self.host.conn.exec(["/opt/test_venv/bin/scauto", "gui", "done"])
+        return result == 0
```
ikerexxe, line 80: This should be handled in the teardown phase when `scauto` has previously been initialized.

sssd_test_framework/utils/idp.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Authenticate to Keycloak using External IdP Device Authorization Grant and return output
+
+        :return: Tuple containing [return code, stdout, stderr]
+        :rtype: Tuple[int, str, str]
+        """
+
+        command = self.fs.mktmp(
+            rf"""
+            #!/opt/sssd_test_venv/bin/python3
```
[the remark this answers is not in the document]
ikerexxe: I insist on this one. You should create a file with the selenium code, and copy it to the remote host in the python code. This will make the selenium code easier to review and maintain.

sssd_test_framework/utils/gdm.py
ikerexxe: Still applies

sssd_test_framework/utils/gdm.py
ikerexxe: Still applies

ikerexxe:
Some minor issues still apply. Everything else looks good. Please rebase on top of master to avoid any merging conflicts.

sssd_test_framework/utils/idp.py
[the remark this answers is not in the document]
spoore1: Ok, I've moved it to sssd-ci-containers in that PR so I'll add you to review that one as well.

sssd_test_framework/utils/gdm.py
spoore1: I implemented this.  Take a look and let me know if I missed something.

sssd_test_framework/utils/gdm.py
ikerexxe: I think this can be handled by implementing the `teardown` method for this class. That will take care of cleaning everything when the test case is finished. As an example you can check `sssd_test_framework/utils/authselect.py`.

spoore1: Ok, looks like it ran as expected from the util teardown.    So unless there's something I should update, I think this one at least is resolved.

ikerexxe (approved):

PR #166, opened 2025-03-31
sssd.py: Allow systemd override drop-in file for service user

jakub-vavra-cz:
On read only filesystem systems sssd service file can not be changed directly and override file needs to be created instead.

### DIFF
--- a/sssd_test_framework/utils/sssd.py
+++ b/sssd_test_framework/utils/sssd.py
@@ -6,7 +6,6 @@ import configparser
 from io import StringIO
 from typing import TYPE_CHECKING, Literal
 
-import pytest
 from pytest_mh import MultihostHost, MultihostRole, MultihostUtility
 from pytest_mh.conn import Process, ProcessLogLevel, ProcessResult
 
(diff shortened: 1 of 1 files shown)


sssd_test_framework/utils/sssd.py
```diff
@@ -360,21 +359,27 @@ def set_service_user(self, user: str) -> None:
...
+                "/etc/systemd/system/sssd.service.d/user.conf", contents, mode="0600", user="root", group="root"
+            )
         else:
-            raise ValueError("Unexpected value of 'user'")
+            self.fs.backup(service_file)
+            cmd += f'sed -i "s/^User=.*/User={user}/g" {service_file}\n'
+            cmd += f'sed -i "s/^Group=.*/Group={user}/g" {service_file}\n'
+            if user == "root":
+                cmd += f'sed -i "s/^#SupplementaryGroups=sssd$/SupplementaryGroups=sssd/g" {service_file}\n'
+                cmd += f'sed -i "s/sssd:sssd/root:root/g" {service_file}\n'
```
alexey-tikhonov, line 377: This edits args of 'chown' in main file. This is missing in override.

alexey-tikhonov:
I think it won't work.
But I'm not sure we still need to keep this code.
Maybe it's fine to drop service-user parametrization.

sssd_test_framework/utils/sssd.py
jakub-vavra-cz: Added extra code for this and trying again.

sssd_test_framework/utils/sssd.py
```diff
@@ -360,21 +359,27 @@ def set_service_user(self, user: str) -> None:
...
-        cmd += f'sed -i "s/^Group=.*/Group={user}/g" {service_file}\n'
-        if user == "root":
-            cmd += f'sed -i "s/^#SupplementaryGroups=sssd$/SupplementaryGroups=sssd/g" {service_file}\n'
-            cmd += f'sed -i "s/sssd:sssd/root:root/g" {service_file}\n'
-        elif user == "sssd":
-            cmd += f'sed -i "s/^SupplementaryGroups=sssd$/#SupplementaryGroups=sssd/g" {service_file}\n'
-            cmd += f'sed -i "s/root:root/sssd:sssd/g" {service_file}\n'
+            self.logger.info(f"Using service overlay on read only filesystem on {self.host.hostname}")
+            self.fs.mkdir("/etc/systemd/system/sssd.service.d")
+            supp = "SupplementaryGroups=sssd\n" if user == "root" else ""
```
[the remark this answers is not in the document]
jakub-vavra-cz: Would SupplementaryGroups="" work?

sssd_test_framework/utils/sssd.py
alexey-tikhonov: I don't need.
This needs to be tested.
Could you please capture output of `systemctl cat sssd` both for 'sssd -> root' and 'root -> sssd' transition?

sssd_test_framework/utils/sssd.py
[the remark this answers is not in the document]
alexey-tikhonov: I don't need.
This needs to be tested.
Could you please capture output of `systemctl cat sssd` both for 'sssd -> root' and 'root -> sssd' transition?

sssd_test_framework/utils/sssd.py
alexey-tikhonov: We need to see resulting service file to make sure the order is right (i.e. override's 'ExecStartPre' go *after* original content):
```
Multiple command lines are allowed, regardless of the service type (i.e. Type=),
and the commands are executed one after the other, serially.
```

jakub-vavra-cz: The override was shown after the original unit wnen using systecmtcl cat

jakub-vavra-cz:
> I think it won't work. But I'm not sure we still need to keep this code. Maybe it's fine to drop service-user parametrization.

We can drop it just for image mode and run the test if the configured user matches the tested one .

jakub-vavra-cz:
> I think it won't work. But I'm not sure we still need to keep this code. Maybe it's fine to drop service-user parametrization.

I am reworking this to skip on ro/fs when the user is differet than the current one. That way we will still have coverage for the default user.

PR #168, opened 2025-04-05
topologies: adding any directory server (ds)

danlavu:
DS, any server that contains ldap, kerberos and a ca by default. This group will contain AD, Samba and IPA. Dynamic DNS, PKI tests will use this TopologiesGroup.

### DIFF
--- a/sssd_test_framework/topology.py
+++ b/sssd_test_framework/topology.py
@@ -145,6 +145,11 @@ class KnownTopologyGroup(KnownTopologyGroupBase):
     .. topology-mark:: KnownTopologyGroup.AnyAD
     """
 
+    AnyDC = [KnownTopology.AD, KnownTopology.Samba, KnownTopology.IPA]
+    """
+    ..topology-mark:: KnownTopologyGroup.AnyDC
+    """
+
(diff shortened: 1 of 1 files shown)


sssd_test_framework/topology.py
```diff
@@ -145,6 +145,11 @@ def test_ldap(client: Client, provider: GenericProvider):
     .. topology-mark:: KnownTopologyGroup.AnyAD
     """
 
+    AnyDS = [KnownTopology.AD, KnownTopology.Samba, KnownTopology.IPA]
+    """
+    ..topology-mark:: KnownTopologyGroup.AnyDS
```
pbrezina, line 150: Maybe AnyDomainController?

DS = Directory Server includes also 389ds, OpenLDAP... well any LDAP.

pbrezina (changes requested):

sssd_test_framework/topology.py
danlavu: I was thinking domain controller, but wrote down ds.

danlavu: Updated.

PR #171, opened 2025-04-10
implemented smartcard util

krishnavema:


sssd_test_framework/roles/client.py
```diff
@@ -126,3 +131,67 @@ def sss_ssh_authorizedkeys(self, *args: str) -> ProcessResult:
...
+        :param key_id: ID to associate with the private key.
+        :param cert_id: ID to associate with the certificate.
+        :param pin: User PIN for smart card authentication.
+        :param subj: Subject DN for the certificate.
+        :return: Tuple of paths (key_path, cert_path) on the remote host.
+        """
+        key_path, cert_path = self.smart_card.generate_ca_cert(subj=subj)
+        self.smart_card.add_key(key_path, key_id, pin)
+        self.smart_card.add_cert(cert_path, cert_id, pin)
+        return key_path, cert_path
```
spoore1, line 197: So this functions like the other utils, I think these additional helper methods for the client role class can be dropped and referenced directly from "client.smart_card.function()".

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        self.fs = LinuxFileSystem(host.conn)
+    
+    def init(self, label: str, so_pin: str, user_pin: str) -> None:
+        """
+        Initializes a SoftHSM token with the given label and PINs.
+        
+        :param label: The label for the new token.
+        :param so_pin: Security Officer PIN.
+        :param user_pin: User PIN.
+        """
```
spoore1, line 28: I'm not sure it belongs in init() but we will need to make sure to do some additional cleanup of caches when we initialize or teardown.   We need something like this:
```
        self.host.conn.run('rm -rf $HOME/.cache/opensc/ /run/sssd/.cache/opensc/ /opt/test_ca/tokens')
        self.host.conn.run('mkdir -p /opt/test_ca/tokens')
```

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        self.host.conn.exec([
+            "softhsm2-util", "--init-token",
+            "--label", label,
+            "--free",
+            "--so-pin", so_pin,
+            "--pin", user_pin
+        ])
```
spoore1, line 35: In order for the virtual smart card setup to work, we need to make sure that softhsm2-util and libsofthsm2.so are using the correct config file by adding it like this:
```
        self.host.conn.exec([
            "softhsm2-util", "--init-token",
            "--label", "sc_test",
            "--free",
            "--so-pin", self.so_pin,
            "--pin", self.user_pin
        ],
        env={"SOFTHSM2_CONF": "/opt/test_ca/softhsm2.conf"})
```

So you should just need to add `env={"SOFTHSM2_CONF": "/opt/test_ca/softhsm2.conf"}` to the args for each `host.conn.exec()`

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        :param pin: User PIN for authentication.
+        """
+        self.fs.copy(key_path, "/tmp/key.pem")
+        self.host.conn.exec([
+            "pkcs11-tool", "--module", "/usr/lib64/pkcs11/libsofthsm2.so",
+            "-l", "--pin", pin,
+            "--write-object", "/tmp/key.pem",
+            "--type", "privkey",
+            "--id", key_id
+        ])
```
spoore1, line 52: need to set env SOFTHSM2_CONF here too

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        :param pin: User PIN for authentication.
+        """
+        self.fs.copy(cert_path, "/tmp/cert.pem")
+        self.host.conn.exec([
+            "pkcs11-tool", "--module", "/usr/lib64/pkcs11/libsofthsm2.so",
+            "-l", "--pin", pin,
+            "--write-object", "/tmp/cert.pem",
+            "--type", "cert",
+            "--id", cert_id
+        ])
```
spoore1, line 69: need to set env SOFTHSM2_CONF here too

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        Restarts the virtual smart card service.
+        """
+        self.host.svc.restart("virt_cacard.service")
+    
+    def insert_card(self) -> None:
+        """
+        Starts the virtual smart card service to simulate card insertion.
+        """
+        self.host.conn.exec(["systemctl", "start", "virt_cacard.service"])
```
spoore1, line 81: Might want to change to `virt_cacard -i`

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+    
+    def remove_card(self) -> None:
+        """
+        Stops the virtual smart card service to simulate card removal.
+        """
+        self.host.conn.exec(["systemctl", "stop", "virt_cacard.service"])
```
spoore1, line 87: Might want to change to `virt_cacard -r`

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self, host: MultihostHost):
+        """
+        Initializes the SmartCardUtils with a given host.
+        
+        :param host: The remote host where smart card operations will be executed.
+        """
+        self.host = host
+        self.fs = LinuxFileSystem(host.conn)
+    
+    def init(self, label: str, so_pin: str, user_pin: str) -> None:
```
spoore1, line 21: I'd suggest dropping the params from init (and most of them from the other methods too).  In most cases we're not going to need to use different labels or pins.   If you want to leave it in case it's needed later, I'd suggest making them optional parameters.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        self.host.conn.exec([
+            "softhsm2-util", "--init-token",
+            "--label", label,
+            "--free",
+            "--so-pin", so_pin,
+            "--pin", user_pin
+        ])
+    
+    def add_key(self, key_path: str, key_id: str, pin: str) -> None:
```
spoore1, line 37: I'd change key_id and pin to optional and set something reasonable like '01' and '123456' as defaults.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        Adds a private key to the smart card.
+        
+        :param key_path: Path to the private key file.
+        :param key_id: ID for the key in the smart card.
+        :param pin: User PIN for authentication.
+        """
+        self.fs.copy(key_path, "/tmp/key.pem")
```
spoore1, line 45: You can skip the copy and directly use the key_path variable below.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+    
+    def add_cert(self, cert_path: str, cert_id: str, pin: str) -> None:
+        """
+        Adds a certificate to the smart card.
+        
+        :param cert_path: Path to the certificate file.
+        :param cert_id: ID for the certificate in the smart card.
+        :param pin: User PIN for authentication.
+        """
+        self.fs.copy(cert_path, "/tmp/cert.pem")
```
spoore1, line 62: You can skip the copy here also and use cert_path directly below.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        self.host.conn.exec(["systemctl", "start", "virt_cacard.service"])
+    
+    def remove_card(self) -> None:
+        """
+        Stops the virtual smart card service to simulate card removal.
+        """
+        self.host.conn.exec(["systemctl", "stop", "virt_cacard.service"])
+
+    def generate_ca_cert(self, key_path: str = "/tmp/ca.key", cert_path: str = "/tmp/ca.crt", subj: str = "/CN=Test CA") -> tuple[str, str]:
```
spoore1, line 89: I'd suggest for this change currently, to change this to `generate_self_signed_cert()` and change CA references to just Self Signed.   I believe I must have misspoken and mentioned CA in reference to creating a cert like this for the user.   Creating an openssl CA is a little more involved and we will need to do that later but, this code should work for creating the self-signed certificate.  I have one suggestion to simplify into one command though.  Besides that, just a name change for method and variables/paths.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,108 @@
...
+        self.host.conn.exec([
+            "openssl", "genrsa", "-out", key_path, "2048"
+        ])
+        self.host.conn.exec([
+            "openssl", "req", "-x509", "-new", "-nodes",
+            "-key", key_path,
+            "-sha256", "-days", "365",
+            "-out", cert_path,
+            "-subj", subj
+        ])
```
spoore1, line 107: This can be combined into a single command like this:
```
        self.host.conn.exec([
            "openssl", "req", "-x509", "-nodes",
            "-sha256", "-days", "365",
            "-newkey", "rsa:2048",
            "-keyout", key_path,
            "-out", cert_path,
            "-subj", subj
        ])
```

spoore1 (changes requested):
This is a great first draft for the smart card utils.   I was able to take this with a few small changes and get it working with other related in progress work to actually run a local user smart card test.

sssd_test_framework/utils/virtual_smartcard.py
krishnavema: Resloved

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,134 @@
...
+    TOKEN_STORAGE_PATH = "/opt/test_ca/tokens"
+    OPENSC_CACHE_PATHS = [
+        "$HOME/.cache/opensc/",
+        "/run/sssd/.cache/opensc/",
+    ]
+
+    def __init__(self, host: MultihostHost) -> None:
+        super().__init__(host)
+
+    def init(self, label: str, so_pin: str, user_pin: str) -> None:
```
spoore1, line 27: I'd suggest changing the args to set reasonable defaults.   Something like:
```
def init(self, label: str = "sc_test", so_pin: str = "12345678", user_pin: str = "123456") -> None:
```

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,134 @@
...
+                "--id", cert_id
+            ],
+            env={"SOFTHSM2_CONF": self.SOFTHSM2_CONF_PATH}
+        )
+
+    def reset_service(self) -> None:
+        """
+        Restarts the virtual smart card service.
+        """
+        self.host.svc.restart("virt_cacard.service")
```
spoore1, line 96: I get a  mypy warning here:
```
sssd_test_framework/utils/virtual_smartcard.py:104: error: "MultihostHost[Any]" has no attribute "svc"  [attr-defined]
```
I think we need to add svc to the class so we can use it properly.   It can be setup like it is in the utils/sssd.py util class.   It also needs to be added to client to pass self.svc along with self.fs.

sssd_test_framework/roles/client.py
```diff
@@ -80,6 +81,11 @@ def __init__(self, *args, **kwargs) -> None:
         The D-bus destination for infopipe.
         """
 
+        self.smart_card: SmartCardUtils = SmartCardUtils(self.host)
```
spoore1, line 84: To add svc to the class (from a comment below), you'll need to add self.svc to the SmartCardUtils() call.

spoore1 (changes requested):
Thank you for the updates.   Below are some suggested changes that you'll need to address.   Mostly docstrings need to be updated and I believe the svc call for reset needs to be adjusted so it passes the linters.

PR #175, opened 2025-04-25
Uniform admin user declaration

shridhargadekar:
Declaring an administrative user across ad, ipa, samba hosts. This uniform declaration will help to use admin user intuitively in tests

### DIFF
--- a/sssd_test_framework/hosts/samba.py
+++ b/sssd_test_framework/hosts/samba.py
@@ -30,7 +30,7 @@ class SambaHost(BaseLDAPDomainHost, BaseLinuxHost):
 
         self._features: dict[str, bool] | None = None
 
-        self.admin: str = self.config.get("username", "Administrator")
+        self.adminuser: str = self.config.get("username", "Administrator")
         """Username of the admin user, defaults to value of ``Administrator``."""
 
         self.adminpw: str = self.config.get("adminpw", self.bindpw)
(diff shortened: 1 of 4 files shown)


sssd_test_framework/hosts/samba.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, *args, **kwargs) -> None:
 
         self._features: dict[str, bool] | None = None
 
-        self.admin: str = self.config.get("username", "Administrator")
+        self.adminuser: str = self.config.get("username", "Administrator")
```
spoore1, line 33: Do we know if `admin` is used anywhere?   Just adding a new var is easy but, will this change break something else?

spoore1 (commented):
just a question about the Samba admin variable change.

pbrezina (approved):

sssd_test_framework/hosts/samba.py
pbrezina: It is used in the framework in SambaGPO

pbrezina (changes requested):

sssd_test_framework/hosts/samba.py
pbrezina: It is only used in SambaGPO in the framework, I did not found other locations

spoore1 (approved):
LGTM.  I tested this with the sssd system alltests-critical set.

PR #179, opened 2025-04-28
Smartcard

krishnavema:


spoore1 (commented):
Just a quick comment to start. Compared to PR#171, the reset_service, insert and remove card methods are not included here. Was that on purpose?

krishnavema:
> Just a quick comment to start. Compared to PR#171, the reset_service, insert and remove card methods are not included here. Was that on purpose?

sorry !! scoot, I think while reverting  changes they went off I will fix it now

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        Simulates card removal by stopping the smart card service.
+        """
+        self.host.conn.exec(["systemctl", "stop", "virt_cacard.service"])
+
+    def reset_service(self) -> None:
+        """
+        Restarts the virtual smart card service.
+        """
+        self.host.conn.exec(["systemctl", "restart", "virt_cacard.service"])
```
spoore1, line 171: In addition to what you have above, I would suggest adding a function to wrap the setup for local smart card configs into a single helper with something like this:
```
    def setup_local_card(self, client: Client, username: str) -> None:
        """
        Setup local system for smart card authentication

        .. code-block:: python
            :caption: Example usage
            
            @pytest.mark.topology(KnownTopology.Client)
            def test_example(client: Client):
                client.smart_card.setup_local_card(client, 'localuser1')

                result = client.host.conn.run("su - localuser1 -c 'su - localuser1 -c whoami'", input="123456")
                assert "PIN" in result.stderr
                assert "localuser1" in result.stdout

        :param client: client object needed for sssd and other util configurations.
        :type client: Client
        :param username: Name of user to create and configure for local smart card authentication.
        :type username: str
        """
        (key, cert) = self.generate_self_signed_cert()

        self.init()
        self.add_key(key)
        self.add_cert(cert)
        self.reset_service()

        
        client.sssd.sssd["domains"] = "local"
        client.sssd.common.local()
        client.sssd.dom("local")["local_auth_policy"] = "only"
        client.sssd.section(f"certmap/local/{username}")["matchrule"] = "<SUBJECT>.*CN=Test Cert*"
        client.sssd.pam["pam_cert_auth"] = "True"
        client.host.conn.run("echo >> /etc/sssd/pki/sssd_auth_ca_db.pem")
        client.host.conn.run(f"cat {cert} >> /etc/sssd/pki/sssd_auth_ca_db.pem")
        client.sssd.start()

        client.local.user(f"{username}").add()
```

To use this, you'll need to import Client like this I believe:
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from ..roles.client import Client
```

One note about the above is that it should probably have a more descriptive docstring to help with using it later.

spoore1 (changes requested):
This looks good.  I'd remove the extra lines between the param/type lines and also add some kind of wrapper function to help make local setup a little easier in the tests.

thalman (approved):
LGTM, Thanks

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,205 @@
...
+        client.sssd.sssd["domains"] = "local"
+        client.sssd.common.local()
+        client.sssd.dom("local")["local_auth_policy"] = "only"
+        client.sssd.section(f"certmap/local/{username}")["matchrule"] = "<SUBJECT>.*CN=Test Cert*"
+        client.sssd.pam["pam_cert_auth"] = "True"
+        client.host.conn.run("echo >> /etc/sssd/pki/sssd_auth_ca_db.pem")
+        client.host.conn.run(f"cat {cert} >> /etc/sssd/pki/sssd_auth_ca_db.pem")
+        client.sssd.start()
+
+        client.local.user(f"{username}").add()
```
danlavu, line 205: I think this should not be included because you can set up a card and use a local certificate instead of a local account. So, this should remain a setup step.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,205 @@
...
+        Cleans cache directories and prepares the token directory.
+
+        :param label: Token label, defaults to "sc_test"
+        :type label: str, optional
+        :param so_pin: Security Officer PIN, defaults to "12345678"
+        :type so_pin: str, optional
+        :param user_pin: User PIN, defaults to "123456"
+        :type user_pin: str, optional
+        """
+        for path in self.OPENSC_CACHE_PATHS:
```
danlavu, line 43: We have methods in pytest-mh to manipulate files and directories. 

https://pytest-mh.readthedocs.io/en/latest/articles/bundled-utilities/fs.html

danlavu (changes requested):
It looks good, just a few places where you should use the methods provided, instead of ad-hoc commands. 

Couple of other notable things, don't be afraid to modify the actual system files. Do not shove everything in /opt as a test directory.  

All backup/restores should be kept in one place hosts/client.py

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,205 @@
...
+                cert_path,
+                "--type",
+                "cert",
+                "--id",
+                cert_id,
+            ],
+            env={"SOFTHSM2_CONF": self.SOFTHSM2_CONF_PATH},
+        )
+
+    def generate_self_signed_cert(
```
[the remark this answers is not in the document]
spoore1: No, this PR is for very basic capabilities.  We will be adding CA functionality after this.   That will require much more effort overall I think so we want to get a basic set started first.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,205 @@
...
+    "SmartCardUtils",
+]
+
+
+class SmartCardUtils(MultihostUtility[MultihostHost]):
+    """
+    Utility class for managing smart card operations using SoftHSM and PKCS#11.
+    """
+
+    SOFTHSM2_CONF_PATH = "/opt/test_ca/softhsm2.conf"
```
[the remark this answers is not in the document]
spoore1: The virtual smart cards use this in most configurations we've used in recent years (maybe longer).   We can certainly move it but, that will require further updates as well to sssd-ci-containers virtual smart card implementations.   This is just using what should already be configured there.

sssd_test_framework/roles/client.py
```diff
@@ -14,6 +14,7 @@
 from ..utils.sss_override import SSSOverrideUtils
 from ..utils.sssctl import SSSCTLUtils
 from ..utils.sssd import SSSDUtils
+from ..utils.virtual_smartcard import SmartCardUtils
```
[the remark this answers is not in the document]
spoore1: agreed

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,205 @@
...
+        :param user_pin: User PIN, defaults to "123456"
+        :type user_pin: str, optional
+        """
+        for path in self.OPENSC_CACHE_PATHS:
+            self.host.conn.run(f"rm -rf {path}")
+
+        self.host.conn.run(f"rm -rf {self.TOKEN_STORAGE_PATH}")
+        self.host.conn.run(f"mkdir -p {self.TOKEN_STORAGE_PATH}")
+
+        self.host.conn.exec(
```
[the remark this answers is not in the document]
spoore1: ConfigParser or CLIBuilder?

We don't really have any need to run softhsm2-util in other ways than this.  Since the softhsm2-util command here never really changes (other than the passed vars which won't be changing much either), is that really necessary?

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,205 @@
...
+    def generate_self_signed_cert(
+        self,
+        key_path: str = "/tmp/selfsigned.key",
+        cert_path: str = "/tmp/selfsigned.crt",
+        subj: str = "/CN=Test Cert",
+    ) -> tuple[str, str]:
+        """
+        Generates a self-signed certificate and private key.
+
+        :param key_path: Output path for the private key, defaults to "/tmp/selfsigned.key"
```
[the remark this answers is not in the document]
spoore1: IIRC adding certificates and keys to /etc/pki paths allows other apps (like firefox) to access them.   In these tests, we'll want anything using certificates/keys to get them from the virtual smart card via opensc/pcscd.

TBH, I think a tmpdir would probably be better here but I didn't know how to do that as method param default.   Maybe default of None and set it in the method directly using the utils.fs calls?

sssd_test_framework/utils/virtual_smartcard.py
spoore1: You're right.    That should move into the test because it would allow us to cover different scenarios.   Also, that reminds me that the matchrule above should use a parameter too.   I'll add a comment about that too.

sssd_test_framework/utils/virtual_smartcard.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self.init()
+        self.add_key(key)
+        self.add_cert(cert)
+        self.reset_service()
+
+        client.sssd.sssd["domains"] = "local"
+        client.sssd.common.local()
+        client.sssd.dom("local")["local_auth_policy"] = "only"
+        client.sssd.section(f"certmap/local/{username}")["matchrule"] = "<SUBJECT>.*CN=Test Cert*"
```
spoore1, line 199: I think the subject should be passed in like in generate_self_signed_cert().   Then this line could be changed to set matchrule = `f"<SUBJECT>.*{subj}*"`

spoore1 (changes requested):
In addition to Dan's suggested changes, I think the setup_local_card() can be updated to add subject to it's passed parameters.   That will make it more versatile later.

sssd_test_framework/utils/virtual_smartcard.py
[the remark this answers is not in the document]
danlavu: That sounds like a good compromise.

sssd_test_framework/utils/virtual_smartcard.py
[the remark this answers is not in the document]
danlavu: As long as we head in that direction, it doesn't have to be in this PR.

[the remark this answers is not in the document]
danlavu: As long as we head in that direction, it doesn't have to be in this PR.

sssd_test_framework/utils/virtual_smartcard.py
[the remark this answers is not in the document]
danlavu: Ack, marking resolved, not in the scope.

sssd_test_framework/utils/virtual_smartcard.py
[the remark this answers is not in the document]
danlavu: Actually, what does the official documentation say to put it? If it is /etc, having them be in that location is a control to make sure nothing interferes with it.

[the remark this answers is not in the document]
spoore1: I'm not sure we have any "official" documentation we can reference.  Most of what I can find only really reference the CA trust stuff stored in /etc/pki or is really old.   But those do state that it's for shared system certificates/keys.   That's not what we are using these files for in testing though.   We only want them accessible from the virtual smart card.

I think tmpdir is the safest place for this because it should be created and then written to and used from the virtual smart card (and the sssd ca_db file in the case of self-signed certificates) and /etc/sssd/pki/sssd_auth_ca_db.pem for SSSD to trust it.    At that point, we don't really care about the files and we don't want them available via another possible mechanism other than the virtual smart card.

sssd_test_framework/utils/virtual_smartcard.py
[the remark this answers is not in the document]
spoore1: I do too mostly but, I guess I've gotten so used to the exec() dict style that I didn't think about it.

Are we standardizing on run over exec or giving guidance on when to use exec?   Maybe something for the test coding guide?

Do you want the exec()'s in the class changed to run()'s for readability here?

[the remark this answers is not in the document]
danlavu: They're both available, no particular standard. The application is for whatever makes the most sense. exec is especially helpful with CLIBuilder with args and **kwargs, run in this case will make the command way more readable, so yes, let's change it to run.

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,203 @@
...
+    def setup_local_card(self, client: Client, username: str) -> None:
+        """
+        Setup local system for smart card authentication
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.Client)
+            def test_example(client: Client):
+                client.smart_card.setup_local_card(client, 'localuser1')
```
danlavu, line 181: @spoore1, you're right, it does look strange. 

```client.smart_card```, can we make an executive decision and just do ```client.smartcard``` for code and ```smart card``` in docs?

File: sssd_test_framework/utils/smartcard.py [hunk 11]
```diff
@@ -0,0 +1,203 @@
...
+
+                result = client.host.conn.run("su - localuser1 -c 'su - localuser1 -c whoami'", input="123456")
+                assert "PIN" in result.stderr
+                assert "localuser1" in result.stdout
+        """
+        key, cert = self.generate_self_signed_cert()
+
+        self.init()
```
danlavu, line 189: This is nitpicky, but if we are to look at this as steps. 

```
1. generate_self_signed_certs
2. init
3. add_key
4. add_cert
5. reset
```
I think ```init``` is to close to ```__init__```, I propose we rename the methods. 

1. create_certs()
2. setup()
3. add_key
4. add_cert
5. Move it the svc call

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,203 @@
...
+            "login": (self.cli.option.SWITCH, True),
+            "pin": (self.cli.option.VALUE, pin),
+            "write-object": (self.cli.option.VALUE, key_path),
+            "type": (self.cli.option.VALUE, "privkey"),
+            "id": (self.cli.option.VALUE, key_id),
+        }
+        command_line = self.cli.command("pkcs11-tool", args)
+        self.host.conn.run(command_line, env={"SOFTHSM2_CONF": self.SOFTHSM2_CONF_PATH})
+
+    def add_cert(self, cert_path: str, cert_id: str = "01", pin: str = "123456") -> None:
```
danlavu, line 100: we can add another parameter to this function, ```private: bool | None = False```

Then in add_key function, call add_cert with a simple conditional in certs. If I'm not mistaken the only difference is "type". So

```
def add_cert(self, cert_path: str, cert_id: str = "01", pin: str = "123456", private: bool | None = False) -> None:
    ... blah blah ...

    if private: 
      "type": (self.cli.option.Value, "key")
    else:
      "type": (self.cli.option.Value, "cert") 


def add_key(self, key_path: str, key_id: str = "01", pin: str = "123456"):
    add_cert(key, key_id, pin, private=True)
```

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,203 @@
...
+        self.add_cert(cert)
+        self.reset_service()
+
+        client.sssd.sssd["domains"] = "local"
+        client.sssd.common.local()
+        client.sssd.dom("local")["local_auth_policy"] = "only"
+        client.sssd.section(f"certmap/local/{username}")["matchrule"] = "<SUBJECT>.*CN=Test Cert.*"
+        client.sssd.pam["pam_cert_auth"] = "True"
+        client.host.conn.run("echo >> /etc/sssd/pki/sssd_auth_ca_db.pem")
+        client.host.conn.run(f"cat {cert} >> /etc/sssd/pki/sssd_auth_ca_db.pem")
```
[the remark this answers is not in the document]
krishnavema: tried the append the tests are being failed due to that

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,195 @@
...
+        obj_type = "privkey" if private else "cert"
+        args: CLIBuilderArgs = {
+            "module": (self.cli.option.VALUE, "/usr/lib64/pkcs11/libsofthsm2.so"),
+            "login": (self.cli.option.SWITCH, True),
+            "pin": (self.cli.option.VALUE, pin),
+            "write-object": (self.cli.option.VALUE, cert_path),
+            "type": (self.cli.option.VALUE, obj_type),
+            "id": (self.cli.option.VALUE, cert_id),
+        }
+        command_line = self.cli.command("pkcs11-tool", args)
```
danlavu, line 106: Same as the previous comment.

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,195 @@
...
+            "x509": (self.cli.option.SWITCH, True),
+            "nodes": (self.cli.option.SWITCH, True),
+            "sha256": (self.cli.option.SWITCH, True),
+            "days": (self.cli.option.VALUE, "365"),
+            "newkey": (self.cli.option.VALUE, "rsa:2048"),
+            "keyout": (self.cli.option.VALUE, key_path),
+            "out": (self.cli.option.VALUE, cert_path),
+            "subj": (self.cli.option.VALUE, subj),
+        }
+        command_line = self.cli.command("openssl req", args)
```
danlavu, line 150: Same as the previous comment.

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,194 @@
...
+    SOFTHSM2_CONF_PATH = "/opt/test_ca/softhsm2.conf"
+    TOKEN_STORAGE_PATH = "/opt/test_ca/tokens"
+    OPENSC_CACHE_PATHS = [
+        "$HOME/.cache/opensc/",
+        "/run/sssd/.cache/opensc/",
+    ]
+
+    def __init__(self, host: MultihostHost, fs: LinuxFileSystem, svc: SystemdServices) -> None:
+        """
+        :param host: Multihost host.
```
danlavu, line 32: Nitpick, ```:param host: Multihost object.```, we use the word object everywhere else.

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,194 @@
...
+        :type host: MultihostHost
+        :param fs: Filesystem utility instance.
+        :type fs: LinuxFileSystem
+        :param svc: Systemd svc utility instance.
```
danlavu, line 36: nitpick, ```s/instance/object/g```

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.fs: LinuxFileSystem = fs
+        """Filesystem utility used to handle file operations."""
+
+        self.svc: SystemdServices = svc
+        """Systemd utility to manage and interact with svc."""
+
+    def create_certs(self, label: str = "sc_test", so_pin: str = "12345678", user_pin: str = "123456") -> None:
+        """
+        Initializes a SoftHSM token with the given label and PINs.
```
danlavu, line 52: I think this method is poorly named. I'll need @spoore1 to time in. 

It's called ```create_certs```, so it should create the public/private key? Does ```initialize_card```, ```provision_card``` a better description of what this method does?

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,194 @@
...
+        :param key_path: Path to the private key.
+        :type key_path: str
+        :param key_id: Key ID, defaults to "01"
+        :type key_id: str, optional
+        :param pin: User PIN, defaults to "123456"
+        :type pin: str, optional
+        """
+        self.add_cert(cert_path=key_path, cert_id=key_id, pin=pin, private=True)
+
+    def generate_cert(
```
danlavu, line 122: When I see ```create_certs```, I think of this method.

sssd_test_framework/utils/smartcard.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+        self.svc.start("virt_cacard.service")
+
+    def remove_card(self) -> None:
+        """
+        Simulates card removal by stopping the smart card service.
+        """
+        self.svc.stop("virt_cacard.service")
+
+    def setup_local_card(self, client: Client, username: str) -> None:
```
danlavu, line 165: I'd prefer if this moves to ```SSSDCommonConfiguration``` class in utils/sssd.py. It makes more sense to keep SSSD setup with the SSSD stuff and keep smartcard.py with just utilities. 

Setting up SSSD for smartcards could look like this
```client.sssd.common.smardcard()```

WDYT?

danlavu (commented):
Everything looks good, thank you! Just a few nitpicks and some possible refactoring.

sssd_test_framework/utils/smartcard.py
spoore1: Personally, I think of this as a smart card specific method because it does more than just the sssd setup.   It was originally intended as just a helper method to wrap some common smart card setup needed for the first local tests to make the tests shorter.

With that being said, I see your point about moving it to the `SSSDCommonConfiguration` class along with the other common setups.  However, I'm not sure everything in here would belong in a method in that class (e.g. the card/cert/key steps).  And some of the sssd config steps probably won't be needed after we add support for a Proxy provider to the framework.

With that in mind would you still suggest it be moved to `SSSDCommonConfiguration`?   Maybe move the whole thing there and then we adjust that as necessary with future updates (thinking about the certificate authority and proxy provider/topology planned updates)?

spoore1 (changes requested):
Fix Dan's suggestions and let's see what he says about my questions for the setup_local_card() method.

sssd_test_framework/utils/smartcard.py
danlavu: Agreed, either works.

sssd_test_framework/utils/smartcard.py
danlavu: I'm not particularly tied to anything, just a proposal. We'll leave it as is.

danlavu (approved):
I haven't had time to test it, but it looks good, approving. Thank you!

spoore1 (approved):
LGTM

PR #183, opened 2025-04-30
topologies: add keycloak topology

sumit-bose:


sumit-bose:
Hi,

as you can see I added the `kcadm` calls to create an IdP client in Keycloak twice. The main reason is that I haven't found a way to call the method from `class KeycloakIdPClient` in `class KeycloakTopologyController`. If there is a way to do it I think it would be best to only have `class KeycloakIdPClient`. If there is no way, please let me know on which level a class like `class KeycloakIdPClient` should be added to avoid code duplication.

Thanks.

bye,
Sumit

spoore1:
> Hi,
> 
> as you can see I added the `kcadm` calls to create an IdP client in Keycloak twice. The main reason is that I haven't found a way to call the method from `class KeycloakIdPClient` in `class KeycloakTopologyController`. If there is a way to do it I think it would be best to only have `class KeycloakIdPClient`. If there is no way, please let me know on which level a class like `class KeycloakIdPClient` should be added to avoid code duplication.
> 
> Thanks.
> 
> bye, Sumit

@pbrezina would it be better to move KeycloakIdPClient from the role to a utils class for this?   Would that allow us to then reference it from both the Keycloak role and the topology controller without the duplication of code?

pbrezina:
> > Hi,
> > as you can see I added the `kcadm` calls to create an IdP client in Keycloak twice. The main reason is that I haven't found a way to call the method from `class KeycloakIdPClient` in `class KeycloakTopologyController`. If there is a way to do it I think it would be best to only have `class KeycloakIdPClient`. If there is no way, please let me know on which level a class like `class KeycloakIdPClient` should be added to avoid code duplication.
> > Thanks.
> > bye, Sumit
> 
> @pbrezina would it be better to move KeycloakIdPClient from the role to a utils class for this? Would that allow us to then reference it from both the Keycloak role and the topology controller without the duplication of code?

Yes, converting it to MultihostReentrantUtility would allow you to share the code between roles and hosts. That would be cleanest solution, but maybe an overkill for this use case.

You can implement this functionality on the host object and then call this from the KeycloakIdPClient. That might be easier since you don't need the utility setup and teardown code.

pbrezina (approved):

pbrezina:
@sumit-bose I am approving this but if you want to remove the duplications see the suggestions in my previous comment.

sumit-bose:
Hi,

thank you for the suggestions, but if I see it correctly either way would also require to duplicate some parts of `class Keycloak` and `class KeycloakObject`. So I removed the role part for the time being, since it is not needed for the current tests and only keep the topology parts.

bye,
Sumit

spoore1 (approved):
LGTM.   Tested with tests from PRs [#7871](https://github.com/SSSD/sssd/pull/7871) and [#7937](https://github.com/SSSD/sssd/pull/7937):

```
tests/test_idp.py::test_idp__user (keycloak) PASSED
tests/test_idp.py::test_idp__group (keycloak) PASSED
tests/test_idp.py::test_idp__user_groups (keycloak) PASSED
tests/test_idp.py::test_idp__group_members (keycloak) PASSED
tests/test_oidc_child.py::test_oidc_child__get_user (keycloak) PASSED
tests/test_oidc_child.py::test_oidc_child__get_group (keycloak) PASSED
tests/test_oidc_child.py::test_oidc_child__get_user_groups (keycloak) PASSED
tests/test_oidc_child.py::test_oidc_child__get_group_members (keycloak) PASSED

================================================================================== 8 passed, 722 deselected in 452.17s (0:07:32) ==================================================================================
```

pbrezina (approved):

PR #184, opened 2025-04-30
changing client backup restore order

danlavu:
hitting a race condition causing authselect restore to fail because the home directory, containing the backup does not exist.

### DIFF
--- a/sssd_test_framework/hosts/client.py
+++ b/sssd_test_framework/hosts/client.py
@@ -149,13 +149,13 @@ class ClientHost(BaseHost, BaseLinuxHost):
                 fi
             }}
 
+            cp --force --archive "{backup_path}/home/*" /home/ || :
             rm --force --recursive /etc/sssd /var/lib/sss /var/log/sssd /home/*
             restore "{backup_path}/krb5.conf" /etc/krb5.conf
             restore "{backup_path}/krb5.keytab" /etc/krb5.keytab
             restore "{backup_path}/config" /etc/sssd
             restore "{backup_path}/logs" /var/log/sssd
             restore "{backup_path}/lib" /var/lib/sss
-            cp --force --archive "{backup_path}/home/*" /home/ || :
             """,
             log_level=ProcessLogLevel.Error,
         )


danlavu:
```
SKIPPED ../sssd/src/tests/system/tests/test_local.py::test_local__user_authentication (client) - Skipped: unconditional skip
ERROR ../sssd/src/tests/system/tests/test_cli.py::test_cli__is_feature_enabled (client) - pytest_mh._private.errors.TeardownExceptionGroup: One or more error occurred during test teardown (1 sub-exception)
FAILED ../sssd/src/tests/system/tests/test_cli.py::test_cli__is_feature_enabled (client) - pytest_mh.conn.ssh.SSHProcessError: 
Command #376 exited with return code 1:
  Command:
    authselect select local with-mkhomedir --force --backup=multihost.backup
  CWD:
  Env:
  Output:
  Error output:
    [error] Unable to get profile information [2]: No such file or directory
```

pbrezina:
Why do you think it is about home directory? Profile information is read from /usr and backup is stored in /var/lib. There is no home directory involved.

danlavu:
What thing does authselect have to complain about that is a delta and could be missing? Whatever the home directory state was isn't the same when backing up the profile. I haven't tested this, on the principle of undoing changes in the order they were done. I'll make sure it fixes the problem first.

pbrezina:
I'm puzzled. The error `[error] Unable to get profile information [2]: No such file or director` means that profile "local" is not available which may very well be true on fefora41 and lower and rhel9/c9s. The profile is read from `/usr/share/authselect/default`.

`authselect select local with-mkhomedir --force --backup=multihost.backup` does not read or write to /home at all.

sssd_test_framework/hosts/client.py
```diff
@@ -149,13 +149,13 @@ def restore(self, backup_data: Any | None) -> None:
                 fi
             }}
 
+            cp --force --archive "{backup_path}/home/*" /home/ || :
             rm --force --recursive /etc/sssd /var/lib/sss /var/log/sssd /home/*
```
pbrezina, line 153: you are removing /home/* here

pbrezina:
I would like to see some reasoning, why do you think that order here matters? I don't see any reason, why it should matter, the result will still be the same as we only write to /home in this code only at one line. Other restore calls do not touch /home.

sssd_test_framework/hosts/client.py
danlavu: Nice catch, it needs to be moved after that line.

danlavu:
I saw the backup happening last, and that it was being restored last as well. Looking at the error message, while I know it's most likely referring to the backup data in /var  ```  [error] Unable to get profile information [2]: No such file or directory```.

I can't see why that would be missing, but certainly, home didn't have the same contents or subfolders when it was backed up. It's not a good reason, but I wanted it to contain the same content when it restores.

danlavu:
Looking at it again, I misread the code. What I thought it was doing was not correct. This has no impact.

PR #185, opened 2025-05-13
fix(sbus): prevent IndexError in methods with no output args

aborah-sudo:
Skip output parsing when self.output is empty, which happens for void methods like UpdateMemberList. Return None instead of attempting to access values[0].

### DIFF
--- a/sssd_test_framework/utils/sbus.py
+++ b/sssd_test_framework/utils/sbus.py
@@ -177,14 +177,14 @@ class ProxyMethod(ProxyObject):
             raise RuntimeError("Execution failed: no interface.")
 
         objargs = []
-        if self.input is not None:
+        if self.input is not None and len(self.input) > 0:
             for i in range(len(self.input)):
                 obj = self.input[i].mimic()
                 try:
                     obj.value = args[i]
                 except IndexError:
                     raise RuntimeError(
-                        f"Execution failed: {len(self.input)} " f"arguments required but only {i} provided."
+                        f"Execution failed: {len(self.input)} arguments required but only {i} provided."
                     )
                 objargs.append(obj)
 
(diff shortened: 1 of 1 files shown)


sssd_test_framework/utils/sbus.py
```diff
@@ -168,44 +168,43 @@ def __call__(self, *args) -> Any:
         """
         Execute the method
         """
-        # In some internal cases we don't provide the child nor the interface at
-        # at instantiation because we don't use them, but on normal calls they are
-        # required.
```
alexey-tikhonov, line 173: Why do you remove comment?

alexey-tikhonov:
Besides two comments, looks good to me.

aplopez:
Same remarks as Alexey. OK otherwise.

sssd_test_framework/utils/sbus.py
```diff
@@ -168,44 +168,43 @@ def __call__(self, *args) -> Any:
...
                     raise RuntimeError(
-                        f"Execution failed: {len(self.input)} " f"arguments required but only {i} provided."
+                        f"Execution failed: {len(self.input)} arguments required but only {i} provided."
                     )
                 objargs.append(obj)
 
         method = f'{self.interface}.{self.child.attrib["name"]}'
         res = self._run(method, *tuple(objargs))
         if res.rc != 0:
-            raise RuntimeError(f'Execution of \'{self.child.attrib["name"]}{args}\' failed: ' + res.stderr)
```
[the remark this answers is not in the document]
aborah-sudo: Restored now

sssd_test_framework/utils/sbus.py
aborah-sudo: Restored now

PR #187, opened 2025-05-19
Add 'idp-provider' feature and default config

sumit-bose:


sssd_test_framework/hosts/keycloak.py
```diff
@@ -46,6 +46,28 @@ def __init__(self, *args, **kwargs) -> None:
...
+            "idp_userinfo_endpoint",
+            "https://master.keycloak.test:8443/auth/realms/master/protocol/openid-connect/userinfo",
+        )
+        self.client.setdefault(
+            "idp_device_auth_endpoint",
+            "https://master.keycloak.test:8443/auth/realms/master/protocol/openid-connect/auth/device",
+        )
+        self.client.setdefault("idp_id_scope", "profile")
+        self.client.setdefault("idp_auth_scope", "openid profile email")
+
```
alexey-tikhonov, line 70: Doesn't this also need `self.domain = self.client["idp_domain"]` as in 
https://github.com/SSSD/sssd-test-framework/blob/3db73377684e3a697bfc932b1d4b7295c9a33024/sssd_test_framework/hosts/ipa.py#L66
?

sumit-bose: Hi,

if I understand it correctly this is not needed if the client should just use the default domain name `test`. Since there is not strict requirement on the domain name in the IdP provider I think we can just use `test`.

bye,
Sumit

alexey-tikhonov: But this might result in multiple domains named 'test' if some other domain doesn't define non-default name, no?

sumit-bose: Hi,

if I understand it correctly, if you want to change the name if the SSSD domain you have to change the name in the topology definition, e.g.
```
--- a/sssd_test_framework/topology.py
+++ b/sssd_test_framework/topology.py
@@ -123,7 +123,7 @@ class KnownTopology(KnownTopologyBase):
         name="keycloak",
         topology=Topology(TopologyDomain("sssd", client=1, keycloak=1)),
         controller=KeycloakTopologyController(),
-        domains=dict(test="sssd.keycloak[0]"),
+        domains=dict(keycloak="sssd.keycloak[0]"),
         fixtures=dict(
             client="sssd.client[0]",
             keycloak="sssd.keycloak[0]",
```
But I'm not sure if this is expected here. @pbrezina, do you have any recommendations?

bye,
Sumit

sssd_test_framework/hosts/keycloak.py
```diff
@@ -46,6 +46,28 @@ def __init__(self, *args, **kwargs) -> None:
 
         self.adminpw = self.config.get("adminpw", "Secret123")
 
+        self.client.setdefault("id_provider", "idp")
+        self.client.setdefault("enabled", "true")
```
pbrezina, line 50: This shouldn't be needed. The domain is added to "sssd/domains" automatically.

sssd_test_framework/hosts/keycloak.py
```diff
@@ -46,6 +46,28 @@ def __init__(self, *args, **kwargs) -> None:
...
+        self.client.setdefault("idp_type", "keycloak:https://master.keycloak.test:8443/auth/admin/realms/master/")
+        self.client.setdefault("auto_private_groups", "true")
+        self.client.setdefault("use_fully_qualified_names", "true")
```
pbrezina, line 53: Is this needed for all idp provider configurations?

sssd_test_framework/hosts/keycloak.py
```diff
@@ -46,6 +46,28 @@ def __init__(self, *args, **kwargs) -> None:
...
+        self.client.setdefault("debug_level", "9")
```
pbrezina, line 54: This is definitely not needed. We set the debug level automatically when writing configuration.

sssd_test_framework/hosts/keycloak.py
pbrezina: The domain Alexey is referring to is the IPA domain used to realm join, not SSSD domain. Maybe it does not make sense here? Or maybe we can set it to "keycloak.test"?

sssd_test_framework/hosts/keycloak.py
```diff
@@ -46,6 +46,28 @@ def __init__(self, *args, **kwargs) -> None:
 
         self.adminpw = self.config.get("adminpw", "Secret123")
 
+        self.client.setdefault("id_provider", "idp")
+        self.client.setdefault("enabled", "true")
+        self.client.setdefault("idp_type", "keycloak:https://master.keycloak.test:8443/auth/admin/realms/master/")
```
pbrezina, line 51: master.keycloak.test should not be hardcoded. We can use:

```
self.config.setdefault("idp_server", "master.keycloak.test")
```

pbrezina (changes requested):

sssd_test_framework/hosts/keycloak.py
sumit-bose: Removed

sssd_test_framework/hosts/keycloak.py
sumit-bose: I used `{self.hostname}`, I hope this is what you've meant.

sumit-bose:
> Looks good to me, one more thing, we should use the feature detection. It is possible to skip the whole Keycloak topology if condition is not met. See example in https://pytest-mh.readthedocs.io/en/latest/articles/extending/multihost-topologies.html#topology-controller and https://pytest-mh.readthedocs.io/en/latest/api/pytest_mh.html#pytest_mh.TopologyController.skip

Hi,

thanks for the hint, but although the build of the IdP provider is conditional, oidc_child is always build and its test require the Keycloak topology as well. So if I understand it correctly I think there is no need to skip the whole topology.

bye,
Sumit

sssd_test_framework/hosts/keycloak.py
sumit-bose: I would always recommend to use fully-qualified names, that's why I use them in the test as well. But I agree testing both might be even better.

sumit-bose:
Hi,

not sure why the test are failing with errors like

     sssd_test_framework/utils/tools.py:204: error: Argument "name" to "PasswdEntry" has incompatible type "Any | None"; expected "str"  [arg-type]

I only removed the `use_fully_qualified_names` line and didn't even touch the file mentioned.

bye,
Sumit

pbrezina (approved):

PR #188, opened 2025-05-22
Add IPV6 feature

madhuriupadhye:


madhuriupadhye:
Hello @jakub-vavra-cz, I tried the way you suggested for `dual` network in metadata but it wont work, can I add this way or we can configure `ipv6` with IDM-CI hosts?

jakub-vavra-cz (approved):
LGTM, but it will work only with machine from beaker and it might be better to figure out openstack solution as well..

madhuriupadhye:
Hello @pbrezina, can you please check it again? Can I keep it this way?

thalman (changes requested):
See my previous comments

danlavu:
Sorry, I was on the wrong tab, putting this back in review.

PR #194, opened 2025-06-17
Python utility file for the sssd-test-framework that manages socket-a…

aborah-sudo:
…ctivated responders in SSSD

Python utility file for the sssd-test-framework that manages socket-activated responders in SSSD


sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+        result = self.host.conn.exec(
+            ["systemctl", "show", self.socket_unit, "--property=Listen"], raise_on_error=False
+        )
+        if result.rc != 0:
+            raise RuntimeError(f"Failed to get socket path for {self.socket_unit}: {result.stderr}")
+
+        match = re.search(r"Listen=([^\s]+)", result.stdout)
+        return match.group(1) if match else None
+
+    def trigger_socket(self, command: str, args: list[str] = []) -> ProcessResult:
```
alexey-tikhonov, line 144: Imo, if this method is needed at all, it could be intelligent enough to not take any arguments.

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
+"""Manage socket-activated responders in SSSD for testing purposes."""
```
pbrezina, line 1: sssd-sockets.py maybe?

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+        self.host = host
+        self.responder = responder
+        self.socket_unit = f"sssd-{responder}.socket"
+        self.service_unit = f"sssd-{responder}.service"
+
+    def enable_socket(self) -> None:
+        """
+        Enable the responder's socket unit.
+        """
+        result = self.host.conn.exec(["systemctl", "enable", self.socket_unit], raise_on_error=False)
```
pbrezina, line 80: All these calls are already implemented in [SystemdServices](https://pytest-mh.readthedocs.io/en/latest/api/pytest_mh.utils.services.html#pytest_mh.utils.services.SystemdServices) with error handling, logging and restoration. You should pass svc to this class and use these calls.

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+        Get or create an SSSDSocketResponder object for the given responder.
+
+        :param responder: Responder type (e.g., 'nss', 'pam').
+        :type responder: str
+        :return: Responder object.
+        :rtype: SSSDSocketResponder
+        """
+        if responder not in self.responders:
+            self.responders[responder] = SSSDSocketResponder(self.host, responder)
+        return self.responders[responder]
```
pbrezina, line 187: This looks like an overkill, simple iterate the responders here as properties - they can even be lazy if you want, (@cachedproperty) but that's probably not needed. Then you can remove all the checking code.

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def remove_services_from_sssd_conf(self) -> None:
+        """
+        Remove the 'services' line from sssd.conf to test socket activation.
+        """
+        result = self.host.conn.exec(
+            ["sed", "-i", "/^services[[:space:]]*=.*$/d", "/etc/sssd/sssd.conf"], raise_on_error=False
```
alexey-tikhonov, line 194: I don't think entire line 'services' should be removed, rather socket activated responders.

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+        result = self.host.conn.exec(
+            ["sed", "-i", f"/\\[sssd\\]/a services = {services}", "/etc/sssd/sssd.conf"], raise_on_error=False
+        )
+        if result.rc != 0:
+            raise RuntimeError(f"Failed to restore services line to sssd.conf: {result.stderr}")
+
+        # Restart SSSD to apply changes
+        self.host.conn.exec(["systemctl", "restart", "sssd.service"])
+
+    def verify_socket_activation(self, responder: str, test_command: str, test_args: list[str] = []) -> bool:
```
pbrezina, line 218: This looks very much like a full blown test (not something that is used to support arbitrary test), maybe it would be better to place it into a test instead?

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        if responder not in [
+            ResponderType.NSS,
+            ResponderType.PAM,
+            ResponderType.SUDO,
+            ResponderType.SSH,
+            ResponderType.PAC,
+            ResponderType.AUTOFS,
+        ]:
+            raise ValueError(f"Unknown responder: {responder}")
```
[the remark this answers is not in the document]
pbrezina: This can be removed, see my comments to get_responder().

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+        :type command: str
+        :param args: Arguments for the command.
+        :type args: list[str]
+        :return: Process result of the command.
+        :rtype: ProcessResult
+        """
+        result = self.host.conn.exec([command, *args], raise_on_error=False)
+        if result.rc != 0 and "Connection refused" in result.stderr:
+            raise RuntimeError(f"Socket activation failed for {self.responder}: {result.stderr}")
+        return result
```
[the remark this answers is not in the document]
pbrezina: This is very important, you can't default to [], always default to None

pbrezina (changes requested):
Hi, see my comments above. It's pretty good, but we can simplify and reuse other code.

But do we really need this? How often are we going to access individual sockets? Maybe it would be enough to use what we already have, eg. `client.svc.start("sssd-nss.socket")`?

I would personally incline to use `client.svc` and add configuration with blank "services" line in sssd.conf to SSSDCommonConfiguration.

aborah-sudo:
> @aborah-sudo, could you please provide a link to the test example using this API?

https://paste.centos.org/view/5e10f090

sssd_test_framework/roles/client.py
```diff
@@ -91,6 +92,10 @@ def __init__(self, *args, **kwargs) -> None:
         """
         The D-bus destination for infopipe.
         """
+        self.socket: SSSDSocketUtils = SSSDSocketUtils(self.host)
```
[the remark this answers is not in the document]
aborah-sudo: Added

sssd_test_framework/utils/socket.py
aborah-sudo: renamed to sssd_sockets.py

sssd_test_framework/utils/socket.py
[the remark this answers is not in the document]
aborah-sudo: removed

sssd_test_framework/utils/socket.py
```diff
@@ -0,0 +1,245 @@
...
+            ResponderType.AUTOFS,
+        ]:
+            raise ValueError(f"Unknown responder: {responder}")
+
+        self.host = host
+        self.responder = responder
+        self.socket_unit = f"sssd-{responder}.socket"
+        self.service_unit = f"sssd-{responder}.service"
+
+    def enable_socket(self) -> None:
```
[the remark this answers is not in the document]
aborah-sudo: removed socket word

sssd_test_framework/utils/socket.py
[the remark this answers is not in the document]
aborah-sudo: this function is removed as per Alexey's review

sssd_test_framework/utils/socket.py
aborah-sudo: removed

sssd_test_framework/utils/socket.py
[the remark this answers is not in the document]
aborah-sudo: removed from this file

sssd_test_framework/utils/socket.py
aborah-sudo: changed

sssd_test_framework/utils/socket.py
aborah-sudo: All of the methods are not available via svc. So not implementing

aborah-sudo:
> Hi, see my comments above. It's pretty good, but we can simplify and reuse other code.
> 
> But do we really need this? How often are we going to access individual sockets? Maybe it would be enough to use what we already have, eg. `client.svc.start("sssd-nss.socket")`?
> 
> I would personally incline to use `client.svc` and add configuration with blank "services" line in sssd.conf to SSSDCommonConfiguration.

Hi, the other method implemented this file is not available via client.svc 

currently we do not have any socket related test/library in our repo, this will be the one and the upcoming socket related tests. We are already able to find bugs with https://github.com/SSSD/sssd/issues/8005 

 In future we need one dedicated library for socket

jakub-vavra-cz (changes requested):
I do not see a backup + teardown code that remembers and undoes changes to to systemd.  Init should probably get the settings of all of the responders (enabled, active) and teardown should restore that state. 

Alternatively teardown can just set the expected state of the responders.

I take that some other part of the framework undoes changes to sssd .conf so it is not needed to be done here.

sssd_test_framework/utils/socket.py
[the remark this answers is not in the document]
alexey-tikhonov: IIRC, as it was implemented it was a test: method was triggering a responder and making sure it's started.

There should be logical distinction between a "facility" and a "test".

Code to configure SSSD - is a "facility"; code to make sure responder starts automatically when expected - is a "test".

sssd_test_framework/utils/sssd_sockets.py
```diff
@@ -0,0 +1,278 @@
...
+            ResponderType.NSS,
+            ResponderType.PAM,
+            ResponderType.SUDO,
+            ResponderType.SSH,
+            ResponderType.PAC,
+            ResponderType.AUTOFS,
+        ]
+
+        # Extract current services line
+        result = self.host.conn.exec(["grep", "^services[[:space:]]*=", "/etc/sssd/sssd.conf"], raise_on_error=False)
```
alexey-tikhonov, line 231: I think there is a dedicated class to manage sssd config.
Note that there are might be snippets, etc.
At the very least a path to config file shouldn't be hardcoded.

jakub-vavra-cz (approved):
LGTM

aborah-sudo:
> Sorry, I still incline to use the already available systemd service class, it already does everything for you in much more thorough way then what is implemented here. The only thing missing is "is-active", that can be easily implemented.
> 
> You do not need to enable or disable the sockets, just start and stop. Enabling and disabling does not take any effect unless you reboot the system.

I agree when we all agree @alexey-tikhonov @jakub-vavra-cz what do you say about this ?

alexey-tikhonov:
> > Sorry, I still incline to use the already available systemd service class, it already does everything for you in much more thorough way then what is implemented here. The only thing missing is "is-active", that can be easily implemented.

+1

> > You do not need to enable or disable the sockets, just start and stop. Enabling and disabling does not take any effect unless you reboot the system.
> 
> I agree when we all agree @alexey-tikhonov @jakub-vavra-cz what do you say about this ?

aborah-sudo:
> Sorry, I still incline to use the already available systemd service class, it already does everything for you in much more thorough way then what is implemented here. The only thing missing is "is-active", that can be easily implemented.
> 
> You do not need to enable or disable the sockets, just start and stop. Enabling and disabling does not take any effect unless you reboot the syst

> > > Sorry, I still incline to use the already available systemd service class, it already does everything for you in much more thorough way then what is implemented here. The only thing missing is "is-active", that can be easily implemented.
> 
> +1
> 
> > > You do not need to enable or disable the sockets, just start and stop. Enabling and disabling does not take any effect unless you reboot the system.
> > 
> > 
> > I agree when we all agree @alexey-tikhonov @jakub-vavra-cz what do you say about this ?

As per agreement i have modified this PR and added another PR : https://github.com/next-actions/pytest-mh/pull/108

sssd_test_framework/utils/sssd.py
```diff
@@ -1039,3 +1039,28 @@ def pam(self, features: list[str] | None = None) -> None:
...
+
+        # Set empty services to enable socket activation
+        self.sssd.sssd["services"] = ""
+
+        # Ensure socket units are enabled and started
+        for responder in responders:
+            socket_unit = f"sssd-{responder}.socket"
+
+            # Start the socket unit (raises on error by default)
+            self.sssd.host.conn.run(f"systemctl start {socket_unit}")
```
pbrezina, line 1066: This should use the svc utility, it will make sure the change is reverted before starting other test and it will also print all available logs on an error.

pbrezina (changes requested):
Hi, this is waay simpler, thank you. I merged the pytest-mh pull request.

pbrezina (approved):

jakub-vavra-cz (approved):
LGTM

PR #198, opened 2025-08-01
Add utility for issuing and managing IPA certificates

krishnavema:
Implements a reusable utility for IPA Certificate Authority integration:

- Automated certificate issuance with CSR and key generation (OpenSSL-based)
- Support for revocation workflows using RFC 5280-compliant reason codes
- Certificate hold and unhold operations (cert-revoke, cert-remove-hold)
- Retrieval and parsing of certificate metadata via ipa cert-show


sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,276 @@
...
+
+        if add_service:
+            self.host.conn.run(f"ipa service-add {shlex.quote(principal)}", raise_on_error=False)
+
+        cmd = (
+            f"ipa cert-request {shlex.quote(csr_path)} "
+            f"--principal={shlex.quote(principal)} "
+            f"--validity={validity_days}"
+        )
+        result = self.host.conn.run(cmd, raise_on_error=False)
```
spoore1, line 108: You can add the --certificate option to the ipa cert-request command so it writes the certificate file out to a known location.   Might be less work and more reliable than reading it from stdout.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,276 @@
...
+        if result.rc != 0:
+            raise RuntimeError(f"IPA cert-request failed: {result.stderr}")
+
+        m = re.search(
+            r"certificate:\s*(-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----)",
+            result.stdout,
+            re.DOTALL,
+        )
+        if not m:
+            raise RuntimeError(f"Failed to parse certificate from IPA response: {result.stdout}")
```
spoore1, line 119: If you add --certificate option to ipa cert-request, I think this section is unnecessary.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,276 @@
...
+
+        return cert_path, key_path, csr_path
+
+    def revoke(self, cert_path: str, reason: RevocationReason = "unspecified") -> None:
+        """
+        Revoke a certificate with an RFC 5280 revocation reason.
+
+        :param cert_path: Remote certificate file path
+        :type cert_path: str
+        :param reason: Revocation reason
```
spoore1, line 132: param should mention the default if it has one like reason does here.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,276 @@
...
+            else:
+                hostname = principal.split("@")[0]
+
+            if not hostname:
+                raise ValueError(f"Cannot derive subject from principal '{principal}'")
+            subject = f"/CN={hostname}"
+
+        if add_service:
+            self.host.conn.run(f"ipa service-add {shlex.quote(principal)}", raise_on_error=False)
+
```
spoore1, line 102: You need to add _generate_cert() before the ipa cert-request.   something like this I think:
```
self._generate_csr(key_path, csr_path, subject)
```

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,276 @@
...
+        :type output: str
+        """
+        info = {}
+        for line in output.splitlines():
+            if ":" in line:
+                key, value = line.split(":", 1)
+                info[key.strip()] = value.strip()
+        return info
+
+    def setup_service_management(
```
spoore1, line 251: While this might be useful for other areas of testing, I'm not sure if it'll be needed anywhere for smart card testing.   Did you find this setup in the old tests somewhere?

krishnavema: just took reference of smartcard util

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,236 @@
...
+            subject: Optional OpenSSL subject (e.g., `/CN=example`). If omitted, derived from principal.
+            add_service: Whether to add the principal as an IPA service.
+            key_size: RSA key size in bits.
+
+        Returns:
+            A tuple of (certificate_path, key_path, csr_path).
+
+        Raises:
+            ValueError: If subject cannot be derived from principal.
+            RuntimeError: If CSR generation fails.
```
spoore1, line 85: I think we're supposed to list exceptions to raise like this:
```
    :raises ValueError: If subject cannot be derived from principal.
    :raises RuntimeError.: If CSR generation fails.
```
I think you can see examples of this in utils/sssd.py

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,236 @@
...
+            f"rsa:{key_size}",
+            "-nodes",
+            "-keyout",
+            key_path,
+            "-out",
+            csr_path,
+            "-subj",
+            subject,
+        ]
+        cmdline = " ".join(shlex.quote(p) for p in cmd)
```
spoore1, line 196: This appears to be working for me now but, I forgot that you'd used CLIBuilderArgs in the previous smartcard PR:
```
        args: CLIBuilderArgs = {
            "x509": (self.cli.option.SWITCH, True),
            "nodes": (self.cli.option.SWITCH, True),
            "sha256": (self.cli.option.SWITCH, True),
            "days": (self.cli.option.VALUE, "365"),
            "newkey": (self.cli.option.VALUE, "rsa:2048"),
            "keyout": (self.cli.option.VALUE, key_path),
            "out": (self.cli.option.VALUE, cert_path),
            "subj": (self.cli.option.VALUE, subj),
        }
        self.host.conn.run(self.cli.command("openssl req", args))
        return key_path, cert_path
```

Can you try to put CLIBuilderArgs back and see if changing the run to use "openssl req" like you did there?   I'm not quite clear how CLIBuilderArgs determines if it should use one or two hyphens but, it seems like it does that correctly in the smartcard util.

spoore1 (changes requested):
I'd suggest putting the param/type type-hint docstrings back to match the other code in the framework.  Also, I need to apologize for pushing away from CLIBuilderArgs when I ran into the issue.   I forgot you used it with the openssl command in the previous smartcard PR so it should work.   If so, that would be best.

My test is passing so functionally, it looks good so far.   Just some minor things to clean up I think before we request more Reviewers.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,264 @@
...
+
+        :param cert_path: Path to the certificate file.
+        :type cert_path: str
+        :raises RuntimeError: If hold removal fails.
+        """
+        serial = self._get_cert_serial(cert_path)
+        args: CLIBuilderArgs = {"serial": (self.cli.option.VALUE, serial)}
+        result = self.host.conn.run(self.cli.command("ipa cert-remove-hold", args), raise_on_error=False)
+        if result.rc != 0:
+            raise RuntimeError(f"ipa cert-remove-hold failed: {result.stderr}")
```
sumit-bose, line 160: Hi,

I'm fine with the patch in general. I just have one question about the purpose of  `revoke_hold_remove()` because `ipa cert-remove-hold` will return an error if there is no hold on the certificate and hence `revoke_hold_remove()` will raise an exception.

If the purpose is just to make sure that there is no hold on the certificate it might be better to check the status of the certificate and only call `ipa cert-remove-hold` if there is a hold. 

If the purpose is to remove a hold from a certificate which is known to have a hold then no change is needed.

bye,
Sumit

krishnavema: The purpose of revoke_hold_remove() is to explicitly remove a hold from a certificate that is already known to be on hold

sumit-bose (approved):
Hi,

thank you for the patch, ACK.

bye,
Sumit

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,264 @@
...
+            "serial": (self.cli.option.VALUE, serial),
+            "revocation-reason": (self.cli.option.VALUE, str(reason_code)),
+        }
+        result = self.host.conn.run(self.cli.command("ipa cert-revoke", args), raise_on_error=False)
+        if result.rc != 0:
+            raise RuntimeError(f"IPA cert-revoke failed: {result.stderr}")
+
+    def revoke_hold(self, cert_path: str) -> None:
+        """
+        Place a certificate on hold.
```
danlavu, line 141: It's unclear to me what the difference is between revoke and hold. Isn't revoking inherently putting the certificate on hold? Or are you putting a hold on the revoke? 

Per this definition, 
```
In PKI, revoking a certificate means it is permanently invalidated and cannot be used, while placing a certificate on hold indicates a temporary status where it may be reinstated later if the issue is resolved. The hold status allows for the possibility of un-revoking the certificate, whereas revocation is irreversible.
``` 

It sounds like two different actions; is revoke the same as invalidate?

Does it make sense to just call this method hold? and the following method, remove_hold() 

Most of the methods are named with the verb first then noun, i.e. add_certificate, remove_certificate, I think the naming should follow this.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,264 @@
...
+    def _parse_cert_info(self, output: str) -> Dict[str, str]:
+        """
+        Parse ipa cert-show output into a dictionary.
+
+        :param output: Raw command output from ipa cert-show.
+        :type output: str
+        :returns: Dictionary of certificate attributes.
+        :rtype: dict[str, str]
+        """
+        info: Dict[str, str] = {}
```
danlavu, line 259: Use the built-in dict type.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,264 @@
...
+        for line in (output or "").splitlines():
```
danlavu, line 260: See if you can re-use parse_attr from misc/__init__.py it does take output that is formatted "atttribute: value" and returns a dictionary.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,264 @@
...
+        :type cert_path: str
+        :returns: The certificate serial number as a lowercase hex string.
+        :rtype: str
+        :raises RuntimeError: If serial extraction fails.
+        """
+        cmd = ["openssl", "x509", "-in", cert_path, "-noout", "-serial"]
+        cmdline = " ".join(shlex.quote(p) for p in cmd)
+        result = self.host.conn.run(cmdline, raise_on_error=False)
+        if result.rc != 0:
+            raise RuntimeError(f"Failed to get serial from certificate: {result.stderr}")
```
danlavu, line 221: nitpick add an ! at the end of the error message.

sssd_test_framework/utils/ipa_ca.py
```diff
@@ -0,0 +1,264 @@
+from __future__ import annotations
+
+import os
+import re
+import shlex
+import uuid
+from typing import Dict, Literal, Optional, Tuple
```
danlavu, line 7: Please use the built-in Tuple and Dict.

danlavu (changes requested):
Overall, it looks really good, just a few implementation questions.

spoore1 (approved):
Ran test with updates moving the code to the IPA role and it's still working as expected.

sssd_test_framework/roles/ipa.py
```diff
@@ -2235,3 +2271,228 @@ def delete(self) -> None:
...
+        :param host: Remote test host.
+        :type host: MultihostHost
+        :param fs: Filesystem helper for remote file operations.
+        :type fs: LinuxFileSystem
+        """
+        self.host = host
+        self.fs = fs
+        self.cli: CLIBuilder = host.cli
+        self.temp_dir = f"/tmp/ipa_test_certs_{os.getpid()}_{uuid.uuid4().hex}"
+        self.fs.mkdir_p(self.temp_dir, mode="700")
```
danlavu, line 2304: ```self.fs.backup(self.temp_dir)``` should reset the state of the directory. If it doesn't exist, it'll be deleted.

danlavu (changes requested):
It looks good except for some minor changes in-line. After the teardown is updated. I'll test the code. Do you have a test I can run?

spoore1 (approved):
LGTM.

PR #199, opened 2025-08-07
Adcli, adjust parameter

shridhargadekar:
Rearranging domain parameter, it should be keyword-argument.

### DIFF
--- a/sssd_test_framework/roles/ad.py
+++ b/sssd_test_framework/roles/ad.py
@@ -70,6 +70,16 @@ class AD(BaseWindowsRole[ADHost]):
         Kerberos realm.
         """
 
+        self.name: str = "ad"
+        """
+        Generic provider name.
+        """
+
+        self.server: str = self.host.hostname
+        """
+        Generic server name.
+        """
+
         self.auto_ou: dict[str, bool] = {}
         """Organizational units that were automatically created."""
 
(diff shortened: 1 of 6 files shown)


sssd_test_framework/utils/adcli.py
```diff
@@ -30,7 +30,7 @@ def test_adcli_join(client: Client, provider: GenericADProvider):
        kerberos based authentication.
     """
 
-    def info(self, domain: str, *, args: list[str] | None = None) -> ProcessResult:
+    def info(self, *, domain: str, args: list[str] | None = None) -> ProcessResult:
```
jakub-vavra-cz, line 33: You might want to have the verbose configurable to reduce the output (noise) in the future and possibly to test verbosity related bugs.

```suggestion
    def info(self, *, domain: str, verbose: bool=True, args: list[str] | None = None) -> ProcessResult:
```

jakub-vavra-cz (changes requested):
Please make sure that --verbose can be switched off. I expect that tests that will not be testing the function itself might want to have less noise in the output and we might want to test the non-verbose mode in the future. (Like making sure that the error messages on failures are helpful even without the verbose output).

thalman:
> > I do not understand what you are trying to solve/fix here. What is the benefit?
> 
> Trying to have a uniform parameter definition for `domain` option. The `adcli` command uses `domain` option in different ways. With existing code, `domain` value gets duplicated if used with subcommands like `create-user`, `delete-user` etc.
> 
> From `man adcli` ` SYNOPSIS
> 
> ```
>    adcli info domain.example.com
> 
>    adcli join domain.example.com
> 
>    adcli update
> 
>    adcli testjoin
> 
>    adcli create-user [--domain=domain.example.com] user
> 
>    adcli delete-user [--domain=domain.example.com] user
> ```
> 

OK, I got the idea and I agree with the code change. I would appreciate more verbose commit message like `some adcli sub-commands like create-user does not need the domain parameter. So far this parameter was mandatory in our framework and thus redundant in such cases. Moving it to keywords make the usage of adcli class cleaner/more readable/etc..`.

But this is not mandatory. Up to you.

PR #200, opened 2025-08-27
Revert "creating symlink for dsctl binary"

danlavu:
Reverts SSSD/sssd-test-framework#196

This was a temporary fix.


spoore1 (approved):
LGTM.   Manually tested and confirmed location of dsctl command:
```
[root@master ~]# ls -l /usr/sbin/dsctl
-rwxr-xr-x. 1 root root 5799 Aug  5 00:00 /usr/sbin/dsctl
```

PR #202, opened 2025-09-11
ci: run SSSD tests to avoid breaking it

pbrezina:
Running only against one distribution is enough, since we are not
testing SSSD, just the framework.


justin-stephenson (dismissed):
Ack, CI run shows this would have found the issue https://github.com/SSSD/sssd-test-framework/actions/runs/17643092872/job/50134480404?pr=202

ikerexxe (commented):
Just a minor comment inline

ikerexxe (approved):
LGTM!

PR #207, opened 2025-09-24
Hostnameutils: import hostnameutils

shridhargadekar:
Incorporate hostnameutils tools


sssd_test_framework/roles/base.py
```diff
@@ -189,6 +190,11 @@ def __init__(self, *args, **kwargs) -> None:
         Authentication helpers.
         """
 
+        self.hostnameutils: HostnameUtils = HostnameUtils(self.host)
```
pbrezina, line 193: maybe add `.postpone_setup()` to avoid running __enter__ code if the utility is not used

jakub-vavra-cz (approved):
LGTM>

PR #209, opened 2025-10-06
Add Certificate Authority functionality for AD

krishnavema:
Implement certificate authority for AD

### DIFF
--- a/sssd_test_framework/roles/ad.py
+++ b/sssd_test_framework/roles/ad.py
@@ -2,9 +2,12 @@
 
 from __future__ import annotations
 
+import os
+import re
+import uuid
 from abc import ABC
 from datetime import datetime
-from typing import Any, TypeAlias
+from typing import Any, Literal, TypeAlias
 
 from pytest_mh.cli import CLIBuilderArgs
 from pytest_mh.conn import ProcessResult
@@ -30,6 +33,19 @@ __all__ = [
     "GPO",
     "ADDNSServer",
     "ADDNSZone",
+]
+
+RevocationReason = Literal[
+    "unspecified",
+    "key_compromise",
+    "ca_compromise",
+    "affiliation_changed",
+    "superseded",
+    "cessation_of_operation",
+    "certificate_hold",
+    "remove_from_crl",
+    "privilege_withdrawn",
+    "aa_compromise",
 ]
 
 
@@ -139,6 +155,35 @@ class AD(BaseWindowsRole[ADHost]):
                         'keys': [str(key3)]
                     },
                 }
+        """
+        self.ca = ADCertificateAuthority(self.host)
+        """
+        AD Certificate Authority management.
+        Provides certificate operations:
+        - Request certificates using templates
+        - Request smartcard certificates with Enrollment Agent
+        - Revoke certificates with configurable reasons
+        - Manage certificate holds
+        - Export certificates as PFX
+        - Retrieve certificate and template details
+        Example:
+            # Request a user certificate
+            cert, key, csr = ad.ca.request(
+                template="User",
+                subject="CN=testuser"
+            )
+            # Request a smartcard certificate
+            cert, key, csr = ad.ca.request_smartcard(
+                template="SmartcardLogon",
+                subject="CN=Smartcard User",
+                enrollment_agent_cert_hash="thumbprint_here",
+                requester_name="AD\\Administrator"
+            )
+
+            # Revoke certificate
+            ad.ca.revoke(cert, reason="key_compromise")
+            # Export as PFX
+            ad.ca.export_pfx(cert, "C:\\temp\\cert.pfx")
         """
 
     @property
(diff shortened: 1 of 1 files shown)


sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2376,396 @@ def print(self) -> str:
...
+        """
+
+        self.host.conn.run(
+            f"""
+            $content = @"
+            {inf_content}
+            "@
+            Set-Content -Path "{inf_path}" -Value $content
+            """
+        )
```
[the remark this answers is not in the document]
krishnavema: I have explored but I can find only Linux fs there , I can't find any Windows fs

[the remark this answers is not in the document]
spoore1: Ok, ignore then.

sssd_test_framework/roles/ad.py
```diff
@@ -140,6 +156,45 @@ def test_example_autofs(client: Client, ad: AD, nfs: NFS):
                     },
                 }
         """
+        self.ca = ADCertificateAuthority(self.host)
+        """
+        AD Certificate Authority management.
+
+
+        Provides certificate operations:
+
```
danlavu: nitpick, indentation, no blank lines needed.

sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2386,393 @@ def print(self) -> str:
 
 
 ADNetgroupMember: TypeAlias = LDAPNetgroupMember[ADUser, ADNetgroup]
+
+
+class ADCertificateAuthority:
+    """
+    Provides helper methods for Active Directory Certificate Authority operations.
```
danlavu: Nitpick, ```Certificate Authority server management.```

sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2386,393 @@ def print(self) -> str:
...
+
+    def __init__(self, host: ADHost) -> None:
+        """
+        Initialize the AD Certificate Authority helper.
+
+        :param host: Remote AD host.
+        :type host: ADHost
+        """
+        self.host = host
+        self.cli = host.cli
```
danlavu: We need to clean up the folder on teardown. The easiest thing to do, I think would be to rename the directory to something unique c:\pki then in hosts\ad.py, delete the directory in ```def restore```

sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2386,393 @@ def print(self) -> str:
...
+            "superseded": 4,
+            "cessation_of_operation": 5,
+            "certificate_hold": 6,
+            "remove_from_crl": 8,
+            "privilege_withdrawn": 9,
+            "aa_compromise": 10,
+        }
+        return reason_map[reason]
+
+    def _parse_cert_info(self, output: str) -> dict[str, list[str]]:
```
danlavu: I think the parsers should go in misc/__init__.py

spoore1 (commented):
Another glance and it's looking very good.   Just a question about the PSIni module calls you have in the request methods.   I can't seem to find those.

sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2395,387 @@ def print(self) -> str:
...
+
+        :param template: Certificate template name (e.g., "User").
+        :type template: str
+        :param subject: Certificate subject (e.g., "CN=testuser").
+        :type subject: str
+        :param key_size: RSA key size in bits.
+        :type key_size: int
+        :returns: A tuple of (certificate_path, key_path, csr_path).
+        :rtype: tuple[str, str, str]
+        :raises RuntimeError: If certificate request fails.
```
danlavu: This is the first time I'm seeing the RunTimeError docstring, I think it's a great idea, thanks.

danlavu (commented):
This is great, nitpicks and missing unit tests for the misc functions.

sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2394,387 @@ def print(self) -> str:
...
+
+    This class allows requesting, revoking, placing/removing certificate holds,
+    and retrieving certificate information via certreq and certutil commands.
+    """
+
+    def __init__(self, host: ADHost) -> None:
+        """
+        Initialize the AD Certificate Authority helper.
+
+        :param host: Remote AD host.
```
danlavu: Nitpick, all the hosts are remote, so ```AD host.``` is sufficient.

sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2394,538 @@ def print(self) -> str:
...
+        :type subject: str
+        :param password: Password for PFX export, defaults to "Secret123"
+        :type password: str, optional
+        :param key_size: RSA key size in bits, defaults to 2048
+        :type key_size: int, optional
+        :return: A tuple of (certificate_path, pfx_path, csr_path).
+        :rtype: tuple[str, str, str]
+        :raises RuntimeError: If certificate request fails.
+        """
+        return self.__request_enrollment(template, subject, password, key_size)
```
spoore1: I'm not sure I understand why request_basic() would call __request_enrollment().   I thought the latter was just for requesting "Enrollment Agent" certificates to be used to request a smart card certificate using "Enrollment On Behalf Of" functionality since we're using Administrator to request certificates.

sssd_test_framework/roles/ad.py
```diff
@@ -2331,3 +2394,538 @@ def print(self) -> str:
...
+KeySpec = 1
+KeyUsage = 0xA0
+Exportable = TRUE
+RequestType = CMC
+RequesterName = \"{requester_name}\"
+ProviderName = \"Microsoft Software Key Storage Provider\"
+
+[RequestAttributes]
+CertificateTemplate = \"{template}\"
+"@
```
spoore1: Can these lines not be indented?   I thought the conn.run() would handle indentation properly as long as it was indented along with the `$infContent` first line.

sssd_test_framework/roles/ad.py
```diff
@@ -2315,3 +2395,612 @@ def print(self) -> str:
...
+        This method signature matches the IPA CA interface for consistency.
+
+        :param template: Smartcard certificate template name (e.g., "SmartcardLogon").
+        :type template: str
+        :param subject: Certificate subject (e.g., "CN=testuser").
+        :type subject: str
+        :return: A tuple of (certificate_path, pfx_path, csr_path).
+        :rtype: tuple[str, str, str]
+        :raises RuntimeError: If smartcard certificate request fails.
+        """
```
spoore1: First let's get base_dn and user_dn to make sure we have the correct values we'll need later:
```
        base_dn =  ",".join(f"dc={part}" for part in self.host.domain.split(".") if part)
        user_dn = subject
        if base_dn.lower() not in subject.lower():
            user_dn = f"{subject},CN=Users,{base_dn}"
```

sssd_test_framework/roles/ad.py
```diff
@@ -2315,3 +2395,612 @@ def print(self) -> str:
...
+        req_path = os.path.join(self.temp_dir, f"{base}.req")
+        signed_req_path = os.path.join(self.temp_dir, f"{base}_signed.req")
+        cert_path = os.path.join(self.temp_dir, f"{base}.cer")
+        pfx_path = os.path.join(self.temp_dir, f"{base}.pfx")
+
+        self.host.conn.run(
+            f"""
+$infContent = @"
+[NewRequest]
+Subject = "{subject}"
```
spoore1: Change `{subject}` to `{user_dn}`

sssd_test_framework/roles/ad.py
```diff
@@ -2315,3 +2395,612 @@ def print(self) -> str:
...
+        :param cert_path: Path to certificate file.
+        :type cert_path: str
+        :return: Certificate thumbprint.
+        :rtype: str
+        :raises RuntimeError: If thumbprint extraction fails.
+        """
+        result = self.host.conn.run(f"(Get-PfxCertificate -FilePath {cert_path}).Thumbprint", raise_on_error=False)
+        if result.rc != 0:
+            raise RuntimeError("Thumbprint not found in certificate output")
+        return result.stdout.strip()
```
spoore1: This wasn't working for me in my tests but, I was able to replace it with this which did work:
```
result = self.host.conn.run(f"(Get-PfxCertificate -FilePath {cert_path}).Thumbprint", raise_on_error=False)
```

spoore1 (changes requested):
Ok, after a lot of digging and working through some test environment issues, I got a good test with the following changes I suggest.   I'll send more info directly on some test env setup changes you'll need to make as well.

sssd_test_framework/roles/ad.py
```diff
@@ -2315,3 +2379,656 @@ def print(self) -> str:
...
+
+[RequestAttributes]
+CertificateTemplate = "EnrollmentAgent2"
+"@
+Set-Content -Path "{inf_path}" -Value $infContent
+""")
+
+        result = self.host.conn.run(f'certreq -q -new "{inf_path}" "{req_path}"', raise_on_error=False)
+        if result.rc != 0:
+            return
```
spoore1: Considering that the default behavior for a host.conn.run() is to throw an exception if it fails, do we need to bother checking result.rc here (and in some other places) and returning from the method?   Or can we drop this extra check to shorten and simplify the code some?    

This is a question for @danlavu as well.   Do we prefer to write out specific checks and error messages or leave the run()'s to raise an exception if it fails?    I can definitely see some benefit in some of the explanations and error messages but, I'm not sure if we have a preference here.

sssd_test_framework/roles/ad.py
```diff
@@ -2315,3 +2379,656 @@ def print(self) -> str:
...
+
+        templates_result = self.host.conn.run("certutil -catemplates", raise_on_error=False)
+        if "EnrollmentAgent" not in templates_result.stdout:
+            return
```
spoore1: Seems like if EnrollmentAgent isn't there, you probably want to fail this method don't you?   What do we gain by silently returning if the method can't issue the certificate?

sssd_test_framework/roles/ad.py
```diff
@@ -159,6 +185,45 @@ def test_example(client: Client, ad: AD):
...
+        - Revoke certificates with configurable reasons
+        - Manage certificate holds
+        - Export certificates as PFX
+        - Retrieve certificate and template details
+
+        .. code-block:: python
+            :caption: Example usage
+
+            @pytest.mark.topology(KnownTopology.AD)
+            def test_example(client: Client, ad: AD):
```
danlavu: I don't know if we want to worry about this now, but here is the challenging part of abstraction. 

How can we make the following have the parameters and values? 
   
AD            
cert, key, csr = ad.ca.request(template="SmartcardLogon", subject="CN=testuser")

IPA
cert, key, csr = ipa.ca.request(principal="HTTP/client.ipa.test")

@spoore1

sssd_test_framework/misc/__init__.py
```diff
@@ -333,3 +333,29 @@ def get_attr(data: dict[str, Any], key: str, default: Any | None = None) -> Any
...
+    :param output: certutil output.
+    :type output: str
+    :returns: Dictionary of certificate attributes.
+    :rtype: dict[str, list[str]]
+    """
+    lines = [line.strip() for line in (output or "").splitlines() if line.strip()]
+    return attrs_parse(lines)
+
+
+def attrs_ad_parse(output: str) -> dict[str, list[str]]:
```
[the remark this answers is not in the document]
krishnavema: I originally planned to use a separate name for the certificate parser. Later, I thought about combining them, but I seem to have overlooked it. Thanks for pointing it out.

sssd_test_framework/roles/ad.py
krishnavema: Might need some changes and you can check here:[test](https://github.com/SSSD/sssd/pull/8453)

PR #231, opened 2026-01-29
Adding changes for dyndns tests

danlavu:


spoore1 (commented):
Just one question about the nslookup return value.   Running the provided test passes currently:

```
PASSED tests/test_dynamic_dns.py::test_dynamic_dns__updates_forward_records (ad)                                                                                                                                    
PASSED tests/test_dynamic_dns.py::test_dynamic_dns__updates_forward_records (ipa)                                                                                                                                   
PASSED tests/test_dynamic_dns.py::test_dynamic_dns__updates_forward_records (samba)                                                                                                                                 
```

sssd_test_framework/misc/__init__.py
```diff
@@ -198,34 +198,67 @@ def seconds_to_timespan(seconds: int) -> str:
     return f"{d:02d}:{h:02d}:{m:02d}:{s:02d}:00"
 
 
-def ip_to_ptr(ip_address: str) -> str:
+def ip_to_ptr(ip_address: str, prefixlen: int | None = None) -> str:
```
justin-stephenson: Why not use https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.reverse_pointer ?

danlavu: I'm not sure, I think I wrote most of the function, then stumbled upon ipaddress module.

tests/test_misc.py
```diff
@@ -287,14 +288,31 @@ def test_seconds_to_timespan(seconds: tuple[int, str]):
...
+    [
+        ("hostname", False),
+        ("hostname.domain.com", False),
+        ("192.168.1.1", True),
+        ("192.168.1.256", False),
+        ("::1", True),
+        ("001:db8::", True),
+    ],
+)
+def test_ip_is_valid(value, expected):
```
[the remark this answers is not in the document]
danlavu: Ack, will remove it then.

sssd_test_framework/roles/ad.py
```diff
@@ -2287,18 +2287,26 @@ def add_record(self, name: str, data: str | int) -> ADDNSZone:
...
         return self
 
-    def delete_record(self, name: str) -> None:
+    def delete_record(self, name: str, data: str) -> None:
         """
         Delete DNS record.
 
         :param name: Name of the record.
         :type name: str
+        :param data: Record data.
```
jakub-vavra-cz: What is data? Does not look clear to me what is expected there.

jakub-vavra-cz (changes requested):
I do not see what "data: str" is at first glance and the docstring does not elaborate.
Do we even need data when we want to delete some record by a name?

I think cleaner approach would be to have the dns records as objects (like users, groups, etc.) and keep the "data" needed from CRUD inside them. That way can providers implement the same approach including record removal on clenaup.

sssd_test_framework/roles/ad.py
```diff
@@ -2287,18 +2287,26 @@ def add_record(self, name: str, data: str | int) -> ADDNSZone:
         self.host.conn.run(f"Add-DnsServerResourceRecord -ZoneName {self.zone_name} {args} ")
         return self
 
-    def delete_record(self, name: str) -> None:
+    def delete_record(self, name: str, data: str) -> None:
```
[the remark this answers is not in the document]
danlavu: I reworked the method and put all the discovery logic inside.

sssd_test_framework/roles/ad.py
danlavu: I reworked the logic entirely.

danlavu:
> I would expect the creating/deleting DNS record method interface only requires 'dns zone', 'record type', and 'name' arguments (similar to `ipa dnsrecord-*` commands)

Technically, it is... because I'm working in the confines of the command. ;) I don't think we should remove the SSHFP attributes, adding more complexity. I reworked it, so it takes less parameters.

sssd_test_framework/roles/ipa.py
```diff
@@ -2891,26 +2892,52 @@ def add_record(self, name: str, data: str | int) -> IPADNSZone:
...
+                    self.host.conn.run(f"ipa dnsrecord-del {self.zone_name} --aaaa-rec={record}")
+
+        for ptr_records in records:
+            ptr_record = self.host.conn.run(f"dig +short -x +norecurse {ptr_records} '@{self.server}'").stdout_lines
+            ptr_record = [r.rstrip("\r") for r in ptr_record]
+            if ptr_record:
+                self.host.conn.run(f"ipa dnsrecord-del {self.zone_name} --ptr-rec={record}")
+        return None
+
+        time.sleep(5)  # Wait for the record to be deleted
```
justin-stephenson, line 2940: The `ipa dnsrecord-del` command should exit only when the delete operation is complete, so i'm curious why you need this sleep

sssd_test_framework/roles/samba.py
```diff
@@ -1360,37 +1360,56 @@ def add_record(self, name: str, data: str) -> SambaDNSZone:
         :rtype: SambaDNSZone
```
justin-stephenson, line 1360: Not just for samba, but does the dns generic `add_record` method still require this `data` argument?

justin-stephenson:
Can you investigate `CI / system (fedora-latest)` failures?

justin-stephenson:
> Can you investigate `CI / system (fedora-latest)` failures?

Disregard, They are unrelated to your PR.

sssd_test_framework/roles/samba.py
danlavu: Ipa requires it if we want to retain the sshfp record

sssd_test_framework/roles/ipa.py
danlavu: Without it, some cache still kept the record resolving, adding a sleep helped.

danlavu: Clarification, testing from a client. Performing a lookup on the record, it would still find the record due to some cache. Adding the sleep, the record was reliably and expectedly gone.

sssd_test_framework/misc/__init__.py
danlavu: Looking at it, the output isn't exactly perfect. Especially the ipv6 output. Let's just keep this for now?

justin-stephenson: I think you can format the output with https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.__format__ but I will leave the decision up to you if you want to use what you have now.

justin-stephenson (dismissed):
Ack, thank you.

jakub-vavra-cz (approved):
Looks Good

thalman (approved):
Looks good to me, thanks

PR #237, opened 2026-03-10
Replace umockdev with vfido for passkey testing

ikerexxe:
Migrate the passkey testing infrastructure from `umockdev` to `vfido`, providing a more robust and reliable solution for passkey authentication testing. The migration includes updating client feature detection to support `vfido`'s requirements, refactoring authentication utilities for improved integration, and updating the documentation to reflect the new `vfido` usage patterns. These changes enhance the overall passkey testing experience while maintaining backward compatibility where applicable.

SSSD PRCI run for these changes is available in https://github.com/SSSD/sssd/pull/8510

### DIFF
--- a/docs/guides/testing-passkey.rst
+++ b/docs/guides/testing-passkey.rst
@@ -4,20 +4,34 @@ Testing Passkeys
 Passkey can be tested using passkey related methods from
 :class:`sssd_test_framework.utils.sssctl.SSSCTLUtils` and
 :meth:`sssd_test_framework.utils.authentication.SUAuthenticationUtils.passkey`.
-It requires umockdev in order to correctly mock the passkey hardware token and
-record and playback the communications that happens between SSSD and
-the token.
+Tests use `virtual-fido <https://github.com/bulwarkid/virtual-fido>`_ to simulate a
+virtual FIDO2 authenticator. Pass ``virt_type="vfido"`` to
+:meth:`~sssd_test_framework.utils.sssctl.SSSCTLUtils.passkey_register`,
+:meth:`~sssd_test_framework.utils.authentication.SUAuthenticationUtils.passkey`, and
+:meth:`~sssd_test_framework.roles.ipa.IPAUser.passkey_add_register` to select the
+vfido backend. Prepare the device with :attr:`~sssd_test_framework.roles.client.Client.vfido`
+(:class:`~sssd_test_framework.utils.vfido.Vfido`).
 
-Three umockdev files are required to mock the device and playback the communication:
+**System Requirements**
 
-* device description (can be shared with all tests)
-* ioctl description (can be shared with all tests)
-* script of communication (mostly unique for each test)
+The ``vhci-hcd`` kernel module must be installed and loaded for virtual-fido
+to function properly:
 
-You can store the files in data directories that are returned by
-:func:`sssd_test_framework.fixtures.moduledatadir` (device and ioctl) and
-:func:`sssd_test_framework.fixtures.testdatadir` (script, passkey mapping)
-fixtures.
+.. code-block:: bash
+
+    # Install the kernel module
+    sudo dnf install -y kernel-modules-extra  # On Fedora/RHEL
+    # or
+    sudo apt install -y linux-modules-extra-$(uname -r)  # On Ubuntu/Debian
+
+    # Load the module
+    sudo modprobe vhci-hcd
+
+    # Verify the module is loaded
+    lsmod | grep vhci_hcd
+
+    # Load the module automatically at boot
+    sudo sh -c 'echo "vhci-hcd" > /etc/modules-load.d/vhci-hcd.conf'
 
 Test examples
 =============
@@ -33,57 +47,76 @@ Test examples
     from sssd_test_framework.topology import KnownTopology
 
 
-    @pytest.mark.tier(0)
     @pytest.mark.topology(KnownTopology.Client)
-    def test_passkey__register__sssctl(client: Client, moduledatadir: str, testdatadir: str):
-        """ Test registration of the passkey token with sssctl passkey-register"""
+    @pytest.mark.builtwith(client=["passkey", "vfido"])
+    def test_passkey__register_sssctl(client: Client):
+        """
+        Test registration of the passkey token with sssctl passkey-register
+        """
+        client.vfido.reset()
+        client.vfido.pin_enable()
+        client.vfido.pin_set(123456)
+        client.vfido.start()
+
         mapping = client.sssctl.passkey_register(
             username="user1",
             domain="ldap.test",
             pin=123456,
-            device=f"{moduledatadir}/umockdev.device",
-            ioctl=f"{moduledatadir}/umockdev.ioctl",
-            script=f"{testdatadir}/umockdev.script",
+            virt_type="vfido"
         )
-        with open(f"{testdatadir}/passkey-mapping") as f:
-            assert mapping == f.read().strip()
+
+        assert mapping.startswith("passkey:"), f"Invalid mapping prefix: {mapping}"
 
 
-    @pytest.mark.tier(0)
     @pytest.mark.topology(KnownTopology.IPA)
-    def test_passkey__register__ipa(ipa: IPA, moduledatadir: str, testdatadir: str):
-        """ Test registration of the passkey token with ipa user-add-passkey --register"""
+    @pytest.mark.builtwith(client=["passkey", "vfido"], ipa="passkey")
+    def test_passkey__register_ipa(client: Client, ipa: IPA):
+        """
+        Test registration of the passkey token with ipa user-add-passkey --register
+        """
+        client.vfido.reset()
+        client.vfido.pin_enable()
+        client.vfido.pin_set(123456)
+        client.vfido.start()
+
         mapping = (
             ipa.user("user1")
             .add()
             .passkey_add_register(
+                client=client,
                 pin=123456,
-                device=f"{moduledatadir}/umockdev.device",
-                ioctl=f"{moduledatadir}/umockdev.ioctl",
-                script=f"{testdatadir}/umockdev.script",
+                virt_type="vfido"
             )
         )
 
-        with open(f"{testdatadir}/passkey-mapping") as f:
-            assert mapping == f.read().strip()
+        assert mapping.startswith("Passkey mapping: passkey:")
 
 
-    @pytest.mark.tier(0)
     @pytest.mark.topology(KnownTopology.LDAP)
     @pytest.mark.topology(KnownTopology.IPA)
-    def test_passkey__su(client: Client, provider: GenericProvider, moduledatadir: str, testdatadir: str):
-        """ Test passkey authentication with su"""
-        suffix = type(provider).__name__.lower()
+    @pytest.mark.builtwith(client=["passkey", "vfido"], provider="passkey")
+    def test_passkey__su_user(client: Client, provider: GenericProvider):
+        """
+        Test passkey authentication with su
+        """
+        client.vfido.reset()
+        client.vfido.pin_enable()
+        client.vfido.pin_set(123456)
+        client.vfido.start()
 
-        with open(f"{testdatadir}/passkey-mapping.{suffix}") as f:
-            provider.user("user1").add().passkey_add(f.read().strip())
+        user = provider.user("user1").add()
+        if isinstance(provider, IPA):
+            user.passkey_add_register(client=client, pin=123456, virt_type="vfido")
+        else:
+            mapping = client.sssctl.passkey_register(
+                username="user1", domain=provider.domain, pin=123456, virt_type="vfido"
+            )
+            user.passkey_add(mapping)
 
         client.sssd.start()
 
         assert client.auth.su.passkey(
             username="user1",
             pin=123456,
-            device=f"{moduledatadir}/umockdev.device",
-            ioctl=f"{moduledatadir}/umockdev.ioctl",
-            script=f"{testdatadir}/umockdev.script.{suffix}",
+            virt_type="vfido",
         )
--- a/sssd_test_framework/roles/ipa.py
+++ b/sssd_test_framework/roles/ipa.py
@@ -1241,7 +1241,11 @@ class IPAUser(IPAObject, GenericUser):
             retry += 1
             time.sleep(1)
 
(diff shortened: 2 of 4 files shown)


sssd_test_framework/utils/authentication.py
```diff
@@ -417,6 +417,7 @@ def passkey_with_output(
     ) -> tuple[int, int, str, str]:
         """wrapper for passkey_with_output methods"""
         if "virt_type" in kwargs and kwargs["virt_type"] == "vfido":
+            del kwargs["virt_type"]
```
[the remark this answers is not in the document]
ikerexxe: This isn't necessary because the called doesn't use `kwargs` for anything

sssd_test_framework/utils/authentication.py
```diff
@@ -417,6 +417,7 @@ def passkey_with_output(
...
             return self.vfido_passkey_with_output(**kwargs)
```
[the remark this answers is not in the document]
ikerexxe: This isn't necessary because the called doesn't use `kwargs` for anything

sssd_test_framework/utils/authentication.py
```diff
@@ -427,6 +428,7 @@ def passkey(
     ) -> bool:
         """wrapper for passkey methods"""
         if "virt_type" in kwargs and kwargs["virt_type"] == "vfido":
+            del kwargs["virt_type"]
             return self.vfido_passkey(**kwargs)
```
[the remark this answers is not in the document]
ikerexxe: This isn't necessary because the called doesn't use `kwargs` for anything

sssd_test_framework/utils/authentication.py
```diff
@@ -738,71 +743,68 @@ def vfido_passkey_with_output(
...
             spawn {test_venv_bin}/vfido_touch
             set ID_touch $spawn_id
 
             expect {{
-                -i $ID_su -re "Authentication failure" {{exitmsg "Authentication failure" 1}}
-                -i $ID_su eof {{exitmsg "Passkey authentication successful" 0}}
-                -i $ID_su timeout {{exitmsg "Unexpected output" 213}}
+                -i $ID_touch eof {{}}
+                -i $ID_touch timeout {{}}
             }}
```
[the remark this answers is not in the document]
ikerexxe: This isn't working and it breaks many tests.

When authentication succeeds or fails, the script calls `exitmsg` which exits immediately, but the cleanup line `expect -i $ID_touch eof tries` to run after the main expect block. However, by the time it tries to access `$ID_touch`, the spawn ID is no longer valid because the `exitmsg` function has already closed processes.

sssd_test_framework/roles/ipa.py
```diff
@@ -1179,7 +1179,7 @@ def run_expect():
             retry += 1
             time.sleep(1)
 
-        return result.stdout_lines[-1].strip()
+        return result.stdout_lines[-2].strip()
```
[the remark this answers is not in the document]
ikerexxe: Done

docs/guides/testing-passkey.rst
```diff
@@ -4,20 +4,29 @@ Testing Passkeys
...
+    # Install the kernel module
+    sudo dnf install -y kernel-modules-extra  # On Fedora/RHEL
+    # or
+    sudo apt install -y linux-modules-extra-$(uname -r)  # On Ubuntu/Debian
+
+    # Load the module
+    sudo modprobe vhci-hcd
+
+    # Verify the module is loaded
+    lsmod | grep vhci
```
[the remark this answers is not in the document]
ikerexxe: Done

sssd_test_framework/utils/authentication.py
```diff
@@ -692,12 +694,15 @@ def vfido_passkey_with_output(
         """
 
         match auth_method:
-            case PasskeyAuthenticationUseCases.PASSKEY_PIN | PasskeyAuthenticationUseCases.PASSKEY_PIN_AND_PROMPTS:
+            case (
+                PasskeyAuthenticationUseCases.PASSKEY_PIN
+                | PasskeyAuthenticationUseCases.PASSKEY_PIN_AND_PROMPTS
+                | PasskeyAuthenticationUseCases.PASSKEY_FALLBACK_TO_PASSWORD
```
madhuriupadhye, line 702: Does the fallback-to-password scenario always require a PIN?

ikerexxe: As it stands currently SSSD always requests for the PIN prompt, even though the SSSD/FIDO2 configuration doesn't require one. This may change in the future, but for now the fallback-to-password always requires a PIN

alexey-tikhonov: Maybe worth a comment in the code.

sssd_test_framework/utils/authentication.py
```diff
@@ -740,71 +745,68 @@ def vfido_passkey_with_output(
...
-                        -i $ID_su -re "Password:*" {{send -i $ID_su "Secret123\r"}}
-                        -i $ID_su timeout {{exitmsg "Unexpected output" 209}}
-                        -i $ID_su eof {{exitmsg "Unexpected end of file" 210}}
-                    }}
+            }}
+
+            # Password fallback method needs password prompt
+            if {{ ($auth_method eq "{PasskeyAuthenticationUseCases.PASSKEY_FALLBACK_TO_PASSWORD}") }} {{
+                expect {{
+                    -i $ID_su -re "Password:*" {{send -i $ID_su "Secret123\r"}}
```
alexey-tikhonov, line 771: Would it make sense to make password a parameter?

sssd_test_framework/utils/authentication.py
```diff
@@ -740,71 +745,68 @@ def vfido_passkey_with_output(
...
+            }}
+
+            # Phase 3: handle touch prompts (for methods that show them)
+            if {{ ($auth_method eq "{PasskeyAuthenticationUseCases.PASSKEY_PROMPTS_NO_PIN}")
+                || ($auth_method eq "{PasskeyAuthenticationUseCases.PASSKEY_PIN_AND_PROMPTS}") }} {{
+                # Wait for touch prompt to appear and acknowledge it
+                expect {{
+                    -i $ID_su -re "{touch_prompt}*" {{
+                        # Touch prompt appeared - send Enter to acknowledge it
+                        send -i $ID_su "\n"
```
alexey-tikhonov, line 784: In old code it was `-i $ID_su -re "{touch_prompt}*" {{ }}`  --  without "\n".
Is this intentional?

sssd_test_framework/utils/authentication.py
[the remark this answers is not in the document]
alexey-tikhonov: Maybe it would be better to make `virt_type` an explicit arg?

alexey-tikhonov:
In general looks good to me. See a couple of remarks/questions inline.

docs/guides/testing-passkey.rst
```diff
@@ -4,20 +4,29 @@ Testing Passkeys
 Passkey can be tested using passkey related methods from
```
madhuriupadhye, line 4: Use `Passkeys can be tested` to match the title.

docs/guides/testing-passkey.rst
```diff
@@ -33,57 +42,76 @@ Test examples
...
             )
         )
 
-        with open(f"{testdatadir}/passkey-mapping") as f:
-            assert mapping == f.read().strip()
+        assert mapping.startswith("Passkey mapping: passkey:")
 
 
-    @pytest.mark.tier(0)
     @pytest.mark.topology(KnownTopology.LDAP)
```
madhuriupadhye, line 95: Misleading for LDAP topoloy.

docs/guides/testing-passkey.rst
```diff
@@ -4,20 +4,29 @@ Testing Passkeys
 Passkey can be tested using passkey related methods from
 :class:`sssd_test_framework.utils.sssctl.SSSCTLUtils` and
 :meth:`sssd_test_framework.utils.authentication.SUAuthenticationUtils.passkey`.
-It requires umockdev in order to correctly mock the passkey hardware token and
-record and playback the communications that happens between SSSD and
-the token.
+It requires `virtual-fido <https://github.com/bulwarkid/virtual-fido>`_ in order to
```
madhuriupadhye: virt_type undocumented, can we add like this?
```
Tests use `virtual-fido <https://github.com/bulwarkid/virtual-fido>`_ to simulate a
virtual FIDO2 authenticator. Pass ``virt_type="vfido"`` to
:meth:`~sssd_test_framework.utils.sssctl.SSSCTLUtils.passkey_register`,
:meth:`~sssd_test_framework.utils.authentication.SUAuthenticationUtils.passkey`, and
:meth:`~sssd_test_framework.roles.ipa.IPAUser.passkey_add_register` to select the
vfido backend. Prepare the device with :attr:`~sssd_test_framework.roles.client.Client.vfido`
(:class:`~sssd_test_framework.utils.vfido.Vfido`).
```

sssd_test_framework/utils/authentication.py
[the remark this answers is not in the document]
ikerexxe: Perhaps, but to be honest, I don’t have the expertise to say for sure, and I’d prefer to leave it as it is since it works. I think that once we’ve made the transition to vfido, we should take a close look at this part of the code and consider refactoring it.

sssd_test_framework/utils/authentication.py
ikerexxe: Yes, but that's out of scope for this PR

sssd_test_framework/utils/authentication.py
ikerexxe: Yes, and I'm surprised that it worked before

docs/guides/testing-passkey.rst
ikerexxe: I'm not following you. The text already states `Passkeys can be tested`

docs/guides/testing-passkey.rst
ikerexxe: What is misleading? Are you referring to `ipa` in `@pytest.mark.builtwith(client=["passkey", "vfido"], ipa="passkey")`?

sssd_test_framework/utils/authentication.py
ikerexxe: Done

docs/guides/testing-passkey.rst
ikerexxe: Done

sssd_test_framework/utils/authentication.py
alexey-tikhonov: ok

alexey-tikhonov (dismissed):

alexey-tikhonov:
ACK

sssd_test_framework/utils/authentication.py
spoore1: Wait, why do you need to send '\n' after su returns the `touch_prompt` message?   I thought in normal operation, we see the `touch_prompt` message and then touch the device and then it completes the action.   We don't usually have to press enter at that point do we?

docs/guides/testing-passkey.rst
madhuriupadhye: Line 95 is fine — it marks the test for the LDAP topology. The issue is line 97: `ipa="passkey"` conflicts with `LDAP` runs. Fixing line 97 to use `provider="passkey"`. or am I missed something?

ikerexxe: Updated

madhuriupadhye (approved):

docs/guides/testing-passkey.rst
```diff
@@ -4,20 +4,34 @@ Testing Passkeys
...
+vfido backend. Prepare the device with :attr:`~sssd_test_framework.roles.client.Client.vfido`
+(:class:`~sssd_test_framework.utils.vfido.Vfido`).
 
-Three umockdev files are required to mock the device and playback the communication:
+**System Requirements**
 
-* device description (can be shared with all tests)
-* ioctl description (can be shared with all tests)
-* script of communication (mostly unique for each test)
+The ``vhci-hcd`` kernel module must be installed and loaded for virtual-fido
```
spoore1, line 17: Should we mention usbip is also needed here?   It's installed by sssd-ci-containers when passkey_support is enabled.   This should include Centos 9/10 and Fedora from what I can tell.

alexey-tikhonov (approved):

spoore1 (commented):
Still reviewing but wanted to post this so it's not lost.

docs/guides/testing-passkey.rst
ikerexxe: In this section I was trying to convey the requirements for the host system. If you run this on your system with `sssd-ci-containers`, then you need to install and load the kernel module before starting the containers. Should I explain this in more detail to make it clearer?

`sssd-ci-containers` already have usbip installed by default, so that shouldn't be a problem, but I take that `sssd-test-framework` runs in more environments so maybe that should also be written. I'm open to suggestions here.

spoore1: Ah, got it.   Then no changes needed.

sssd_test_framework/utils/authentication.py
spoore1: Seems to be working so I'm ok with leaving it for now.

spoore1 (approved):
LGTM.  I ran tests with the [SSSD PR#8510](https://github.com/SSSD/sssd/pull/8510) using Fedora containers and only 1 failure which I think may be handled in the test:
```
FAILED tests/test_passkey.py::test_passkey__ipa_server_offline (ipa) - AssertionError: Credential cache found
============================================================================ 1 failed, 48 passed, 874 deselected in 1484.03s (0:24:44) =============================================================================
```

ikerexxe:
> LGTM. I ran tests with the [SSSD PR#8510](https://github.com/SSSD/sssd/pull/8510) using Fedora containers and only 1 failure which I think may be handled in the test:
> 
> ```
> FAILED tests/test_passkey.py::test_passkey__ipa_server_offline (ipa) - AssertionError: Credential cache found
> ============================================================================ 1 failed, 48 passed, 874 deselected in 1484.03s (0:24:44) =============================================================================
> ```

I'll rebase and clean up everything on https://github.com/SSSD/sssd/pull/8510. If the test still fails I'll check it and see how to fix it.

Merging...

PR #239, opened 2026-03-15
Add support for multi-token management in SmartCardUtils

krishnavema:


sssd_test_framework/utils/smartcard.py
```diff
@@ -150,6 +205,40 @@ def generate_cert(
         self.host.conn.run(self.cli.command("openssl req", args))
         return key_path, cert_path
 
+    def register_for_p11_child(self) -> None:
```
spoore1: I'm thinking this should move to the SSSDCommonConfiguration class along with the settings from https://github.com/SSSD/sssd/pull/8519 for configure_sssd_for_smartcard().   You could create a smartcard_with_softhsm() method that wraps all of that together into a single method you could call from the test with something like:
`client.sssd.common.smartcard_with_softhsm()`

sssd_test_framework/utils/smartcard.py
```diff
@@ -76,12 +76,43 @@ def initialize_card(self, label: str = "sc_test", so_pin: str = "12345678", user
             self.cli.command("softhsm2-util --init-token", args), env={"SOFTHSM2_CONF": self.SOFTHSM2_CONF_PATH}
         )
 
+    def initialize_additional_token(self, label: str, so_pin: str = "12345678", user_pin: str = "123456") -> None:
```
spoore1: Could the initialize_card() method be modified to take an option to skip the cleanup if set instead of having an additional method?

spoore1 (changes requested):
I'll have more comments in the test PR but this is a start.

sssd_test_framework/utils/authentication.py
```diff
@@ -844,6 +850,54 @@ def vfido_passkey(
         rc, _, _, _ = self.vfido_passkey_with_output(username=username, pin=pin, command=command)
         return rc == 0
 
+    def smartcard_with_su_output(self, username: str, pin: str, *, num_certs: int = 1) -> ProcessResult:
```
spoore1: I would rename this to `smartcard_with_output` to better match the other functions in the class.   And since we'll call it with `client.auth.su.smartcard_with_output` I don't think we need the "su" in the method name.

sssd_test_framework/utils/authentication.py
```diff
@@ -844,6 +850,54 @@ def vfido_passkey(
...
+            time.sleep(USER_RESOLVABLE_INTERVAL_S)
+            check = self.host.conn.run(f"getent passwd {username}", raise_on_error=False)
+            if check.rc == 0:
+                break
+            if attempt == USER_RESOLVABLE_CACHE_EXPIRY_ATTEMPT:
+                self.host.conn.run("sss_cache -E", raise_on_error=False)
+        else:
+            raise AssertionError(
+                f"User '{username}' was not resolvable by SSSD after {USER_RESOLVABLE_ATTEMPTS} attempts"
+            )
```
spoore1: Did you run into issues resolving the users when you wrote this?   We haven't usually needed this type of check in the past.   I don't have anything against the extra checks but, I'm wondering if they were added because of a particular issue we might need to fix elsewhere.

sssd_test_framework/utils/authentication.py
```diff
@@ -844,6 +850,54 @@ def vfido_passkey(
...
+
+        su_input = f"1\n{pin}" if num_certs > 1 else pin
```
spoore1: While this may work as expected with multiple tokens (I'm not 100% sure ATM),  won't this force selection of the first certificate from a token that has multiple certificates?   We might want an optional argument for certificate selection with 1 being the default.

sssd_test_framework/utils/authentication.py
```diff
@@ -844,6 +850,54 @@ def vfido_passkey(
...
+        return self.host.conn.run(
+            f"su - {username} -c 'su - {username} -c whoami'",
+            input=su_input,
+            raise_on_error=False,
+        )
+
+    def smartcard_with_su(self, username: str, pin: str, *, num_certs: int = 1) -> bool:
```
spoore1: Like the other method, I would suggest renaming this one to just `smartcard` to closer match the other methods in the class.

sssd_test_framework/utils/smartcard.py
```diff
@@ -76,12 +76,43 @@ def initialize_card(self, label: str = "sc_test", so_pin: str = "12345678", user
...
+        """
+        args: CLIBuilderArgs = {
+            "label": (self.cli.option.VALUE, label),
+            "free": (self.cli.option.SWITCH, True),
+            "so-pin": (self.cli.option.VALUE, so_pin),
+            "pin": (self.cli.option.VALUE, user_pin),
+        }
+        self.host.conn.run(
+            self.cli.command("softhsm2-util --init-token", args), env={"SOFTHSM2_CONF": self.SOFTHSM2_CONF_PATH}
+        )
```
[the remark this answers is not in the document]
spoore1: I could be missing something here but, I don't see any code duplication.  I think this can be ignored unless someone sees the duplicate code.

sssd_test_framework/utils/sssd.py
```diff
@@ -1178,3 +1179,31 @@ def subid(self) -> None:
         Configure SSSD for subid.
         """
         self.sssd.authselect.select("sssd", ["with-subid"])
+
+    def smartcard_with_softhsm(self, smartcard: SmartCardUtils) -> None:
```
spoore1, line 1183: Just a side note.  You may want a `smartcard_with_virtcacard` or similar to handle the setup since it's fairly standard (I think).

sssd_test_framework/utils/sssd.py
```diff
@@ -1178,3 +1179,31 @@ def subid(self) -> None:
...
+
+        :param smartcard: SmartCardUtils instance.
+        :type smartcard: SmartCardUtils
+        """
+        conf = smartcard.SOFTHSM2_CONF_PATH
+        token_storage = smartcard.TOKEN_STORAGE_PATH
+        module = "/usr/lib64/pkcs11/libsofthsm2.so"
+
+        smartcard.host.conn.run(f"grep -q 'slots.removable' {conf} || echo 'slots.removable = true' >> {conf}")
+        smartcard.host.conn.run(f"cp {conf} /etc/softhsm2.conf")
```
spoore1: I believe we want to try to use python native solutions when possible over these types of file edits.   I think you can do something like:
```
softhsm_conf = smartcard.fs.read(conf)
if "slots.removable" not in softhsm_conf:
    smartcard.fs.append(conf, "\nslots.removable = true")
```

If you want to work in a temp file before copying to the real one, you could also make use of the `fs.copy()` method.

sssd_test_framework/utils/sssd.py
```diff
@@ -1178,3 +1179,31 @@ def subid(self) -> None:
...
+        smartcard.host.conn.run(f'echo "module: {module}" > /etc/pkcs11/modules/softhsm2.module')
```
spoore1: Same thing here.   Use the `fs` methods where possible.

sssd_test_framework/utils/sssd.py
```diff
@@ -1178,3 +1179,31 @@ def subid(self) -> None:
...
+        smartcard.fs.mkdir_p("/etc/systemd/system/sssd.service.d")
+        smartcard.host.conn.run(
+            f'printf "[Service]\\nEnvironment=SOFTHSM2_CONF={conf}\\n" '
+            f"> /etc/systemd/system/sssd.service.d/softhsm.conf"
```
spoore1: I think you can do this one with a simple `fs.write()`

sssd_test_framework/utils/sssd.py
```diff
@@ -1178,3 +1179,31 @@ def subid(self) -> None:
...
+        )
+        smartcard.host.conn.run("systemctl daemon-reload")
```
spoore1: Use `smartcard.svc.reload_daemon()` here I think.

sssd_test_framework/utils/sssd.py
```diff
@@ -1178,3 +1179,31 @@ def subid(self) -> None:
...
+        smartcard.host.conn.run("chmod -R o+rX /opt/test_ca/")
+        smartcard.host.conn.run(f"chown -R sssd:sssd {token_storage}/ && chmod -R 770 {token_storage}/")
```
spoore1: I think you should be able to use `fs.chown()` and `fs.chmod()` here.

spoore1 (changes requested):
I have a few questions and suggestions.  Mostly changing some of the shell commands to `fs` calls for the file edits/manipulations.

spoore1 (commented):
Overall it's looking good.  Thanks for the updates.  Only thing left I think is if you move enroll_to_token from SSSD#8519 to here.

sssd_test_framework/utils/smartcard.py
```diff
@@ -199,3 +229,57 @@ def test_example(client: Client):
...
+        :param cert_id: PKCS#11 object ID, defaults to "01".
+        :type cert_id: str, optional
+        :param pin: User PIN for the token, defaults to "123456".
+        :type pin: str, optional
+        :param init: Initialize (and reset) the token before enrolling,
+            defaults to False.
+        :type init: bool, optional
+        """
+        if init:
+            self.initialize_card(label=token_label or "sc_test", user_pin=pin)
```
sumit-bose: Hi,

would it make sense to use `sc_test` as default for `token_label`?

bye,
Sumit

krishnavema: fixed Thanks!

sssd_test_framework/utils/sssd.py
spoore1: This is something for another PR.  My comment here wasn't meant to hold up this PR so I'm closing this as resolved.

spoore1 (approved):
LGTM.

sumit-bose (approved):
Hi,

thank you for the updates, ACK.

bye,
Sumit

spoore1:
Test failure unrelated:

```
ERROR tests/test_kcm.py::test_kcm__tgt_renewal_updates_ticket_as_configured (client) - TypeError: _FlakyPlugin._make_test_flaky() got an unexpected keyword argument 'reruns'
2621
```

PR #240, opened 2026-03-23
Test: changes for krb_misc

madhuriupadhye:


sssd_test_framework/roles/kdc.py
```diff
@@ -163,21 +163,38 @@ def __init__(self, role: KDC, name: str) -> None:
         self.name: str = name
         """Principal name."""
 
-    def add(self, *, password: str | None = "Secret123") -> KDCPrincipal:
+    def add(
+        self,
+        *,
+        password: str | None = "Secret123",
+        requires_preauth: bool = False,
+        extra_options: str | None = None,
```
danlavu: nitpick, we use `args` everywhere else. I would just use the same variable name.

sssd_test_framework/roles/kdc.py
```diff
@@ -163,21 +163,38 @@ def __init__(self, role: KDC, name: str) -> None:
...
+        :type extra_options: str | None, optional
         :return: Self.
         :rtype: KDCPrincipal
         """
+        opts: list[str] = []
+        if requires_preauth:
+            opts.append("+requires_preauth")
+        if extra_options:
+            opts.append(extra_options)
+        opts_str = " " + " ".join(opts) if opts else ""
```
danlavu: Nitpick, I find this strange, for the leading space? I think it's more readable, add a space in the command on line 195, ```-randkey {opt_str} "{self.name}"```

sssd_test_framework/utils/authentication.py
```diff
@@ -1597,3 +1597,55 @@ def password(self, user: str, password: str, new_password: str, retyped: str | N
             raise ExpectScriptError(result.rc)
 
         return result.rc == 0
+
```
danlavu: Why is it important that this is done via SSH?

danlavu (commented):
So, I looked at the tests, and I don't think it's important to do this on LDAP/KRB5. I suggest we convert the tests and use the IPA provider. Thoughts?

madhuriupadhye:
> So, I looked at the tests, and I don't think it's important to do this on LDAP/KRB5. I suggest we convert the tests and use the IPA provider. Thoughts?

Only BZ 773660 (clock skew) can use IPA — krb5_child runs the same way under IPA auth.

BZ 869150 and BZ 805281 cannot — they require manual keytab manipulation (remove, replace), which breaks IPA enrollment. IPA owns its keytab and uses ipa_child instead of ldap_child, so the bug paths aren't exercised.

danlavu:
> Only BZ 773660 (clock skew) can use IPA — krb5_child runs the same way under IPA auth.
> 
> BZ 869150 and BZ 805281 cannot — they require manual keytab manipulation (remove, replace), which breaks IPA enrollment. IPA owns its keytab and uses ipa_child instead of ldap_child, so the bug paths aren't exercised.

Ack, understood. I'd like to better understand the case for passwd over SSH. If it's simply logging in via ssh, then issuing 'passwd', we should make it like the others. 

Because we have 

```client.auth.passwd.password('tuser', 'Secret123', 'New_password123')```

make it

```client.auth.ssh.passwd.password('user', 'password', 'new_pass')``` and create the two method, ```passwd```, ```passwd_with_output```.

File: sssd_test_framework/roles/kdc.py (hunk 2 above)
Comment (thalman) [reply sssd_test_framework/roles/kdc.py]: +1

File: sssd_test_framework/utils/authentication.py (hunk 3 above)
Comment (madhuriupadhye) [reply sssd_test_framework/utils/authentication.py]: Considering other test cases which I already plan to covert has this use case,
```
@pytest.mark.importance("high")
@pytest.mark.authentication
@pytest.mark.ticket(gh=677)
def test_ldap_krb__password_change_via_ssh(
    client: Client, provider: GenericProvider, kdc: KDC
):
    """
    :title: Password change via SSH triggers krb5_child initial auth
    :setup:
        1. Add user puser1 to LDAP and KDC
        2. Configure SSSD with LDAP+KRB5, chpass_provider=krb5
        3. Set debug_level=0xFFF0
        4. Restart SSSD and clear cache
    :steps:
        1. Change puser1 password via SSH session using passwd
        2. Check krb5_child.log for initial auth message
    :expectedresults:
        1. Password change succeeds
        2. "Initial authentication for change password" in krb5_child.log
    :customerscenario: True
    """
```

**Comment by madhuriupadhye** [comment]
> > Only BZ 773660 (clock skew) can use IPA — krb5_child runs the same way under IPA auth.
> > BZ 869150 and BZ 805281 cannot — they require manual keytab manipulation (remove, replace), which breaks IPA enrollment. IPA owns its keytab and uses ipa_child instead of ldap_child, so the bug paths aren't exercised.
> 
> Ack, understood. I'd like to better understand the case for passwd over SSH. If it's simply logging in via ssh, then issuing 'passwd', we should make it like the others.
> 
> Because we have
> 
> `client.auth.passwd.password('tuser', 'Secret123', 'New_password123')`
> 
> make it
> 
> `client.auth.ssh.passwd.password('user', 'password', 'new_pass')` and create the two method, `passwd`, `passwd_with_output`.

Those changes require for next set of tests, so can I remove it from this PR and will create the new fresh PR for with next set of tests.

File: sssd_test_framework/roles/kdc.py (hunk 1 above)
Comment (danlavu) [reply sssd_test_framework/roles/kdc.py]: Let's just name this args, like the rest of the arguement parameter in several functions.

File: sssd_test_framework/roles/kdc.py (hunk 2 above)
Comment (danlavu) [reply sssd_test_framework/roles/kdc.py]: No change?

File: sssd_test_framework/utils/authentication.py (hunk 3 above)
Comment (danlavu) [reply sssd_test_framework/utils/authentication.py]: Ack.

**Comment by danlavu** [comment]
> Because we have
> 
> `client.auth.passwd.password('tuser', 'Secret123', 'New_password123')`
> 
> make it
> 
> `client.auth.ssh.passwd.password('user', 'password', 'new_pass')` and create the two method, `passwd`, `passwd_with_output`.


?

File: sssd_test_framework/utils/krb5.py [hunk 4]
```diff
@@ -0,0 +1,95 @@
+"""Kerberos and keytab utilities."""
```
danlavu: I don't think it should be its own class, we should probably just extend the authentication krb utils.

danlavu (changes requested):

danlavu (dismissed):
This looks great! thank you!

sssd_test_framework/utils/chrony.py
```diff
@@ -0,0 +1,96 @@
+"""Chrony time manipulation utilities."""
```
danlavu: nitpick, since it's the title, every word should be capitalized, but we can also shorten it.  

```Chrony Utilities```

File: sssd_test_framework/utils/chrony.py [hunk 6]
```diff
@@ -0,0 +1,96 @@
...
+    @contextmanager
+    def time_skew(self, offset: str) -> Generator[None, None, None]:
+        """
+        Shift the system clock by ``offset`` and restore it on exit.
+
+        Restarts ``chronyd``, enables manual mode, applies the time offset
+        via ``chronyc settime``, then yields. On exit (normal or exception),
+        the original time is restored and NTP is re-synced.
+
+        :param offset: Chrony time offset string (e.g. ``"+ 1 day"``, ``"- 2 hours"``).
```
danlavu: I would simplify this. I think most of our test cases are that the time is just off, something like

time_skew(self, hours:int, minutes:int, seconds:int) or even just
time_skew(self, seconds:int)

danlavu (changes requested):
Thank you for creating the chorny class. Very pleasant surprise. One suggestion and the comment is inline.

sssd_test_framework/utils/chrony.py
madhuriupadhye: updated.

sssd_test_framework/utils/chrony.py
madhuriupadhye: updated

sssd_test_framework/topology_controllers.py
```diff
@@ -124,6 +128,82 @@ class LDAPTopologyController(ProvisionedBackupTopologyController):
     pass
 
 
+class LDAPKrb5TopologyController(LDAPTopologyController):
```
danlavu: Nitpick, I'd name the class LDAPKRB5TopologyController

madhuriupadhye: updated.

danlavu (approved):

thalman (approved):
LGTM, thanks

PR #242, opened 2026-04-21
Extend framework for sudo use-cases

jakub-vavra-cz:
Implement sudo alias for local provider
Add local netgroups implementation

### DIFF
--- a/sssd_test_framework/roles/client.py
+++ b/sssd_test_framework/roles/client.py
@@ -12,6 +12,7 @@ from ..utils.gdm import GDM
 from ..utils.ldb import LDBUtils
 from ..utils.local_users import (
     LocalGroup,
+    LocalNetgroup,
     LocalSudoAlias,
     LocalSudoAliasKind,
     LocalSudoRule,
@@ -192,6 +193,12 @@ class Client(BaseLinuxRole[ClientHost]):
         Get a sudoers alias object (see :meth:`LocalUsersUtils.sudo_alias`).
         """
         return self.local.sudo_alias(name, kind)
+
+    def netgroup(self, name: str) -> LocalNetgroup:
+        """
+        Get a local netgroup object (see :meth:`LocalUsersUtils.netgroup`).
+        """
+        return self.local.netgroup(name)
 
     def sudorule(self, name: str) -> LocalSudoRule:
         """
--- a/sssd_test_framework/roles/client.py
+++ b/sssd_test_framework/roles/client.py
@@ -188,16 +188,28 @@ class Client(BaseLinuxRole[ClientHost]):
 
         return LocalGroup(self.local, name)
 
-    def sudo_alias(self, name: str, kind: LocalSudoAliasKind) -> LocalSudoAlias:
+    def sudoalias(self, name: str, kind: LocalSudoAliasKind) -> LocalSudoAlias:
         """
-        Get a sudoers alias object (see :meth:`LocalUsersUtils.sudo_alias`).
+        Get sudo alias object.
+        :param name: Sudo alias name.
+        :type name: str
+        :param kind: Alias kind.
+        :type kind: LocalSudoAliasKind
+        :return: New sudo alias object.
+        :rtype: LocalSudoAlias
         """
-        return self.local.sudo_alias(name, kind)
+
+        return self.local.sudoalias(name, kind)
 
     def netgroup(self, name: str) -> LocalNetgroup:
         """
-        Get a local netgroup object (see :meth:`LocalUsersUtils.netgroup`).
+        Get netgroup object.
+        :param name: Netgroup name.
+        :type name: str
+        :return: New netgroup object.
+        :rtype: LocalNetgroup
         """
+
         return self.local.netgroup(name)
 
     def sudorule(self, name: str) -> LocalSudoRule:
@@ -209,4 +221,4 @@ class Client(BaseLinuxRole[ClientHost]):
         :rtype: LocalSudoRule
         """
 
-        return LocalSudoRule(self.local, name)
+        return self.local.sudorule(name)
--- a/sssd_test_framework/roles/client.py
+++ b/sssd_test_framework/roles/client.py
@@ -10,7 +10,14 @@ from ..utils.adcli import AdcliUtils
 from ..utils.automount import AutomountUtils
 from ..utils.gdm import GDM
 from ..utils.ldb import LDBUtils
-from ..utils.local_users import LocalGroup, LocalSudoRule, LocalUser, LocalUsersUtils
+from ..utils.local_users import (
+    LocalGroup,
+    LocalSudoAlias,
+    LocalSudoAliasKind,
+    LocalSudoRule,
+    LocalUser,
+    LocalUsersUtils,
+)
 from ..utils.realmd import RealmUtils
 from ..utils.sbus import DBUSDestination, DBUSKnownBus
 from ..utils.smartcard import SmartCardUtils
@@ -179,6 +186,12 @@ class Client(BaseLinuxRole[ClientHost]):
         """
 
         return LocalGroup(self.local, name)
+
+    def sudo_alias(self, name: str, kind: LocalSudoAliasKind) -> LocalSudoAlias:
+        """
+        Get a sudoers alias object (see :meth:`LocalUsersUtils.sudo_alias`).
+        """
+        return self.local.sudo_alias(name, kind)
 
     def sudorule(self, name: str) -> LocalSudoRule:
         """
--- a/sssd_test_framework/utils/local_users.py
+++ b/sssd_test_framework/utils/local_users.py
@@ -11,10 +11,14 @@ from pytest_mh.cli import CLIBuilder, CLIBuilderArgs
 from pytest_mh.conn import ProcessLogLevel
 from pytest_mh.utils.fs import LinuxFileSystem
 
+from ..roles.generic import GenericNetgroupMember
+
 __all__ = [
     "LocalGroup",
     "LocalUser",
     "LocalUsersUtils",
+    "LocalNetgroup",
+    "LocalNetgroupMember",
     "LocalSudoAlias",
     "LocalSudoAliasKind",
     "LocalSudoRule",
@@ -45,6 +49,10 @@ class LocalUsersUtils(MultihostUtility[MultihostHost]):
         self.fs: LinuxFileSystem = fs
         self._users: list[str] = []
         self._groups: list[str] = []
+        self._netgroup_baseline: str | None = None
+        self._netgroup_initialized: bool = False
+        self._netgroup_names_touched: set[str] = set()
+        self._netgroups: dict[str, LocalNetgroup] = {}
         self._sudoaliases: list[LocalSudoAlias] = []
         self._sudorules: list[LocalSudoRule] = []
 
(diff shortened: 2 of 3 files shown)


sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +556,335 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+        :param members: One or more users/groups (for ``user`` / ``runas``), hostnames (for ``host``),
+            or commands (for ``command``).
+        :type members: str | LocalUser | LocalGroup | list[str | LocalUser | LocalGroup]
+        :param order: Optional ordering prefix for the drop-in file name (lower sorts first).
+        :type order: int | None, optional
+        :return: Self.
+        :rtype: LocalSudoAlias
+        """
+        self._order = order
+        orderstr = f"{order:02d}" if order is not None else str(len(self.util._sudoaliases))
```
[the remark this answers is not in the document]
jakub-vavra-cz: done

jakub-vavra-cz:
The ci failures are unrelated. There is some issue in  docs building and failing test_smartcard__su_as_local_user that probably has a fix in https://github.com/SSSD/sssd-test-framework/pull/239.

sssd_test_framework/roles/client.py
```diff
@@ -180,6 +188,18 @@ def group(self, name: str) -> LocalGroup:
 
         return LocalGroup(self.local, name)
 
+    def sudo_alias(self, name: str, kind: LocalSudoAliasKind) -> LocalSudoAlias:
```
madhuriupadhye: Please align docstrings with `user` / `group`: include `:param`,` :return`, and `:rtype`.
same for `def netgroup`

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +556,335 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+        :param group: Not supported for local netgroups.
+        :type group: LocalGroup | str | None, optional
+        :param hostgroup: Not supported for local netgroups.
+        :type hostgroup: str | None, optional
+        :param ng: Nested netgroup, defaults to None
+        :type ng: LocalNetgroup | str | None, optional
+        """
+        if group is not None or hostgroup is not None:
+            raise ValueError(
+                "Local /etc/netgroup netgroups do not support group or hostgroup members " "(use IPANetgroup for IPA)."
```
madhuriupadhye: single readable ValueError string, 
Merge the split string literals into one message.
```
        if group is not None or hostgroup is not None:
            raise ValueError(
                "Local /etc/netgroup netgroups do not support group or hostgroup members "
                "(use IPANetgroup for IPA)."
            )
```

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +556,335 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+
+        :param members: New member list, defaults to None (keep previous).
+        :type members: str | LocalUser | LocalGroup | list[str | LocalUser | LocalGroup] | None, optional
+        :param order: Optional ordering prefix for the drop-in file name.
+        :type order: int | None, optional
+        :return: Self.
+        :rtype: LocalSudoAlias
+        """
+        prev_order = self._order
+        self.delete()
```
madhuriupadhye, line 917: modify calls `delete()` then `add()`, so changing `order` (or first-time add after delete) can change the drop-in filename. Please document that so callers are not surprised.
something like this
```
        Changing ``order`` (or calling ``modify`` after the alias was removed) may assign
        a new drop-in file name under ``/etc/sudoers.d/`` because ``delete`` clears the
        previous ``filename``.
```

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +556,335 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+        """
+        self.util.logger.info(f'Adding members to local netgroup "{self.name}" on {self.util.host.hostname}')
+
+        if not members:
+            return self
+
+        if self.name not in self.util._netgroups:
+            raise RuntimeError(f'Netgroup "{self.name}" was not created; call add() first')
+
+        for m in members:
```
madhuriupadhye, line 720: Either state explicitly that duplicate member strings are allowed, or avoid appending duplicates. Pick one and document it.
To avoid duplicates,
```
        for m in members:
            s = m.to_member_string()
            if s not in self._members:
                self._members.append(s)
```

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +556,335 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+        self.filename: str | None = None
+        self.alias_str: str | None = None
+        self.__members: Any = None
+        self._order: int | None = None
+
+    def __str__(self) -> str:
+        return self.name
+
+    @staticmethod
+    def _format_member(kind: LocalSudoAliasKind, item: str | LocalUser | LocalGroup) -> str:
```
madhuriupadhye, line 819: please add doc string.

madhuriupadhye:
Please check CI failures also.

sssd_test_framework/roles/client.py
jakub-vavra-cz: done

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: I could not keep it that way, black keeps screwing it up after I manually break it :-(

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: done

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: done

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: added

justin-stephenson:
@jakub-vavra-cz you can add the following change to this PR as `IdEntry` is used only inside `tools.py` it does not need to be exported. It should resolve the error.

```
diff --git a/sssd_test_framework/utils/tools.py b/sssd_test_framework/utils/tools.py
index 10ddfe4..7730314 100644
--- a/sssd_test_framework/utils/tools.py
+++ b/sssd_test_framework/utils/tools.py
@@ -12,7 +12,6 @@ from pytest_mh.utils.fs import LinuxFileSystem
 __all__ = [
     "GetentUtils",
     "GroupEntry",
-    "IdEntry",
     "LinuxToolsUtils",
     "PasswdEntry",
     "UnixGroup",

```

sssd_test_framework/roles/client.py
```diff
@@ -180,6 +188,30 @@ def group(self, name: str) -> LocalGroup:
...
```
danlavu: nitpick: I don't like the _ compared to the other methods in this class, sudorule, netgroups, but it does match /etc/sudoers. I'm fine with it, but I'd prefer it to be consistent with the class.

sssd_test_framework/roles/client.py
```diff
@@ -180,6 +188,30 @@ def group(self, name: str) -> LocalGroup:
...
+        """
+        Get sudo alias object.
+        :param name: Sudo alias name.
+        :type name: str
+        :param kind: Alias kind
```
danlavu: Nitpick, ```.``` The pattern is a ```.``` at the end of the param description, unless it's it has a ```,``` 

i.e.
```
:param kind: Alias kind. 
```

if it was optional and contains a ```,```
```
:param kind: Alias kind, optional 
```

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +565,401 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+        """
+        return self.remove_members(
+            [LocalNetgroupMember(host=host, user=user, group=group, hostgroup=hostgroup, ng=ng)]
+        )
+
+    def remove_members(self, members: list[LocalNetgroupMember]) -> LocalNetgroup:
+        """
+        Remove netgroup members
+
+        :param members: Members to remove (must match strings produced when added).
```
danlavu: nitpick, we can assume the statement within the (), I think this should be removed.

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +565,401 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+        :param name: Netgroup name.
+        :type name: str
+        """
+        self.util = util
+        self.name = name
+        self._members: list[str] = []
+
+    def __str__(self) -> str:
+        """
+        Return the netgroup name (same as ``self.name``).
```
danlavu: nitpick, we can assume the statement within the (), I think this should be removed.

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +565,401 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+            raise ValueError("Netgroup member must specify host, user, and/or nested netgroup (ng)")
+
+        h = self.host if self.host is not None else "-"
+        u = self.user if self.user is not None else "-"
+        return f"({h},{u},)"
+
+
+class LocalNetgroup(object):
+    """
+    Local netgroup management via ``/etc/netgroup`` (``files`` NSS backend).
```
danlavu: Nitpick, our audience is QE, we should know what netgroups are. ```Local Netgroup management.``` is sufficient.

sssd_test_framework/utils/local_users.py
```diff
@@ -120,6 +143,106 @@ def test_example(client: Client):
...
+        addition_str = "\n".join(additions)
+        if addition_str:
+            addition_str += "\n"
+        self.fs.write("/etc/netgroup", body + addition_str)
+
+    def sudo_alias(self, name: str, kind: LocalSudoAliasKind) -> LocalSudoAlias:
+        """
+        Get a sudoers alias object (``User_Alias``, ``Runas_Alias``, ``Host_Alias``, or ``Cmnd_Alias``).
+
+        Alias names must match sudoers rules: start with an uppercase letter and contain only
```
danlavu, line 213: This docstring is good, short, and informative. This will help others on the team to understand the code.

sssd_test_framework/utils/local_users.py
```diff
@@ -442,6 +565,401 @@ def remove_members(self, members: list[LocalUser]) -> LocalGroup:
...
+        self.name = name
+        self.kind = kind
+        self.filename: str | None = None
+        self.alias_str: str | None = None
+        self.__members: Any = None
+        self._order: int | None = None
+
+    def __str__(self) -> str:
+        """
+        Return the alias name (same as ``self.name``).
```
danlavu: nitpick, we can assume the statement within the (), I think this should be removed.

sssd_test_framework/utils/local_users.py
```diff
@@ -120,6 +143,106 @@ def test_example(client: Client):
...
+        if self._netgroup_initialized:
+            return
+        self.fs.backup("/etc/netgroup")
+        result = self.host.conn.exec(["cat", "/etc/netgroup"], raise_on_error=False)
+        self._netgroup_baseline = result.stdout if result.rc == 0 else ""
+        self._netgroup_initialized = True
+
+    def _rewrite_netgroup_file(self) -> None:
+        """
+        Write ``/etc/netgroup`` from the cached baseline, dropping lines for netgroup names this
```
danlavu, line 183: I would say this is common knowledge for us, each netgroup is a line, and a class is a netgroup.

danlavu (changes requested):
The code is great, thank you. I don't like the docstrings though. I think it should be less verbose and concise. I made some comments as examples of what I think should change, but I didn't apply them to the entire file. Can you go through the docstrings with the comments in mind? Thanks.

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: done

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: done

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: done

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: shortened it a bit, but keeping the info that is necessary to understand the usage.

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: done

sssd_test_framework/utils/local_users.py
jakub-vavra-cz: This actually explains how the function handles pre-existing file with test-created netgroups.
I want to keep it there.

sssd_test_framework/roles/client.py
jakub-vavra-cz: done

sssd_test_framework/roles/client.py
jakub-vavra-cz: done

jakub-vavra-cz:
@madhuriupadhye I added the missing docstrings and refactored alias to allow changing order and handled duplicates for netgroup.
@danlavu I tried to cut down the docstring fluff a bit and add missing "." to make them more consistent.
I also changed sudo_alias to sudoalias but I will need to update the tests that are using it. :-(
Thanks @justin-stephenson, the fix worked.

madhuriupadhye (dismissed):
Looks good to me, just waiting to finish the CI.

sssd_test_framework/utils/local_users.py
danlavu: ack.

danlavu (dismissed):
It's great, and I'll approve it, but I need a trivial change. There was a misunderstanding, type: never has ```.``` at the end. The pattern only applies to the :param: docstring. 

sed '/:type/s/\.$//'  should do the trick.

jakub-vavra-cz:
> It's great, and I'll approve it, but I need a trivial change. There was a misunderstanding, type: never has `.` at the end. The pattern only applies to the :param: docstring.
> 
> sed '/:type/s/.$//' should do the trick.

Did that for :type and :rtype that I both touched. Hopefully it is okay now.

danlavu (approved):

madhuriupadhye (approved):

PR #243, opened 2026-04-27
Add LDAP/Kerberos helpers (DNS SRV, GSSAPI, getent)

madhuriupadhye:
Add LDAP/Kerberos helpers (DNS SRV, GSSAPI, getent)
Extend sssd-test-framework so LDAP+Kerberos system tests
can configure DNS, resolvers, and Directory Server GSSAPI
without per-test `named`, `dig`, or `getent` boilerplate.

- **dig** — Parse A/PTR/SRV via `jc`; auto-append `SRV` for `_service` queries; strip trailing dots in results.
- **has_srv_record** — True when SRV exists (parsed `dig` or `dig +short` fallback).
- **prepare_ldap_krb5_srv_discovery** — Single path for LDAP/Kerberos SRV discovery on the test client:
  - Use lab DNS when `_ldap._tcp` / `_kerberos._udp` SRV already resolve.
  - Else try topology `dns.test`.
  - Else local `named` on `127.0.0.1` with LDAP/KDC A records and SRV in the discovery domain.
  - Always pin LDAP and KDC FQDNs in `/etc/hosts` (krb5_child uses `/etc/krb5.conf`, not only SSSD `_srv_` discovery).
  - Probe KDC TCP port 88 before password authentication.
- **setup_sasl_canonicalize_bogus_ptr** — Local `named` + `/etc/hosts` for `ldap_sasl_canonicalize` with bogus PTR; backs up and restores files via `client.fs`.
- **named helpers** — Forwarders injection and `dnssec-validation` handling so `named-checkconf` does not fail on duplicate options.
- **AHostSv4Entry** and **ahostsv4** — Parsed `getent ahostsv4` output.
- **resolve_ipv4** — Use multihost `host.ip` when set, otherwise `ahostsv4`.
- **LDAP.enable_gssapi(kdc)** — Service principal, keytab, SASL maps, DS restart for GSSAPI/SASL tests.
- **KDC.config** — `rdns = false` in `[libdefaults]` for krb5 tests using `krb5_auth`.
- **docs/guides/testing-ldap-krb5.rst** — Documents `LDAP_KRB5` topology helpers, `client.net.dig`, `prepare_ldap_krb5_srv_discovery`, `setup_sasl_canonicalize_bogus_ptr`, and `resolve_ipv4`.
- **tests/test_tools.py** — Unit tests for `AHostSv4Entry.FromOutput`.

### DIFF
--- a/sssd_test_framework/roles/ldap.py
+++ b/sssd_test_framework/roles/ldap.py
@@ -3,9 +3,11 @@
 from __future__ import annotations
 
 import base64
+import tempfile
+import time
 from datetime import datetime
 from enum import Enum
-from typing import Any, Generic, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, TypeVar
 
 import ldap
 import ldap.ldapobject
@@ -16,6 +18,9 @@ from ..utils.ldap import LDAPRecordAttributes, LDAPUtils
 from .base import BaseLinuxLDAPRole, BaseObject, DeleteAttribute, HostType
 from .generic import GenericNetgroupMember, GenericPasswordPolicy, ProtocolName
 from .nfs import NFSExport
+
+if TYPE_CHECKING:
+    from .kdc import KDC
 
 __all__ = [
     "LDAPRoleType",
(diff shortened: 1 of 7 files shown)


danlavu (changes requested):
@madhuriupadhye 

So I wrote this a while ago, we have a dig utility

https://tests.sssd.io/en/latest/api/sssd_test_framework.utils.network.html#sssd_test_framework.utils.network.NetworkUtils

To add some more complication, I've changed the way it worked, in https://github.com/SSSD/sssd-test-framework/pull/231, which is going to be merged soon. Please rebase onto this PR, the existing dig stuff, and add only the getent.

sssd_test_framework/roles/ldap.py
```diff
@@ -239,6 +244,197 @@ def setup(self) -> None:
...
+        with tempfile.NamedTemporaryFile() as tmp:
+            kdc.host.fs.download(keytab_staging, tmp.name)
+            self.host.fs.upload(tmp.name, keytab_path)
+        kdc.host.conn.run(f"rm -f {keytab_staging}", raise_on_error=False)
+        self.host.conn.run(f"chown dirsrv:dirsrv {keytab_path}")
+        self.host.conn.run(f"chmod 600 {keytab_path}")
+
+        # 4. Copy krb5.conf from KDC to LDAP server
+        krb5_conf = kdc.config()
+        self.host.conn.run(f"cat > /etc/krb5.conf << 'EOFKRB5'\n{krb5_conf}\nEOFKRB5")
```
thalman, line 302: Is there reason to not use fs.write()?

sssd_test_framework/roles/ldap.py
```diff
@@ -239,6 +244,197 @@ def setup(self) -> None:
...
+
+        # Add default_keytab_name to krb5.conf as fallback
+        self.host.conn.run(f"sed -i '/\\[libdefaults\\]/a\\  default_keytab_name = {keytab_path}' /etc/krb5.conf")
+
+        # 5. Configure Cyrus SASL to use the keytab
+        # This is critical - without this, the SASL GSSAPI plugin won't know where to find the keytab
+        self.host.conn.run(
```
thalman, line 309: we have fs.mkdir and fs.write.

sssd_test_framework/roles/ldap.py
```diff
@@ -239,6 +244,197 @@ def setup(self) -> None:
...
+            f"mkdir -p /etc/sasl2 && "
+            f"cat > /etc/sasl2/slapd.conf << 'EOFSASL'\n"
+            f"mech_list: GSSAPI EXTERNAL PLAIN LOGIN\n"
+            f"keytab: {keytab_path}\n"
+            f"EOFSASL"
+        )
+
+        # Also create for other possible SASL application names
+        self.host.conn.run("cp /etc/sasl2/slapd.conf /etc/sasl2/ns-slapd.conf")
```
thalman, line 318: We have fs.copy

sssd_test_framework/roles/ldap.py
```diff
@@ -239,6 +244,197 @@ def setup(self) -> None:
...
+            f"-b '{gssapi_proxy_dn}' -s base '(objectclass=*)' dn",
+            raise_on_error=False,
+        )
+        if verify.rc != 0 or "dn:" not in (verify.stdout or ""):
+            raise RuntimeError(
+                f"SASL GSSAPI proxy entry {gssapi_proxy_dn} is not searchable: " f"{verify.stdout or verify.stderr}"
+            )
+
+        # 8. Reload systemd and restart Directory Server
+        self.host.conn.run("systemctl daemon-reload")
```
thalman, line 425: We have  SystemdServices class for this

sssd_test_framework/utils/network.py
```diff
@@ -145,6 +219,490 @@ def dig(self, address: str, server: str | None = None) -> list[dict] | None:
...
+        :type bogus_label: str, optional
+        :return: Resolved (ldap_ip, kdc_ip) used for the setup.
+        :rtype: tuple[str, str]
+        :raises RuntimeError: if resolution, bind install, or DNS lookups fail.
+        """
+        if ldap_ip is None:
+            ldap_ip = self._role_ipv4(ldap_hostname, role_host=ldap_host, label="LDAP host")
+        if kdc_ip is None:
+            kdc_ip = self._role_ipv4(kdc_hostname, role_host=kdc_host, label="KDC host")
+        if self.host.conn.run("rpm -q bind", raise_on_error=False).rc != 0:
```
thalman, line 565: maybe `rpm  -q bind && rpm -q bind-utils`? WDYT?

thalman (changes requested):
Generally looks good.

What I really do not like is running dnf to install bind. IMO those packages should be preinstalled (and service disabled) if we need them. Is that possible?

sssd_test_framework/utils/tools.py
```diff
@@ -1058,6 +1089,53 @@ def hosts(self, name: str, *, service: str | None = None) -> HostsEntry:
         """
         return self.__exec(HostsEntry, "hosts", name, service)
 
+    def ahostsv4(self, name: str, *, service: str | None = None) -> AHostSv4Entry | None:
```
danlavu, line 1092: I think it'd be better if we made one method to support both ipv4 and ipv6 lookups, so something like this?

```def ahosts(self, name: str, *, services: str | None = None, ipv6: bool = False) -> AhostEntry:

database = "ahostsv6" if ipv6 else "ahostssv4"

command = self.host.conn.exec(["getent", *args, database, str(name)], raise_on_error=False)```

danlavu:
Okay, so as we discussed, this should be changed. My idea is to drop the LDAP changes and use IPA instead. 

* We talked about a new topology, but because it's not a topology, I'd add the ldap_provider setup to the test case. 

```client.sssd.ldap_provider(server=ipa_server, naming_context=ipa.naming_context, bind_user="Admin") ```

* Hopefully, all the changes you need to make for the accounts are supported with the IPA command, but if it isn't, write wrapper methods for IPAUser, IPAGroup to modify the principals, etc.

PR #257, opened 2026-07-15
Fix: increase vfido touch delay to avoid IPA auth timeout

aborah-sudo:
The 2-second sleep before vfido_touch was too short for IPA backend lookups on fedora-43/rawhide, causing the 3rd sequential passkey authentication to fail in test_passkey__su_user_same_key_for_other_users.


aborah-sudo:
We have green run from :  https://github.com/SSSD/sssd/pull/8949

spoore1 (approved):
I'm ok with this.    The tests appear to be working from the checks here:
https://github.com/SSSD/sssd/pull/8949

PR #258, opened 2026-07-15
ci: pin tox<4.56 to fix tox-gh Python version filtering

aborah-sudo:
tox 4.56 broke tox-gh's [gh] section mapping, causing the py311 env to run even when Python 3.x is active. This results in "could not find python interpreter matching any of the specs py311" failures on the tox (3.x, *) CI jobs.


pbrezina:
Closing in favor of https://github.com/SSSD/sssd-test-framework/pull/259

PR #262, opened 2026-07-31
adding allrealms topology.

danlavu:
A workaround for tests that require multiple domains.


pbrezina:
I think you also need to add a topology controller to handle the setup? Or is client topology controller really sufficient? What is the use case?

danlavu:
> I think you also need to add a topology controller to handle the setup? Or is client topology controller really sufficient? What is the use case?

It's sufficient; the original test used dummy Kerberos realms; the realms just need to be resolvable.

pbrezina:
> > I think you also need to add a topology controller to handle the setup? Or is client topology controller really sufficient? What is the use case?
> 
> It's sufficient; the original test used dummy Kerberos realms; the realms just need to be resolvable.

Maybe add a comment there to say that if working Kerberos realm is required, we need to change it?

danlavu:
> Maybe add a comment there to say that if working Kerberos realm is required, we need to change it?

I'm using the term 'dummy' loosely, the realm is working, the test does create users in all raelms and ensures that tickets can be issued. I think it's fine then, is that fine with you?

krishnavema (approved):
LGTM!

PR #263, opened 2026-07-31
topology: Introduce KnownTopology.ADForest

jakub-vavra-cz:
topology: Introduce KnownTopology.ADForest
Using: Cursor

### DIFF
--- a/sssd_test_framework/fixtures.py
+++ b/sssd_test_framework/fixtures.py
@@ -5,6 +5,10 @@ from __future__ import annotations
 import os
 
 import pytest
+from pytest_mh import mh_fixture
+
+from .roles.ad import AD
+from .roles.client import Client
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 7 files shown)


thalman (approved):
LGTM

danlavu:
There is a much better way to do this, and I think you'll see the value in what I'm proposing: _Let's run three clients._ 

- All three domains run in one test — no leave/rejoin churn
- Three small containers is cheap — seconds to boot, no real cost
- No static join fixtures, no repeated join/teardown code
- Controller stops being a no-op — it owns the joins in topology_setup
- Role subclasses for child and tree make the domain hierarchy explicit

I really do not like the static code; if we need to add more, it's just going to get complicated. I think the role subclasses are the way to go. WYDT?

sssd_test_framework/roles/ad.py
```diff
@@ -523,6 +522,42 @@ def test_ad__gpo_is_set_to_enforcing(client: Client, ad: AD):
         """
         return GPO(self, name)
 
+    def scrub_orphan_gpo_links(self, site: str = "Default-First-Site-Name") -> None:
```
danlavu: The cleanup should happen in hosts/ad , can we improve the cleanup there?

sssd_test_framework/roles/ad.py
```diff
@@ -2270,6 +2388,25 @@ def filesyspath(self, path: str) -> GPO:
 
         return self
 
+    @classmethod
```
danlavu: Yea, let's see if it we can fix it in hosts/ad.py

```
            foreach ($r in $gpo) {{
                if (!$backup_gpo.contains($r.DistinguishedName)) {{
                    if ($sites.gplink -Like "*$r*") {{
                        Remove-GPLink -GUID $r.Name -Target "Default-First-Site-Name"
                        Write-Host Removing: "*$r*" from Default-First-Site-Name
                    }}
                    foreach ($g in $link) {{
                        if ($g.gplink -Like "*$r*") {{
                            Remove-GPLink -GUID $r.Name -Target $g
                            Write-Host Removing: "*$r*" from $g
                        }}
``` 

So this is leaving an artifact?

sssd_test_framework/fixtures.py
```diff
@@ -43,3 +47,198 @@ def testdatadir(moduledatadir: str, request: pytest.FixtureRequest) -> str:
 
     name = request.node.originalname
     return os.path.join(moduledatadir, name)
+
```
danlavu: I really like the fixtures, it's going to be hard to maintain, we should look at the three client idea. I think that way we can drop the fixtures and just use subclasses.

PR #180, opened 2025-04-28
realm: add methods for realm utility

shridhargadekar:
realm utility is a tool manage enrollment in realm. This PR is adding methods provided by realm utility.

PR #264, opened 2026-07-31
ldaps: add LDAPS support

shridhargadekar:
- Add export_root_ca_certificate() to ADHost, SambaHost, and IPAHost
- Add CertUtils (client.cert) utility for system-level CA cert install into /etc/openldap/ldap.conf — works with adcli, realmd, ldapsearch
- Add SSSDCommonConfiguration helpers: ad_use_ldaps(), samba_use_ldaps(), ipa_set_tls_cacert() for SSSD-specific LDAPS configuration


sssd_test_framework/hosts/ad.py
```diff
@@ -100,6 +100,56 @@ def naming_context(self) -> str:
 
         return self.__naming_context
 
+    def export_root_ca_certificate(self) -> str:
```
danlavu: These functions doesn't belong in hosts, it should be in roles and actually already have this method for this. 

You should also add it to the generic provider because realm join works for AD, Samba, and IPA, allowing you to write a single test for multiple topologies using TopologyGroup.AnyDC or AnyAD

sssd_test_framework/hosts/ipa.py
```diff
@@ -114,6 +114,27 @@ def setup(self) -> None:
             truncate --size 0 /var/log/krb5kdc.log
             """)
 
+    def export_root_ca_certificate(self) -> str:
```
danlavu: Like the previous comment, this doesn't belong in hosts, it should be in roles,  there is an IPACertifcateAuthority.get() which may work for the CA as well. This CertificateAuthority class should be extended.

sssd_test_framework/hosts/samba.py
```diff
@@ -36,6 +36,9 @@ def __init__(self, *args, **kwargs) -> None:
         self.adminpw: str = self.config.get("adminpw", self.bindpw)
         """Password of the admin user, defaults to value of ``bindpw``."""
 
+        self.ca_cert_path: str = self.config.get("ca_cert_path", "/var/data/certs/ca.crt")
```
danlavu, line 39: Same as the previous comment, but this has no CertificateAuthority class. Please make each role a separate commit.

sssd_test_framework/utils/cert.py
```diff
@@ -0,0 +1,193 @@
...
+
+    Installs CA certificates so that any tool using the system OpenLDAP
+    library (adcli, realmd, ldapsearch, etc.) can connect over LDAPS.
+    All changes are tracked by the framework and restored automatically
+    after each test.
+
+    .. code-block:: python
+        :caption: Example usage with adcli
+
+        @pytest.mark.topology(KnownTopology.Samba)
```
danlavu: I don't think this test scenario is really valid, since the KnownTopology.Samba is already joined when the test starts. Have you tried running this test?

sssd_test_framework/utils/cert.py
```diff
@@ -0,0 +1,193 @@
...
+from pytest_mh import MultihostUtility
+from pytest_mh.utils.fs import LinuxFileSystem
+
+from ..hosts.client import ClientHost
+
+__all__ = ["CertUtils"]
+
+
+class _HasCACert(Protocol):
+    """Any host that can export its root CA certificate."""
```
danlavu: Moving to roles, this shouldn't be necessary.

sssd_test_framework/utils/cert.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+
+    def __init__(self, host: ClientHost, fs: LinuxFileSystem) -> None:
+        super().__init__(host)
+        self.fs: LinuxFileSystem = fs
+
+    def install_ca(
+        self,
+        host: _HasCACert,
+        name: str = "test-ldap-ca.crt",
```
danlavu: Nitpick, the ca.crt doesn't have to be used for ldap, I'd omit ldap from the name.

sssd_test_framework/utils/sssd.py
```diff
@@ -1225,3 +1228,69 @@ def smartcard_with_softhsm(self, smartcard: SmartCardUtils) -> None:
...
+        :type ca_cert: str
+        :param cacert_path: Absolute path on the client where the cert is written.
+        :type cacert_path: str
+        """
+        parent = posixpath.dirname(cacert_path)
+        if parent:
+            self.sssd.fs.mkdir_p(parent)
+        self.sssd.fs.write(cacert_path, ca_cert)
+
+    def ad_use_ldaps(self, ad: AD, cacert_path: str = "/etc/sssd/pki/ad-ca.crt") -> None:
```
danlavu: We already have a method to configure ldap, SSSDCommonConfiugration.ldap_provider, instead of adding three new methods, please extend this method to support ldaps.


    def ldap_provider(
        self,
        server: str,
        naming_context: str,
        bind_user_dn: str,
        bind_password: str,
        subids: bool = False,
        cacert: str = "/etc/ipa/ca.crt",
        tls_reqcert: str = "demand",
        ssl: bool = False,
        config: dict[str, str] | None = None,
    ) -> None:

danlavu (changes requested):

sssd_test_framework/roles/ad.py
```diff
@@ -217,6 +217,50 @@ def test_example(client: Client, ad: AD):
         """
         return self._password_policy
 
+    def export_root_ca_certificate(self) -> str:
```
spoore1, line 220: This won't be needed when we re-enable the AD CA class right?   

I already have an approved [PR to re-enable](https://github.com/SSSD/sssd-test-framework/pull/255) it but, I need to first get [the sssd-ci-containers dependency](https://github.com/SSSD/sssd-ci-containers/pull/174) merged.   I'm planning to do that early next week unless it's urgent to get this in place sooner?

sssd_test_framework/roles/ad.py
```diff
@@ -3106,3 +3150,49 @@ def get_ca_cert(self) -> str:
 
         ca_cert_lines = [ca_cert_b64[i : i + 64] for i in range(0, len(ca_cert_b64), 64)]
         return "-----BEGIN CERTIFICATE-----\n" + "\n".join(ca_cert_lines) + "\n-----END CERTIFICATE-----\n"
+
+    def export_root_ca_certificate(self) -> str:
```
spoore1, line 3131: could this make use of self.get_ca_cert() above to handle the call and then just wrap and write the file and return as you've done here? 

It may require some of your changes here if these are more reliable (may be from a quick glance).

shridhargadekar:
Depends upon https://github.com/SSSD/sssd-test-framework/pull/255

sssd_test_framework/roles/ad.py
shridhargadekar: @spoore1 wanna review this again?

sssd_test_framework/roles/ad.py
shridhargadekar: Once dependent PR is merged, this can modify

sssd_test_framework/roles/client.py
```diff
@@ -222,3 +225,128 @@ def sudorule(self, name: str) -> LocalSudoRule:
...
+
+        self.fs.write(cert_path, pem)
+        self.host.conn.run("update-ca-trust")
+        self._configure_tls_cacert()
+
+        return cert_path
+
+    def _configure_tls_cacert(self) -> None:
+        """Point ``/etc/openldap/ldap.conf`` at the system CA bundle."""
+        system_bundle = "/etc/pki/tls/certs/ca-bundle.crt"
```
sumit-bose: Hi,

with Fedora 44 this changed, see https://fedoraproject.org/wiki/Changes/droppingOfCertPemFile for details.

Since you add the CA certificates to the system store, which sould be used by OpenLDAP by default, maybe the better strategy would be to make sure that all occurrences of `TLS_CACERT` and `TLS_CACERTDIR` are commented out in `/etc/openldap/ldap.conf` to force libldap to use the defaults?

bye,
Sumit

sssd_test_framework/utils/sssd.py
```diff
@@ -1225,3 +1226,51 @@ def smartcard_with_softhsm(self, smartcard: SmartCardUtils) -> None:
...
+
+            @pytest.mark.topology(KnownTopologyGroup.AnyDC)
+            def test_example(client: Client, provider: GenericProvider):
+                client.sssd.common.use_ldaps(provider)
+                client.sssd.start()
+
+        :param provider: Provider role to fetch the root CA certificate from.
+        :type provider: GenericProvider
+        :param cacert_path: Path on the client where the CA cert is written,
+            defaults to ``/etc/sssd/pki/ca.crt``
```
sumit-bose: Hi,

would it be better to let use SSSD the system's CA certificate store by default and only use a dedicated file if one is provided?

bye,
Sumit

sssd_test_framework/roles/client.py
```diff
@@ -222,3 +225,128 @@ def sudorule(self, name: str) -> LocalSudoRule:
...
+        port: int = 636,
+        name: str = "test-ca.crt",
+        cert_path: str | None = None,
+    ) -> str:
+        """
+        Install a CA certificate by fetching it directly from a TLS server.
+
+        Connects to ``hostname:port`` with ``openssl s_client``, captures the
+        certificate the server presents, writes it to the system trust anchor
+        directory, runs ``update-ca-trust``, and points ``TLS_CACERT`` in
```
sumit-bose, line 284: See comments below about the system certificates and Fedora 44.

sssd_test_framework/roles/ad.py
```diff
@@ -217,6 +217,50 @@ def test_example(client: Client, ad: AD):
         """
         return self._password_policy
 
+    def export_root_ca_certificate(self) -> str:
+        """
+        Export the AD root CA certificate in PEM format.
```
sumit-bose, line 222: Hi,

you can get the CA certificate directly with `certutil.exe -ca.cert filename`.

bye,
Sumit

sssd_test_framework/roles/ad.py
spoore1: Both of those PRs are now merged.  You can rebase and pick up the updates to use here.

sssd_test_framework/hosts/samba.py
danlavu: Still needs to move to the ad role.

sssd_test_framework/roles/client.py
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import posixpath
```
danlavu: Everywhere else, we just use os.path I suggest keeping it consistent.

sssd_test_framework/utils/sssd.py
```diff
@@ -1225,3 +1226,51 @@ def smartcard_with_softhsm(self, smartcard: SmartCardUtils) -> None:
...
+        :type ca_cert: str
+        :param cacert_path: Absolute path on the client where the cert is written.
+        :type cacert_path: str
+        """
+        parent = posixpath.dirname(cacert_path)
+        if parent:
+            self.sssd.fs.mkdir_p(parent)
+        self.sssd.fs.write(cacert_path, ca_cert)
+
+    def use_ldaps(self, provider: GenericProvider, cacert_path: str = "/etc/sssd/pki/ca.crt") -> None:
```
danlavu: Please extend the existing method.

sssd_test_framework/roles/ad.py
spoore1: This doesn't seem to work for me (at least not in the VM we have in sssd-ci-containers):
```
PS C:\Users\Administrator\AppData\Local\Temp> certutil.exe -ca.cert C:\temp\ca.crt
Expected no more than 1 args, received 2
CertUtil: Too many arguments
```

The old PowerShell code though did appear to work better but, I'm not sure it actually exported the correct certificate.   It seems to have exported the AD Certificate for `CN=dc.ad.test` not the `CN=ad-RootCA, DC=ad, DC=test` one.   I'm guessing the latter is needed?

sssd_test_framework/roles/ad.py
spoore1: I may have misread this initially.   Is this needed because not all providers have a CA object with an export method?

sssd_test_framework/roles/ad.py
```diff
@@ -214,44 +217,75 @@ def test_example(client: Client, ad: AD):
...
-                ad.ca.revoke(cert, reason="key_compromise")
+        :return: PEM-formatted root CA certificate.
+        :rtype: str
+        :raises RuntimeError: If the root CA certificate cannot be exported.
         """
-        return self._ca
+        result = self.host.conn.run(
+            """
+            $tmp = [System.IO.Path]::GetTempFileName() + ".crt"
+            certutil -ca.cert $tmp | Out-Null
```
spoore1: As I mentioned above, this didn't work for me when I was testing the changes.   The old code in a previous commit came closer but, it exported a certificate for "cn=dc.ad.test" not the "CN=ad-RootCA, DC=ad, DC=test"  Root CA certificate.   The get_ca_cert() method code grabs the ad-RootCA one so maybe just use that for the AD method?

sssd_test_framework/roles/client.py
```diff
@@ -222,3 +225,133 @@ def sudorule(self, name: str) -> LocalSudoRule:
         """
 
         return self.local.sudorule(name)
+
+    def install_ca(
```
spoore1: Maybe name this one install_ca_cert or install_ca_certificate?   My first thought with install_ca was that this was for setting up an openssl local CA.

spoore1 (changes requested):
I'm not sure the AD export method is working as expected.   My testing showed the certutil fail and the old commit worked better but, I'm not sure it's pulling the expected certificate.

sssd_test_framework/utils/sssd.py
shridhargadekar: Are you referring to `client.install_ca()` from `client.py`? Or any other file.

sssd_test_framework/roles/client.py
sumit-bose: Hi,

about `certutil`, you might have to use quotes in PowerShell like e.g. `certutil.exe -"ca.cert" C:\temp\ca.crt` to avoid some shell magic.

HTH

bye,
Sumit

sssd_test_framework/roles/ad.py
shridhargadekar: Hi Sumit,
The suggested method does not work as expected. thus i've reverted to previous one.

sumit-bose: Hi,

sorry, looks like my comment about using quotes landed in a different thread.

The current version, reading `C:\Windows\Temp\ad-root-ca.crt` does not work for me, a Microsoft CA certificate is returned.

The following implementation worked for me, e.g. with the adcli LDAPS tests:

```
    def export_root_ca_certificate(self) -> str:
        """
        Export the AD root CA certificate in PEM format.

        :return: PEM-formatted root CA certificate.
        :rtype: str
        :raises RuntimeError: If the root CA certificate cannot be read.
        """
        result = self.host.conn.run(
            'certutil.exe -f -"ca.cert" C:\\temp\\ca.crt',
            raise_on_error=False,
        )
        result = self.host.conn.run(
            'certutil.exe -f C:\\temp\\ca.crt C:\\temp\\ca.pem',
            raise_on_error=False,
        )
        result = self.host.conn.run(
            'Get-Content C:\\temp\\ca.pem -Raw',
            raise_on_error=False,
        )

        if result.rc != 0:
            raise RuntimeError(f"Failed to export root CA certificate: {result.stderr}")

        return result.stdout.strip() + "\n"
```

HTH

bye,
Sumit

sssd_test_framework/roles/client.py
```diff
@@ -222,3 +225,138 @@ def sudorule(self, name: str) -> LocalSudoRule:
...
+
+        return cert_path
+
+    def _configure_tls_cacert(self) -> None:
+        """
+        Configure ``/etc/openldap/ldap.conf`` for system CA trust and channel binding.
+
+        Removes any explicit ``TLS_CACERT`` and ``TLS_CACERTDIR`` directives so
+        libldap falls back to its compiled-in defaults (the system trust store).
+        Sets ``SASL_CBINDING tls-server-end-point`` for channel binding support.
```
sumit-bose: Hi,

the option value is just `tls-endpoint`, see man ldap.conf for details.

bye,
Sumit

sssd_test_framework/roles/client.py
```diff
@@ -222,3 +225,138 @@ def sudorule(self, name: str) -> LocalSudoRule:
...
+        current = result.stdout if result.rc == 0 else ""
+
+        lines = [
+            line
+            for line in current.splitlines()
+            if not line.startswith("TLS_CACERT")
+            and not line.startswith("TLS_CACERTDIR")
+            and not line.startswith("SASL_CBINDING")
+        ]
+        lines.append("SASL_CBINDING tls-server-end-point")
```
sumit-bose: see above

sssd_test_framework/roles/client.py
sumit-bose: Thanks for fixing this in the latest version.

sssd_test_framework/roles/client.py
sumit-bose: Thanks for fixing this in the latest version.

sumit-bose:
Hi,

the latest version does not use `certutil.exe` as the commit messages says and the copied file does not contain the CA certificate of the CS of the domain. Please see my https://github.com/SSSD/sssd-test-framework/pull/264#discussion_r3916183945 with steps for `certutil.exe` working for me.

HTH

bye,
Sumit

sumit-bose:
Hi,

thank you for the update. I just realized that there is a copy-and-paste error in one of the certutil calls, the conversion from binary to PEM should be
```
self.host.conn.run("certutil.exe -f -encode C:\\temp\\ca.crt C:\\temp\\ca.pem", raise_on_error=False)
```
`-encode` was missing. If this is added the related adcli tests are passing with the SSSD test containers together with the AD instance.

Please add `-encode`, rebase the patches and maybe reduce the number of patches by squashing related ones.

bye,
Sumit

shridhargadekar:
> Hi,
> 
[......]
> Please add `-encode`, rebase the patches and maybe reduce the number of patches by squashing related ones.
> 
> bye, Sumit

Hi Sumit,
Thanks for additional comments. I've added the `--encode`. Please check now.

sumit-bose:
> > Hi,
> 
> [......]
> 
> > Please add `-encode`, rebase the patches and maybe reduce the number of patches by squashing related ones.
> > bye, Sumit
> 
> Hi Sumit, Thanks for additional comments. I've added the `--encode`. Please check now.

Hi,

thank you for the updates. Tests are passing and I have no further comments to the code. Please squash related patches to reduce the number of patches.

bye,
Sumit

shridhargadekar:
> thank you for the updates. Tests are passing and I have no further comments to the code. Please squash related patches to reduce the number of patches.

Hello,
I've squashed the related patches to reduce the total number of patches. Please check

sumit-bose:
> > thank you for the updates. Tests are passing and I have no further comments to the code. Please squash related patches to reduce the number of patches.
> 
> Hello, I've squashed the related patches to reduce the total number of patches. Please check

Hi,

 I would like to suggest to move the changes in sssd_test_framework/roles/ad.py from "roles/client: add install_ca() and install_ca_from_server()" to "roles/ad: add export_root_ca_certificate()" and the changes in sssd_test_framework/utils/sssd.py from "roles/client: add install_ca() and install_ca_from_server()" to "utils/sssd: add SSSDCommonConfiguration.use_ldaps()".

bye,
Sumit

shridhargadekar:
Hi @sumit-bose, I've modified the commits as suggested. Please check.

Thanks,
shridhar

sumit-bose (dismissed):
Hi,

thank you for re-ordering the patches, ACK.

bye,
Sumit

sssd_test_framework/roles/ad.py
spoore1: after the update, this is working for me.

sumit-bose:
Hi,

why did you add the two other methods to get the AD CA certificate? The first variant returns a wrong certificate and makes the tests fail. The second one produces not output, which at least does not do any harm. Why isn't the `certutil.exe` variant sufficient?

bye,
Sumit

shridhargadekar:
Hi Sumit,
I was trying to accommodate for container LDAPS support pr [CI-180](https://github.com/SSSD/sssd-ci-containers/pull/180).  For now i've reverted to previous `certutils.exe` method. I'll work on `#180` separately.
Please check now.
Thanks,
Shridhar

sumit-bose:
> Hi Sumit, I was trying to accommodate for container LDAPS support pr [CI-180](https://github.com/SSSD/sssd-ci-containers/pull/180). For now i've reverted to previous `certutils.exe` method. I'll work on `#180` separately. Please check now. Thanks, Shridhar

Hi,

thank you, still ACK.

bye,
Sumit

spoore1 (dismissed):
LGTM.   

I also ran tests using this adcli MR:
https://gitlab.freedesktop.org/realmd/adcli/-/merge_requests/104

Results:
```
======================================== short test summary info ========================================
PASSED tests/test_adcli.py::test_adcli__join_and_update_with_ldaps (ad)
PASSED tests/test_adcli.py::test_adcli__delete_computer_with_ldaps (ad)
PASSED tests/test_adcli.py::test_adcli__join_and_update_with_ldaps (samba)
PASSED tests/test_adcli.py::test_adcli__delete_computer_with_ldaps (samba)
============================== 4 passed, 84 deselected in 64.38s (0:01:04) ==============================
```

danlavu (changes requested):
It looks better but you haven't addresssed my comments.

sssd_test_framework/hosts/samba.py
shridhargadekar: moved it to samba role

shridhargadekar: moved to mab role

sssd_test_framework/utils/sssd.py
danlavu: No, we have a method sssd.common.ldap_provider(), which is configured to set up IPA using the LDAP provider. Please extend this method, instead of adding 'use_ldaps'

 def ldap_provider(
    self,
    server: str,
    naming_context: str,
    bind_user_dn: str,
    bind_password: str,
    subids: bool = False,
    cacert: str = "/etc/ipa/ca.crt",
    tls_reqcert: str = "demand",
    ssl: bool = False,
    config: dict[str, str] | None = None,
) -> None:

